PR #24, opened 2022-03-03
Off-grid capabilities

adfarth:
Main off-grid capabilities to include:

- [x] Flag to "turn on" off-grid modeling 
- [x] Operating reserve requirements based on load and PV generation 
- [x] Generator replacement year and cost (add to on-grid as well) 
- [x] Minimum load met percent
- [x] Additional annual and upfront capital costs
- [x] Report LCOE and LCOE breakdown --> reporting LCC breakdown instead (applicable for on-grid as well) 
- [x] Adjust default values for off-grid 
- [x] Change all instances of `timestep` to `time_step` and `timesteps` to `time_steps`


src/constraints/load_balance.jl
```diff
@@ -66,8 +66,19 @@ function add_elec_load_balance_constraints(m, p; _n="")
...
         sum( sum(m[Symbol("dvProductionToStorage"*_n)][b, t, ts] for b in p.s.storage.types.elec) + 
         m[Symbol("dvCurtail"*_n)][t, ts] for t in p.techs.elec) +
-        p.s.electric_load.critical_loads_kw[ts]
+        (p.s.electric_load.critical_loads_kw[ts] * m[Symbol("dvOffgridLoadServedFraction"*_n)][ts] )
 	)
+
+    ##Constraint : For off-grid scenarios, annual load served must be >= minimum percent specified  
+    if !p.s.settings.off_grid_flag # fix dvOffgridLoadServedFraction to 100% for "on-grid" analyses
+		for ts in p.time_steps_without_grid
+			fix(m[Symbol("dvOffgridLoadServedFraction"*_n)][ts], 1.0, force=true) ## TODO: Check if this messes with the dvUnservedLoad outage constraints
```
adfarth, line 75: @NLaws This step of fixing the "load served fraction" to 100% for on-grid runs is something we discussed before (you had suggested this as a way to trim down the code by not creating a separate load balance constraint for off-grid runs). However, I'm wondering if you think this will now be an issue with the new outage modeling capabilities (related to dvUnservedLoad)? I haven't done too much digging into this yet but wanted to flag it now in case you see any issues off the bat.

NLaws: The `dvUnservedLoad` variable is only created if `p.s.electric_utility.outage_durations` is non-empty. Can we also create `dvOffgridLoadServedFraction` conditionally?  No outages are modeled with `off_grid_flag=true` right?  We could provide a warning when a user provides outage information and `off_grid_flag=true` that the outage information is not used when `off_grid_flag=true`.

src/core/scenario.jl
```diff
@@ -91,33 +102,39 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
...
+        electric_utility = ElectricUtility(; outage_start_time_step = 1, outage_end_time_step = settings.time_steps_per_hour * 8760) 
     end
 
     storage_structs = Dict{String, AbstractStorage}()
     if haskey(d,  "ElectricStorage")
         storage_dict = dictkeys_tosymbols(d["ElectricStorage"])
+        storage_dict[:off_grid_flag] = settings.off_grid_flag
     else
-        storage_dict = Dict(:max_kw => 0.0)
+        storage_dict = Dict(:max_kw => 0.0) # TODO: do we want to change to the off-grid inputs even if not modeling storage?
```
NLaws, line 137: I don't understand the question for this TODO: what does it mean to "change to the off-grid inputs"?

src/core/utils.jl
```diff
@@ -183,14 +184,37 @@ function dictkeys_tosymbols(d::Dict)
...
+        if k in [
+            "outage_start_time_steps", "outage_durations"
+        ]
+            try
+                v = convert(Array{Int64, 1}, v)
+            catch
+                @warn "Unable to convert $k to a Array{Int64, 1}"
+            end
+        end
+        if k in [
```
NLaws, line 204: I think that all of these converts could be removed by changing their type to Real in the struct definitions.

src/core/utils.jl
```diff
@@ -378,3 +402,16 @@ function convert_gal_to_kwh(delta_T_degF::Real, rho_kg_per_m3::Real, cp_kj_per_k
     kwh_per_gal = kj_per_gal / 3600.0  # divide by kJ/kWh, i.e., sec/hr, to get: [kWh/gal]
     return kwh_per_gal
 end
+
+"""
+    The input offgrid_other_capital_costs is considered to be for depreciable assets. 
+    Straight line depreciation is applied, and the depreciation expense is assumed to reduce the owner's taxable income
+    TODO: should the savings array start at [0] or [1] ??
```
NLaws, line 409: I don't understand the question for this TODO. Can we address it?

src/lindistflow/extend.jl
```diff
@@ -60,43 +60,43 @@ function LDF.constrain_loads(m::JuMP.AbstractModel, p::LDF.Inputs, ps::Array{REo
     for j in p.busses
         if j in keys(p.Pload)
             if parse(Int, j) in reopt_nodes
-                @constraint(m, [t in 1:p.Ntimesteps],
+                @constraint(m, [t in 1:p.Ntime_steps],
```
NLaws: this is a breaking change. (The p here is LDF.Inputs, not REoptInputs). I will fix it.

src/results/electric_tariff.jl
```diff
@@ -34,42 +34,42 @@ Adds the ElectricTariff results to the dictionary passed back from `run_reopt` u
...
 ElectricTariff results:
-- `lifecycle_energy_cost` lifecycle cost of energy from the grid in present value, after tax
-- `year_one_energy_cost` cost of energy from the grid over the first year
-- `lifecycle_demand_cost` lifecycle cost of power from the grid in present value, after tax
-- `year_one_demand_cost` cost of power from the grid over the first year
-- `lifecycle_fixed_cost` lifecycle fixed cost in present value, after tax
-- `year_one_fixed_cost` fixed cost over the first year
-- `lifecycle_min_charge_adder` lifecycle minimum charge in present value, after tax
-- `year_one_min_charge_adder` minimum charge over the first year
+- `lifecycle_energy_cost_after_tax` lifecycle cost of energy from the grid in present value, after tax
```
NLaws, line 37: lots of breaking changes with these name changes. please document in the CHANGELOG

src/results/financial.jl
```diff
@@ -54,19 +71,56 @@ function add_financial_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _
...
         p.third_party_factor
 
     if !isempty(p.techs.fuel_burning)
         r["lifecycle_fuel_costs_after_tax"] = value(m[:TotalFuelCosts]) * (1 - p.s.financial.offtaker_tax_pct)
+    else
+        r["lifecycle_fuel_costs_after_tax"] = 0.0
+    end
+
+    # LCC breakdown:
+    r["lcc_fraction_tech_capital_cost"] = value(m[Symbol("TotalTechCapCosts"*_n)]) / r["lcc"] # Tech capital costs (including replacements)
```
NLaws, line 90: `lcc_..._cost` seems redundant. also, why do we need to report all these fractions? seems like a lot of redundant information as well. Why not just report the costs instead of dividing them by the lcc?

src/results/generator.jl
```diff
@@ -35,13 +35,13 @@ Note: the node number is an empty string if evaluating a single `Site`.
...
-- `year_one_fixed_om_cost` fixed operations and maintenance cost over the first year
-- `lifecycle_variable_om_cost` Lifecycle variable operations and maintenance cost in present value, after tax
-- `year_one_variable_om_cost` variable operations and maintenance cost over the first year
-- `lifecycle_fuel_cost` Lifecycle fuel cost in present value, after tax
-- `year_one_fuel_cost` Fuel cost over the first year
-- `fuel_used_gal` Gallons of fuel used in each year
+- `lifecycle_fixed_om_cost_after_tax` Lifecycle fixed operations and maintenance cost in present value, after tax
+- `year_one_fixed_om_cost_after_tax` fixed operations and maintenance cost over the first year, not considering tax benefits
+- `lifecycle_variable_om_cost_after_tax` Lifecycle variable operations and maintenance cost in present value, after tax
+- `year_one_variable_om_cost` variable operations and maintenance cost over the first year, not considering tax benefits
```
NLaws, line 41: we have some Financial outputs with `_before_tax` but we don't use `_before_tax` here. We should be consistent with the `_after_tax` and `_before_tax` results. I prefer removing `_before_tax` but either way works.

src/results/generator.jl
```diff
@@ -86,15 +86,15 @@ function add_generator_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _
 	r["year_one_to_load_series_kw"] = round.(value.(generatorToLoad), digits=3)
 
     GeneratorFuelUsed = @expression(m, sum(m[:dvFuelUsage][t, ts] for t in p.techs.gen, ts in p.time_steps))
-	r["fuel_used_gal"] = round(value(GeneratorFuelUsed), digits=2)
+	r["average_annual_fuel_used_gal"] = round(value(GeneratorFuelUsed), digits=2)
```
NLaws, line 89: `average_annual_fuel_used_gal` is not really an average right? if not I suggest removing `average_`.

test/test_with_xpress.jl
```diff
@@ -547,7 +547,7 @@ end
     @testset "Coincident Peak Charges" begin
         model = Model(optimizer_with_attributes(Xpress.Optimizer, "OUTPUTLOG" => 0))
         results = run_reopt(model, "./scenarios/coincident_peak.json")
-        @test results["ElectricTariff"]["year_one_coincident_peak_cost"] ≈ 15.0
+        @test results["ElectricTariff"]["year_one_coincident_peak_cost"] ≈ 11.1
```
NLaws, line 550: why did the year_one_coincident_peak_cost change?

NLaws (changes requested):

src/core/financial.jl
```diff
@@ -60,37 +63,56 @@ function Financial(;
...
+        microgrid_upgrade_cost_pct::Real = off_grid_flag ? 0.0 : 0.3,
         macrs_five_year::Array{Float64,1} = [0.2, 0.32, 0.192, 0.1152, 0.1152, 0.0576],  # IRS pub 946
-        macrs_seven_year::Array{Float64,1} = [0.1429, 0.2449, 0.1749, 0.1249, 0.0893, 0.0892, 0.0893, 0.0446]
+        macrs_seven_year::Array{Float64,1} = [0.1429, 0.2449, 0.1749, 0.1249, 0.0893, 0.0892, 0.0893, 0.0446],
+        offgrid_other_capital_costs::Real = 0.0, # only applicable when off_grid_flag is true. Straight-line depreciation is applied to this capex cost, reducing taxable income.
+        offgrid_other_annual_costs::Real = 0.0 # only applicable when off_grid_flag is true. Considered tax deductible for owner.
     )
+        
+        if off_grid_flag && !(microgrid_upgrade_cost_pct == 0.0)
+            @warn "microgrid_upgrade_cost_pct is not applied when off_grid_flag is true. Setting microgrid_upgrade_cost_pct to 0.0."
```
NLaws, line 106: this is incorrect: `microgrid_upgrade_cost_pct` is also used for the uncertain outage model ((t is used to compete against the value of lost load in determining the microgrid capacities). Can we just leave it as non-zero even if it is not used?

src/core/scenario.jl
```diff
@@ -82,6 +82,17 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
     end
     
     site = Site(;dictkeys_tosymbols(d["Site"])...)
+
+    # Check that only PV, storage, and generator are modeled for off-grid
+    if settings.off_grid_flag
+        for key in ["HotThermalStorage", "ColdThermalStorage", "Wind", "FlexibleHVAC",
```
NLaws, line 88: we could make this require less maintenance (i.e. not have to add items to this list) by using a setdiff of keys(d) and the allowed off grid techs:
```julia
julia> offgrid_techs = ["PV", "ElectricStorage", "Generator"]
3-element Vector{String}:
 "PV"
 "ElectricStorage"
 "Generator"

julia> user_keys = ["PV", "CHP"]
2-element Vector{String}:
 "PV"
 "CHP"
 
 julia> unallowed_keys = setdiff(user_keys, offgrid_techs)
1-element Vector{String}:
 "CHP"
 ```
 Then have `if !isempty(unallowed_keys) ...`

**Review by adfarth — COMMENTED** [review]
Replied to comments above

File: src/core/financial.jl [hunk 13]
```diff
@@ -64,6 +67,7 @@ struct Financial
     elec_cost_escalation_pct::Float64
     boiler_fuel_cost_escalation_pct::Float64
     chp_fuel_cost_escalation_pct::Float64
+    generator_fuel_cost_escalation_pct::Float64
     offtaker_tax_pct::Float64
     offtaker_discount_pct
```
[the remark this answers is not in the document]
NLaws: nope thanks!

src/core/scenario.jl
NLaws: I see. No there is no need to update defaults for any item that is not modeled in the Julia package. The Julia package does not affect what is saved for inputs in the API database.

src/results/generator.jl
NLaws: "average" to me implies that we had more than one value that we averaged over the analysis years, but we only have one value and it is the same every year. so technically it is the average - I just find it misleading. we can leave it.

src/core/utils.jl
NLaws: depreciation starts at the end of the first year  (assuming that all capital costs are paid at the beginning of the first year).

src/core/utils.jl
NLaws: it looks like you commented out these lines for converting scalars to Float64 (that are now handled as Real types). Can we remove the lines?

src/core/reopt_multinode.jl
```diff
@@ -84,6 +84,12 @@ function add_variables!(m::JuMP.AbstractModel, ps::AbstractVector{REoptInputs{T}
 
         ex_name = "TotalPerUnitProdOMCosts"*_n
 		m[Symbol(ex_name)] = 0
+
+		ex_name = "TotalPerUnitHourOMCosts"*_n
+		m[Symbol(ex_name)] = 0 # TODO: Is this correct?
```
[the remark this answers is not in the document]
NLaws: `TotalPerUnitProdOMCosts` is just initialized as zero for MPC scenarios. It is added to [here](https://github.com/NREL/REopt.jl/blob/ce7b79ee9804ba12e8d43ba664f36fee8a24f333/src/mpc/model.jl#L188). So I think that `TotalPerUnitHourOMCosts` should be treated the same as `TotalPerUnitProdOMCosts`. I am not familiar with `TotalCHPStandbyCharges` but it looks like a monthly charge and we currently assume that MPC models have a horizon shorter than a month - i.e. we don't model multiple months. So I don't think that we can model `TotalCHPStandbyCharges` in the MPC model.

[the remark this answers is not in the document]
adfarth: Removed `TotalPerUnitHourOMCosts` from `reopt_multinode.jl` as well in https://github.com/NREL/REopt.jl/pull/24/commits/44599d39cf877a00b44c2df350bd3432ecf121ab. This cost is only for CHP currently. The `TotalPerUnitProdOMCosts` for MPC currently only account for generator costs, not CHP. More changes are thus needed to incorporate CHP into mpc.

adfarth (commented):
addressed remaining comments

src/core/financial.jl
NLaws: ah I missed that detail.

src/core/utils.jl
NLaws: That looks correct when comparing to the MACRS depreciation schedule https://github.com/NREL/REopt.jl/blob/ce7b79ee9804ba12e8d43ba664f36fee8a24f333/src/core/utils.jl#L129

Does the test value now not match the API value? Is that because the API assumes depreciation at the beginning of year 1?

src/core/reopt_multinode.jl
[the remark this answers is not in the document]
NLaws: I think that you have confused the multinode model with the MPC model? They are separate.

src/core/utils.jl
adfarth: The commented out code was removed in 3c364b32ce52e2637e1528a22c99aa224ed5203c. Apologies for not linking before!

src/core/utils.jl
adfarth: Yes, agreed regarding alignment with tax_savings_array. We do not currently apply depreciation savings to the `offgrid_other_capital_costs` in the API. This is a new change that was suggested by Dan and Doug.

src/core/reopt_multinode.jl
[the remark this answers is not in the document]
NLaws: I will address adding CHP to the multinode and mpc models

NLaws (approved):

PR #26, opened 2022-03-07
Cooling load

Bill-Becker:
This PR makes sure REopt.jl properly handles all the different ways cooling loads can be input. Tests from the API were migrated over. Some heating load related tests were also migrated. There are sections of the "heating and cooling energy balance" test that are commented out, to be uncommented when AbsorptionChiller, and Hot and Cold Storage results are added.

### DIFF
--- a/src/core/doe_commercial_reference_building_loads.jl
+++ b/src/core/doe_commercial_reference_building_loads.jl
@@ -115,7 +115,7 @@ end
 Scale a normalized Commercial Reference Building according to inputs provided and return the 8760.
 """
 function built_in_load(type::String, city::String, buildingtype::String, 
-    year::Int, annual_energy<:R, monthly_energies::AbstractArray{<:Real,1}
+    year::Int, annual_energy::R, monthly_energies::AbstractArray{<:Real,1}
     ) where {R <: Real}
 
     @assert type in ["electric", "domestic_hot_water", "space_heating", "cooling"]
--- a/src/core/electric_utility.jl
+++ b/src/core/electric_utility.jl
@@ -58,3 +58,5 @@ Base.@kwdef struct ElectricUtility
     net_metering_limit_kw::Real = 0
     interconnection_limit_kw::Real = 1.0e9
 end
+
+# TODO @warn when outage_start/end_timesteps not empty and outage_start/end_timestep is provided


src/core/heating_cooling_loads.jl
```diff
@@ -202,17 +202,17 @@ end
         blended_doe_reference_percents::Array{<:Real,1} = Real[],
         annual_tonhour::Union{Real, Nothing} = nothing,
         monthly_tonhour::Array{<:Real,1} = Real[],
-        fuel_loads_ton_per_hour::Array{<:Real,1} = Real[],
+        thermal_loads_ton::Array{<:Real,1} = Real[],
```
NLaws, line 205: should we append `_per_time_step` to `thermal_loads_ton`? (or is it assumed to be per hour and so we append `_per_hour`?)

src/core/heating_cooling_loads.jl
```diff
@@ -253,32 +253,50 @@ struct CoolingLoad
...
+                for (i, building) in enumerate(blended_doe_reference_names)
+                    default_fraction = get_default_fraction_of_total_electric(city, building, latitude, longitude, 2017)
+                    modified_fraction = default_fraction * blended_doe_reference_percents[i]
+                    if length(site_electric_load_profile) > 8784
+                        modified_fraction = repeat(modified_fraction, inner=time_steps_per_hour / (length(site_electric_load_profile)/8760))
+                        @info "Repeating cooling electric load in each hour to match the time_steps_per_hour."
+                    end
+                    loads_kw += site_electric_load_profile .* modified_fraction
+                end
+                println("Total cooling fraction = ", sum(loads_kw) / sum(site_electric_load_profile))
```
NLaws, line 285: is this a debugging `println`?

src/core/heating_cooling_loads.jl
```diff
@@ -317,7 +335,16 @@ struct CoolingLoad
...
     end
+end
+
+function get_default_fraction_of_total_electric(city, doe_reference_name, latitude, longitude, year)
+    crb_total_elec_loads_kw = BuiltInElectricLoad(city, doe_reference_name, latitude, longitude, year)
 
+    crb_cooling_elec_loads_kw = BuiltInCoolingLoad(city, doe_reference_name, latitude, longitude, year)
+    
+    default_fraction_of_total_electric_profile = crb_cooling_elec_loads_kw ./
+                                                    max(crb_total_elec_loads_kw, repeat([1.0E-6], length(crb_total_elec_loads_kw)))
```
NLaws, line 346: I think that you want `max.` here?
```julia
julia> max([1,2,3],[0,3,4])
3-element Vector{Int64}:
 1
 2
 3
julia> max.([1,2,3],[0,3,4])
3-element Vector{Int64}:
 1
 3
 4
```

src/core/heating_cooling_loads.jl
```diff
@@ -973,6 +1003,9 @@ function BuiltInSpaceHeatingLoad(
     if isnothing(annual_mmbtu)
         annual_mmbtu = spaceheating_annual_mmbtu[city][buildingtype]
     end
+    if length(monthly_mmbtu) == 12
+        monthly_mmbtu = convert(Vector{Real}, monthly_mmbtu)
```
NLaws, line 1007: why do we need this conversion?

src/core/scenario.jl
```diff
@@ -254,7 +254,16 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
 
     max_cooling_demand_kw = 0
     if haskey(d, "CoolingLoad") && !haskey(d, "FlexibleHVAC")
-        add_doe_reference_names_from_elec_to_thermal_loads(d["ElectricLoad"], d["CoolingLoad"])
+        assign_elec_doe_reference_name = true
+        cooling_keys_exceptions = ["annual_fraction", "monthly_fraction", "loads_fraction"]
```
NLaws, line 258: are all of these fractions the fraction of the electric load that is assumed to be cooling load? it would be helpful to put it in the name if so, maybe append `_of_electric_load` and make the monthly and per time step values plural. For example:
```
annual_fraction_of_electric_load
monthly_fractions_of_electric_load
per_time_step_fractions_of_electric_load
```
We should also document how these are used in the `CoolingLoad` doc string. As a user, for example, I would want to know if the cooling load fraction is removed from the electric load or not.

NLaws (changes requested):
Please see the inline comments.

src/core/heating_cooling_loads.jl
Bill-Becker: @NLaws `ton` (refrigeration ton) is a unit of **power**, so it's equivalent to kW where we don't need the `_per_time_step`

src/core/heating_cooling_loads.jl
Bill-Becker: Good call, I changed it.

src/core/heating_cooling_loads.jl
Bill-Becker: I found that I did from a test when I input `monthly_mmbtu`. It was not letting me assign Vector{Float64} (type converted from input file) to Vector{Real} which is the type of `monthly_mmbtu`.

src/core/scenario.jl
Bill-Becker: I like that, so I changed it. I'll update the doc string to clarify.

Bill-Becker: It looks like cooling electric is **not** currently subtracted off of the total electric load, so I'll work on fixing that.

Bill-Becker: I addressed the naming, doc string, and subtracting the cooling electric in the load balance constraints in the last two commits. Waiting to see if I broke anything from the tests.

src/core/scenario.jl
```diff
@@ -254,7 +254,16 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
 
     max_cooling_demand_kw = 0
     if haskey(d, "CoolingLoad") && !haskey(d, "FlexibleHVAC")
-        add_doe_reference_names_from_elec_to_thermal_loads(d["ElectricLoad"], d["CoolingLoad"])
+        assign_elec_doe_reference_name = true
```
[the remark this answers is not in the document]
Bill-Becker: I modified as you described, moving up the "_fraction_of_electric" elseifs in `CoolingLoad`, and then we do this assignment of `doe_reference_name` unconditionally but it doesn't matter if one of those fractions are input.

src/constraints/load_balance.jl
```diff
@@ -52,7 +53,9 @@ function add_elec_load_balance_constraints(m, p; _n="")
                 + m[Symbol("dvCurtail"*_n)][t, ts] 
             for t in p.techs.elec)
             + sum(m[Symbol("dvGridToStorage"*_n)][b, ts] for b in p.s.storage.types.elec)
+            + sum(m[Symbol("dvThermalProduction"*_n)][t, ts] / p.cop[t] for t in p.techs.cooling)
             + p.s.electric_load.loads_kw[ts]
+            - p.s.cooling_load.loads_kw_thermal[ts] / EXISTING_CHILLER_COP
```
NLaws, line 58: Should the subtraction of `cooling_load.loads_kw_thermal` from the electric load balance only happen when using a DoE reference profile? (What if a user provides their own `electric_load.loads_kw` for example, but wants to use DoE `cooling_load`? Or provides both the `electric_load.loads_kw` and `cooling_load`?)

NLaws:
@Bill-Becker there are three items to address:
1. if you point me to a test that fails without the type conversion for `monthly_mmbtu` I will make the typing more robust
2. some TODOs for you in heating_cooling_loads.jl
3. question on subtraction of cooling_load.loads_kw_thermal in electric load balance.

src/core/heating_cooling_loads.jl
Bill-Becker: Testset `Heating and cooling inputs` where I assign `monthly_mmbtu` in line 695.

Bill-Becker:
> @Bill-Becker there are three items to address:
> 
> 1. if you point me to a test that fails without the type conversion for `monthly_mmbtu` I will make the typing more robust
I believe you addressed this.
> 2. some TODOs for you in heating_cooling_loads.jl
I went through and deleted all the relevant TODO's.
> 3. question on subtraction of cooling_load.loads_kw_thermal in electric load balance.
Yes, we always have to subtract the cooling load when explicitly modeling/inputting cooling load. However, when cooling load is not explicitly modeled it is inherently assumed to be part of the electric load, and the cooling load is set to zeros.

@NLaws see above responses, and I made some changes to revert back to the API-way of assigning the COP of the `ExistingChiller` which is conditional on the max value of the `CoolingLoad.loads_ton_thermal` (and `max_thermal_factor_on_peak_load`). This conditional COP was developed in the AMO CHP project and peer reviewed/approved by AMO, so I do not want to change the functionality/logic here. If the user has their own `ExistingChiller.cop` and inputs a `doe_reference_name` with or without an `annual_tonhour` or `monthly_tonhour`, this does not actually change the _thermal_ energy cooling load, but it does change the electric consumption to serve that cooling load. I do realize that the "raw data" of CRB cooling load is electric energy, and therefore the conditional logic for COP versus peak thermal load makes the conversion of electric to thermal dependent on the profile for each of the buildings, but that logic **should** be there for that conversion (we don't know the actual COP for each of the buildings from the Energy Plus model data we got, and we are rely on an ASHRAE guidebook referenced in the User Manual which is reflected in that logic).

NLaws (approved):
good to merge once tests pass

PR #28, opened 2022-03-10
Add emissions capabilities

hdunham:
addresses issue #4


src/core/reopt.jl
```diff
@@ -66,6 +66,9 @@ end
 Solve the model using a `Scenario` or `BAUScenario`.
 """
 function run_reopt(m::JuMP.AbstractModel, s::AbstractScenario)
+	if s.site.CO2_emissions_reduction_min_pct > 0.0 || s.site.CO2_emissions_reduction_max_pct < 1.0
+		@error "To constrain CO2 emissions reduction min or max percentages, the optimal and business as usual scenarios must be run in parallel. Use a version of run_reopt() that takes an array of two models."
```
NLaws, line 70: @hdunham should this error be thrown? (And we could create an issue for adding BAU emissions so that a BAU Scenario is not required in the future).

src/core/reopt_inputs.jl
```diff
@@ -215,18 +239,26 @@ end
...
-    production_factor = DenseAxisArray{Float64}(undef, techs.all, 1:length(s.electric_load.loads_kw))
 
     # REoptInputs indexed on techs:
     max_sizes = Dict(t => 0.0 for t in techs.all)
     min_sizes = Dict(t => 0.0 for t in techs.all)
     existing_sizes = Dict(t => 0.0 for t in techs.all)
     cap_cost_slope = Dict{String, Any}()
     om_cost_per_kw = Dict(t => 0.0 for t in techs.all)
+    production_factor = DenseAxisArray{Float64}(undef, techs.all, 1:length(s.electric_load.loads_kw))
+    tech_renewable_energy_pct = DenseAxisArray([1.0 for _ in 1:length(techs.all)], techs.all)#Dict(t => 1.0 for t in techs.all)
```
NLaws, line 256: @hdunham were you thinking of changing all these DenseAxisArrays to Dicts (as shown in the comments)? I think that is a good idea since I think that DenseAxisArrays are unnecessary and add complexity

test/test_with_xpress.jl
```diff
@@ -32,6 +32,170 @@ using Random
 Random.seed!(42)  # for test consistency, random prices used in FlexibleHVAC tests
 
 
+@testset "Emissions" begin
```
NLaws, line 35: @hdunham the tests are taking over 5 hours! Can we reduce these tests?

hdunham: That's strange, they didn't take long on my machine. Will look into it.

src/core/reopt.jl
hdunham: Throwing it now. There's a lot of other instances of @error existing throughout the code that should probably be thrown as well.

src/core/generator.jl
```diff
@@ -66,6 +66,11 @@ function Generator(;
     production_incentive_max_benefit::Real = 1.0e9,
     production_incentive_years::Int = 0,
     production_incentive_max_kw::Real = 1.0e9,
+    fuel_renewable_energy_pct::Float64 = 0.0,
+    emissions_factor_lb_CO2_per_gal::Union{Missing,Float64} = 22.51,
```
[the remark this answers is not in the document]
hdunham: nope that's a remnant, removed

src/constraints/emissions_constraints.jl
```diff
@@ -0,0 +1,153 @@
...
+	end
+	if !isnothing(p.s.site.CO2_emissions_reduction_max_pct)
+		@constraint(m, MaxEmissionsReductionCon, m[:Lifecycle_Emissions_Lbs_CO2] >= (1-p.s.site.CO2_emissions_reduction_max_pct)*m[:Lifecycle_Emissions_Lbs_CO2_BAU])
+	end
+end
+
+### Year 1 Emissions Calculations
+function add_yr1_emissions_calcs(m,p)
+	# Components:
+	yr1_emissions_from_fuelburningtechs_lbs_CO2, yr1_emissions_from_fuelburningtechs_lbs_NOx, yr1_emissions_from_fuelburningtechs_lbs_SO2, yr1_emissions_from_fuelburningtechs_lbs_PM25 = calc_yr1_emissions_from_onsite_fuel(m,p; tech_array=p.techs.fuel_burning)
```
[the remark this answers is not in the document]
hdunham: Ah no just a remnant from when they were the sum of two variables (the result of calc_yr1_emissions_from_onsite_fuel() using fuel burning techs and boiler techs). That's how it was in the API but I realized this was double counting boiler techs.

src/constraints/generator_constraints.jl
```diff
@@ -42,21 +42,39 @@ end
...
 	# Generator must be on for nonnegative output
-	@constraint(m, [t in p.techs.gen, ts in p.time_steps],
-		m[:dvRatedProduction][t, ts] <= p.s.generator.max_kw * m[:binGenIsOnInTS][t, ts]
-	)
-	# Note: min_turn_down_pct is only enforced when off_grid_flag is true and in p.time_steps_with_grid, but not for grid outages for on-grid analyses
-	if p.s.settings.off_grid_flag 
-		@constraint(m, [t in p.techs.gen, ts in p.time_steps_without_grid],
-			p.s.generator.min_turn_down_pct * m[:dvSize][t] - m[:dvRatedProduction][t, ts] <=
-			p.s.generator.max_kw * (1 - m[:binGenIsOnInTS][t, ts])
+	if hasproperty(p, Symbol("max_sizes")) #regular REopt model, use p.max_sizes because it includes existing_kw
```
NLaws, line 45: It seems like we could simplify all these conditions by using `p.s.generator.max_kw`. We are using `p.s.generator.min_turn_down_pct` and we do not allow for more than one value in `p.techs.gen` so there seems to be no reason to have `t in p.techs.gen` in these constraints. Then we could just apply these constraints to `t="Generator"`. What do you think @hdunham ?

src/core/site.jl
```diff
@@ -61,6 +89,20 @@ struct Site
         roof_squarefeet::Union{Real, Nothing} = nothing,
         min_resil_time_steps::Int=0,
         mg_tech_sizes_equal_grid_sizes::Bool = true,
+        CO2_emissions_reduction_min_pct::Union{Float64, Nothing} = nothing,
```
NLaws, line 92: Please change the `Float64`s to `Real`s throughout (unless you explicitly do not want to allow integer inputs)

src/core/chp.jl
```diff
@@ -87,7 +88,11 @@ can_net_meter::Bool = false
 can_wholesale::Bool = false
 can_export_beyond_nem_limit::Bool = false
 can_curtail::Bool = false
-# emissions_factor_lb_CO2_per_mmbtu::Float64
+fuel_renewable_energy_pct::Union{Missing,Float64} = missing
+emissions_factor_lb_CO2_per_mmbtu::Union{Missing,Float64} = missing
```
NLaws, line 92: Just curious why you used `missing` here and `nothing` in other places?

src/core/electric_utility.jl
```diff
@@ -67,27 +88,101 @@ function ElectricUtility(;
...
+        if typeof(emissions_factor_series_lb_CO2_per_kwh) == Float64
+            emissions_series_dict["CO2"] = repeat([emissions_factor_series_lb_CO2_per_kwh], 8760*time_steps_per_hour)
+        elseif length(emissions_factor_series_lb_CO2_per_kwh) == 1
+            emissions_series_dict["CO2"] = repeat(emissions_factor_series_lb_CO2_per_kwh, 8760*time_steps_per_hour)
+        elseif length(emissions_factor_series_lb_CO2_per_kwh) / time_steps_per_hour ≈ 8760
+            emissions_series_dict["CO2"] = emissions_factor_series_lb_CO2_per_kwh
+        elseif isempty(emissions_factor_series_lb_CO2_per_kwh)
+            emissions_series_dict["CO2"] = emissions_series("CO2", region_abbr, time_steps_per_hour=time_steps_per_hour)
+        else
+            @error "Provided emissions_factor_series_lb_CO2_per_kwh does not match the time_steps_per_hour."
```
NLaws, line 134: do all of these `@error`s in the `ElectricUtility` constructor need to be thrown (i.e. will downstream code break)? We could instead take a default action like setting the emissions factors to zero or use the `emission_series` look-up and `@warn` instead. Not sure which would be more appropriate.

src/core/electric_utility.jl
```diff
@@ -29,35 +29,56 @@
...
+```
 !!! note Outage indexing begins at 1 (not 0) and the outage is inclusive of the outage end time step. 
     For instance, to model a 3-hour outage from 12AM to 3AM on Jan 1, outage_start_time_step = 1 and outage_end_time_step = 3.
     To model a 1-hour outage from 6AM to 7AM on Jan 1, outage_start_time_step = 7 and outage_end_time_step = 7.
 
 """
- struct ElectricUtility
+mutable struct ElectricUtility
+    emissions_region::String
+    distance_to_emissions_region_meters::Real
```
NLaws, line 69: the value for `distance_to_emissions_region_meters` can be `nothing` (returned from the `region_abbreviation` function), which would break here and in the results/electric_utility.jl. Should there be a default high number for the distance? We could return it from the `region_abbreviation` function.

src/core/electric_utility.jl
```diff
@@ -101,3 +196,121 @@ function ElectricUtility(;
...
+        "SW" => "Southwest",
+        "TX" => "Texas",
+        "WMW" => "Upper Midwest",
+        "HI" => "Hawaii (except Oahu)",
+        "HI-Oahu" => "Hawaii (Oahu)"
+    )
+    try
+        return lookup[region_abbr]
+    catch
+        return "None"
```
NLaws, line 219: maybe "not found" or `missing` instead of "None"? how will downstream code handle this "None"?

src/core/financial.jl
```diff
@@ -48,8 +48,21 @@ function Financial(;
     microgrid_upgrade_cost_pct::Real = off_grid_flag ? 0.0 : 0.3,
     macrs_five_year::Array{Float64,1} = [0.2, 0.32, 0.192, 0.1152, 0.1152, 0.0576],  # IRS pub 946
     macrs_seven_year::Array{Float64,1} = [0.1429, 0.2449, 0.1749, 0.1249, 0.0893, 0.0892, 0.0893, 0.0446],
+    CO2_cost_per_tonne::Float64 = 51.0,
```
NLaws, line 51: should we replace all `tonne`s with `metric_ton`? Or is `tonne` commonly used for pollutant costs @adfarth?

src/core/financial.jl
```diff
@@ -134,8 +208,245 @@ struct Financial
...
             offgrid_other_annual_costs
         )
     end
-end
\ No newline at end of file
+end
+
+
+# Convert from 2010$ to 2020$ (source: https://www.in2013dollars.com/us/inflation/2010?amount=100)
+const USD_2010_to_2020 = 1.246
```
NLaws, line 230: This value is only used in `easiur_costs`. Why does it need to be a `const`?

src/core/scenario.jl
```diff
@@ -111,22 +111,77 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
...
 
+    #Handle missing emissions inputs (due to failed lookup and not provided by user)
+    #TODO: condition below code on: if site.off_grid == false
+    if isnothing(electric_utility.emissions_factor_series_lb_CO2_per_kwh)
+        if (!isnothing(site.CO2_emissions_reduction_min_pct) || 
+            !isnothing(site.CO2_emissions_reduction_max_pct) || 
+            settings.include_climate_in_objective)
+            throw(@error "To include CO2 costs in the objective function or enforce emissions reduction constraints, you must either enter custom health emissions factors or a site location within the continental U.S.")
+        else
+            electric_utility.emissions_factor_series_lb_CO2_per_kwh = zeros(8760*settings.time_steps_per_hour)
```
NLaws, line 152: We should do this `electric_utility.emissions_factor_series_lb_CO2_per_kwh = zeros(8760*settings.time_steps_per_hour)` in the `ElectricUtility` constructor and keep the `ElectricUtility` immutable.

src/core/scenario.jl
```diff
@@ -111,22 +111,77 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
...
+        end
+    end
+    for emissions_type in ["NOx", "SO2", "PM25"]
+        if isnothing(getproperty(electric_utility, Symbol("emissions_factor_series_lb_$(emissions_type)_per_kwh")))
+            if settings.include_health_in_objective
+                throw(@error "To include health costs in the objective function, you must either enter custom health emissions factors or a site location within the continental U.S.")
+            else
+                setproperty!(electric_utility,
```
NLaws, line 160: These defaults of zero emissions factors should be done in the `ElectricUtility` constructor. Then we can keep the `ElectricUtility` immutable and we won't need the `Union` types.

src/core/scenario.jl
```diff
@@ -111,22 +111,77 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
...
+    missing_health_inputs = false
+    for emissions_type in ["NOx", "SO2", "PM25"]
+        for health_input in [
+            "$(emissions_type)_grid_cost_per_tonne",
+            "$(emissions_type)_onsite_fuelburn_cost_per_tonne",
+            "$(emissions_type)_cost_escalation_pct"
+        ]
+            if ismissing(getproperty(financial, Symbol(health_input)))
+                missing_health_inputs = true
+                setproperty!(financial, Symbol(health_input), 0)
```
NLaws, line 177: why not default all of the `Financial` health costs to zero and avoid all this checking in the `Scenario` constructor? Would be nice to keep the `Financial` struct immutable for code integrity.

src/core/scenario.jl
```diff
@@ -400,3 +458,57 @@ function add_doe_reference_names_from_elec_to_thermal_loads(elec::Dict, thermal:
         end
     end
 end
+
+"""
+    function set_missing_emissions_factors_to_fuel_defaults(tech::AbstractFuelBurningTech)
+
+Populate emission factors in tech based on its fuel_type field.
+"""
+function set_missing_emissions_factors_to_fuel_defaults(tech::AbstractFuelBurningTech)
```
NLaws, line 467: Can we use this function in the tech constructors so that we do not have to make the techs mutable?

src/core/utils.jl
```diff
@@ -163,7 +172,7 @@ function dictkeys_tosymbols(d::Dict)
             "monthly_totals_kwh",
             "prod_factor_series", 
             "monthly_energy_rates", "monthly_demand_rates",
-            "wholesale_rate", "blended_doe_reference_percents",
```
NLaws, line 166: why was "wholesale_rate" removed?

src/core/utils.jl
```diff
@@ -34,7 +34,16 @@ function annuity(years::Int, rate_escalation::Real, rate_discount::Real)
         i.e. it is a geometric sum of (1+rate_escalation)^n / (1+rate_discount)^n
         for n = 1, ..., years
     """
-    x = (1 + rate_escalation) / (1 + rate_discount)
+    return annuity_two_escalation_rates(years, rate_escalation, 0.0, rate_discount)
```
NLaws, line 37: This is very confusing and there seems to be no need for calling `annuity_two_escalation_rates` from `annuity` since the second escalation rate is always zero. Please revert the change in `annuity`.

src/core/utils.jl
```diff
@@ -34,7 +34,16 @@ function annuity(years::Int, rate_escalation::Real, rate_discount::Real)
...
+end
+
+function annuity_two_escalation_rates(years::Int, rate_escalation1::Float64, rate_escalation2::Float64, rate_discount::Float64)
+    """
+        this formulation assumes cost growth in first period
+        i.e. it is a geometric sum of (1+rate_escalation)^n / (1+rate_discount)^n
```
NLaws, line 43: this doc string is incorrect

src/mpc/structs.jl
```diff
@@ -242,6 +242,42 @@ function MPCElectricTariff(d::Dict)
 end
 
 
+"""
+MPCElectricUtility
```
NLaws, line 246: If we had defaults of zeros for all the new `ElectricUtility` fields would we not need a separate `MPCElectricUtility`? It would be great for maintainability if we could reuse as many structs as possible in the MPC code.

src/results/financial.jl
```diff
@@ -133,7 +136,7 @@ function add_financial_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _
         r["offgrid_microgrid_lcoe_dollars_per_kwh"] = round(r["lcc"] / pwf / value(LoadMet), digits=4)
     end
 
-    d["Financial"] = Dict(k => round(v, digits=4) for (k,v) in r)
+    d["Financial"] = Dict{String, Union{Nothing,Float64}}(k => round(v, digits=4) for (k,v) in r)
```
NLaws, line 139: The `Nothing` type is not serializable so if we are are returning `nothing`s in the results the Julia package will not work in the API.  Can any results that are `nothing` be zero instead?

src/results/results.jl
```diff
@@ -167,6 +208,34 @@ function combine_results(p::REoptInputs, bau::Dict, opt::Dict, bau_scenario::BAU
...
+    else
+        breakeven_cost_denominator = p.pwf_emissions_cost["CO2_grid"] * (
+            bau["ElectricUtility"]["year_one_emissions_tCO2"] - opt["ElectricUtility"]["year_one_emissions_tCO2"]
+        ) + p.pwf_emissions_cost["CO2_onsite"] * (
+            bau["Site"]["year_one_emissions_from_fuelburn_tCO2"] - opt["Site"]["year_one_emissions_from_fuelburn_tCO2"] 
+        )
+        if breakeven_cost_denominator != 0.0
+            opt["Financial"]["breakeven_cost_of_emissions_reduction_per_tCO2"] = -1 * npv_without_modeled_climate_costs / breakeven_cost_denominator
+        else
+            opt["Financial"]["breakeven_cost_of_emissions_reduction_per_tCO2"] = nothing
```
NLaws, line 235: Maybe just exclude the "breakeven_cost_of_emissions_reduction_per_tCO2" key instead of returning a `nothing` value?

test/test_with_xpress.jl
```diff
@@ -32,6 +32,170 @@ using Random
...
+        yr1_total_emissions_calced_tCO2 = yr1_fuel_emissions_tCO2_out + yr1_grid_emissions_tCO2_out 
+        @test year_one_emissions_tCO2_out ≈ yr1_total_emissions_calced_tCO2 atol=1e-1
+        if !isnothing(results["Financial"]["breakeven_cost_of_emissions_reduction_per_tCO2"])
+            @test results["Financial"]["breakeven_cost_of_emissions_reduction_per_tCO2"] >= 0.0
+        end
+        
+        #commented out values are using levelization factors matching API
+        if i == 1
+            @test results["PV"]["size_kw"] ≈ 45.9 atol=1e-1#46.3 atol=1e-1
+            @test results["Wind"]["size_kw"] ≈ 10.0
```
NLaws, line 90: can we remove Wind from these tests? When developing we were often hitting the WindToolKit API request limit due to many Wind tests (which we since reduced to one Wind test).

src/core/site.jl
```diff
@@ -73,7 +115,17 @@ struct Site
         if length(invalid_args) > 0
             error("Invalid argument values: $(invalid_args)")
         end
+        if !isnothing(renewable_electricity_max_pct) && (renewable_electricity_min_pct > renewable_electricity_max_pct)
+            error("renewable_electricity_min_pct must be less than renewable_electricity_max_pct")
```
NLaws, line 119: Should this error be thrown?

src/REopt.jl
```diff
@@ -63,6 +63,9 @@ using Shapefile
 using ArchGDAL
 using Roots: fzero  # for IRR
 global hdl = nothing
+using HDF5, JLD
+using CSV, DataFrames
```
NLaws, line 67: I have found DelimitedFiles to be more efficient at reading in CSV files (if you just need the data from the CSV file).  What are we using DataFrames for in the package? What is JLD used for in the package?

src/core/financial.jl
adfarth: I think `tonne` explicitly equates to `metric ton` and is used often as a differentiator from `ton`. I therefore think it's fine to leave as `tonne`.

src/core/utils.jl
```diff
@@ -34,7 +34,16 @@ function annuity(years::Int, rate_escalation::Real, rate_discount::Real)
...
+        for n = 1,..., analysis_period
+    """
+    x = (1 + rate_escalation1 + rate_escalation2 + rate_escalation1 * rate_escalation2) / (1 + rate_discount)
```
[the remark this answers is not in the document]
adfarth: @NLaws this is the formula suggested by Dan for an annuity with two escalation rates. In an Excel example calculation we show that this formula results in 0% difference from the "True NPV" of annual emissions costs (can share if needed).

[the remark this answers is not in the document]
NLaws: oh I see now:
```math
(1+g1)^y * (1+g2)^y = (1+g1+g2+g1*g2)^y
```
please add something like this to the doc string

src/core/utils.jl
NLaws: it should show how the two escalation (growth) rates are accounted for, like:

(1+g1)^y * (1+g2)^y = (1+g1+g2+g1*g2)^y

src/constraints/emissions_constraints.jl
```diff
@@ -0,0 +1,156 @@
...
+	end
+	if !isnothing(p.s.site.CO2_emissions_reduction_max_pct)
+		@constraint(m, MaxEmissionsReductionCon, m[:Lifecycle_Emissions_Lbs_CO2] >= (1-p.s.site.CO2_emissions_reduction_max_pct)*m[:Lifecycle_Emissions_Lbs_CO2_BAU])
+	end
+end
+
+### Year 1 Emissions Calculations
+function add_yr1_emissions_calcs(m,p)
+	# Components:
+	m[:yr1_emissions_onsite_fuel_lbs_CO2], m[:yr1_emissions_onsite_fuel_lbs_NOx],
```
adfarth, line 43: @hdunham @NLaws do you think it would be worth indexing all of these emissions model vars and associated parameters on pollutant, rather than have each one individually named? Pros: significantly decrease the length of code. Cons: will take some time. 
e.g.
- `p.tech_emissions_factors_CO2[t]` --> `p.tech_emissions_factors[p][t]` where p in P, set [CO2, NOx, SO2, PM2.5] 
- `m[:yr1_emissions_onsite_fuel_lbs_CO2]` --> `m[:yr1_emissions_onsite_fuel_lbs][p]`

src/constraints/emissions_constraints.jl
```diff
@@ -0,0 +1,156 @@
...
+	m[:yr1_emissions_offset_from_elec_exports_lbs_SO2], m[:yr1_emissions_offset_from_elec_exports_lbs_PM25] = 
+	calc_yr1_emissions_offset_from_elec_exports(m,p;tech_array=p.techs.elec)
+
+	# Total:
+	if p.s.site.include_exported_elec_emissions_in_total
+		include_exported_elec_emissions_in_total = 1
+	else
+		include_exported_elec_emissions_in_total = 0
+	end
+	# "net_if_selected" indicates that emissions totals will be net of (subtract) exports IF the user sets "site.include_exported_elec_emissions_in_total" to true
```
adfarth, line 61: I thought this comment was confusing so I clarified in https://github.com/NREL/REopt.jl/pull/28/commits/9526eb041a7842fc7dd98a3d9a416558a01016cc

src/constraints/emissions_constraints.jl
```diff
@@ -0,0 +1,156 @@
...
+	m[:yr1_emissions_from_elec_grid_net_if_selected_lbs_PM25] = m[:yr1_emissions_from_elec_grid_lbs_PM25] - include_exported_elec_emissions_in_total*m[:yr1_emissions_offset_from_elec_exports_lbs_PM25]
+
+	m[:EmissionsYr1_Total_LbsCO2] = m[:yr1_emissions_onsite_fuel_lbs_CO2] + m[:yr1_emissions_from_elec_grid_net_if_selected_lbs_CO2]
+	m[:EmissionsYr1_Total_LbsNOx] = m[:yr1_emissions_onsite_fuel_lbs_NOx] + m[:yr1_emissions_from_elec_grid_net_if_selected_lbs_NOx]
+	m[:EmissionsYr1_Total_LbsSO2] = m[:yr1_emissions_onsite_fuel_lbs_SO2] + m[:yr1_emissions_from_elec_grid_net_if_selected_lbs_SO2]
+	m[:EmissionsYr1_Total_LbsPM25] = m[:yr1_emissions_onsite_fuel_lbs_PM25] + m[:yr1_emissions_from_elec_grid_net_if_selected_lbs_PM25]
+
+end
+
+function calc_yr1_emissions_from_onsite_fuel(m,p; tech_array=p.techs.fuel_burning) # also run this with p.techs.boiler
```
adfarth, line 74: @hdunham  should we create a new issue to make sure we don't forget to include boiler when it's ready?

src/constraints/renewable_energy_constraints.jl
```diff
@@ -0,0 +1,150 @@
...
+	if !isnothing(p.s.site.renewable_electricity_max_pct)
+		@constraint(m, MaxREElecCon, m[:AnnualREEleckWh] <= p.s.site.renewable_electricity_max_pct*m[:AnnualEleckWh])
+	end
+end
+
+# Renewable electricity calculation
+function add_re_elec_calcs(m,p)
+	
+	# User-selected RE electricity accounting methodology:
+	if p.s.site.include_exported_renewable_electricity_in_total
```
adfarth, line 45: I don't think this type conversion is necessary (bools are treated as a subtype of integer with true=1 and false=0 by default). So, I think this if statement could be replaced with: 
`include_exported_re_elec_in_total = p.s.site.include_exported_renewable_electricity_in_total` (to slightly shorten the var name) 
or, perhaps better, we could just use `p.s.site.include_exported_renewable_electricity_in_total` in the below equations

src/constraints/renewable_energy_constraints.jl
```diff
@@ -0,0 +1,177 @@
...
+# OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+# OF THE POSSIBILITY OF SUCH DAMAGE.
+# *********************************************************************************
+
+"""
+	add_re_elec_constraints(m,p)
+
+Function to add minimum and/or maximum renewable electricity (as percentage of load) constraints, if specified by user.
+
+!!! note
```
adfarth, line 36: @hdunham  could you please review these comments I added regarding treatment of RE calcs with single vs multiple outages?

src/core/bau_inputs.jl
```diff
@@ -46,6 +46,12 @@ function BAUInputs(p::REoptInputs)
     cop = Dict(t => 0.0 for t in techs.cooling)
     thermal_cop = Dict{String, Float64}()
     production_factor = DenseAxisArray{Float64}(undef, techs.all, p.time_steps)
+    tech_renewable_energy_pct = DenseAxisArray([p.tech_renewable_energy_pct[t] for t in techs.all], techs.all)
+    # !!! note: tech_emissions_factors in lb / kWh (gets multiplied by kWh of fuel burned, not kWh electricity consumption, ergo the use of the HHV instead of fuel slope)
+    tech_emissions_factors_CO2 = DenseAxisArray([p.tech_emissions_factors_CO2[t] for t in techs.all], techs.all)
```
adfarth, line 51: same comment as previous about potentially indexing emissions factors on pollutant (and tech) rather than having unique variable names for each

src/core/bau_inputs.jl
```diff
@@ -115,15 +121,18 @@ function BAUInputs(p::REoptInputs)
     t0, tf = p.s.electric_utility.outage_start_time_step, p.s.electric_utility.outage_end_time_step
     if tf > t0 && t0 > 0
         original_crit_lds = copy(p.s.electric_load.critical_loads_kw)
-        update_bau_outage_outputs(bau_scenario, original_crit_lds, t0, tf, production_factor)
+        generator_fuel_use_gal = update_bau_outage_outputs(bau_scenario, original_crit_lds, t0, tf, production_factor)
```
adfarth, line 124: Does the `generator_fuel_use_gal` now only apply for modeling single outages (not multiple with probabilities)?

src/core/bau_inputs.jl
```diff
@@ -164,15 +175,134 @@ function BAUInputs(p::REoptInputs)
...
+end
+
+"""
+    setup_bau_emissions_inputs(p::REoptInputs, s_bau::BAUScenario, generator_fuel_use_gal::Float64)
+
+Pre-processing of the BAU emissions to use in determining emissions reductions in the optimal case
+Include BAU grid emissions, existing backup generator emissions, boiler emissions
+Update the `bau_(grid_)emissions_` values in s.site and s_bau.site
+
+Note if existing generation does not sustain a simulated outage, the BAU load and therefore emissions
```
adfarth, line 217: Does this apply for multi-outage situations or just single outages?

src/core/bau_inputs.jl
```diff
@@ -164,15 +175,134 @@ function BAUInputs(p::REoptInputs)
...
+    bau_emissions_lb_CO2_per_year = 0
+    bau_emissions_lb_NOx_per_year = 0
+    bau_emissions_lb_SO2_per_year = 0
+    bau_emissions_lb_PM25_per_year = 0
+
+    ## Grid emissions
+
+    # This function is called after ajust_load_profile() (makes load_kw native if not already) 
+    # so need to calculate net version of loads_kw by removing existing PV to get load served by the grid.
+    # TODO: Should load_profile have loads_kw and loads_kw_net fields, instead of updating loads_kw? Would be easier
```
adfarth, line 231: Should we discuss this and decide?

src/core/bau_inputs.jl
```diff
@@ -164,15 +175,134 @@ function BAUInputs(p::REoptInputs)
...
+    # Must account for levelization factor to align with how PV is modeled in REopt:
+    # Because we only model one year, we multiply the "year 1" PV production by a levelization_factor
+    # that accounts for the PV capacity degradation over the analysis_years. In other words, by
+    # multiplying the pv production_factor by the levelization_factor we are modeling the average pv production.
+    bau_grid_to_load = copy(s_bau.electric_load.loads_kw)
+    bau_grid_to_load_critical = copy(s_bau.electric_load.critical_loads_kw)
+    for pv in p.s.pvs if pv.existing_kw > 0
+        bau_grid_to_load .-= p.levelization_factor[pv.name] * pv.existing_kw * p.production_factor[pv.name, :].data
+    end end
+    for pv in p.s.pvs if pv.existing_kw > 0
```
adfarth, line 242: Can these two for-if statements be combined into one?

src/core/bau_inputs.jl
```diff
@@ -164,15 +175,134 @@ function BAUInputs(p::REoptInputs)
...
+    if "Generator" in p.techs.all
+        bau_emissions_lb_CO2_per_year += generator_fuel_use_gal * p.s.generator.emissions_factor_lb_CO2_per_gal
+        bau_emissions_lb_NOx_per_year += generator_fuel_use_gal * p.s.generator.emissions_factor_lb_NOx_per_gal
+        bau_emissions_lb_SO2_per_year += generator_fuel_use_gal * p.s.generator.emissions_factor_lb_SO2_per_gal
+        bau_emissions_lb_PM25_per_year += generator_fuel_use_gal * p.s.generator.emissions_factor_lb_PM25_per_gal
+    end
+
+    ## Boiler emissions
+    if "ExistingBoiler" in p.techs.all
+        for heat_type in ["space_heating", "dhw"] #TODO: are these heating types the same? can we get them from s?
```
adfarth, line 280: do we need to address this TODO?

src/core/electric_utility.jl
```diff
@@ -29,35 +29,56 @@
...
-
 ElectricUtility data struct with inner constructor:     
+        
 ```julia
 function ElectricUtility(;
-    outage_start_time_step::Int=0,  # for modeling a single outage, with critical load spliced into the baseline load ...
-    outage_end_time_step::Int=0,  # ... utiltity production_factor = 0 during the outage
-    allow_simultaneous_export_import::Bool = true,  # if true the site has two meters (in effect)
-    # variables below used for minimax the expected outage cost,
+    emissions_region::String = "",
```
adfarth, line 36: Should the user be able to provide an emissions region, as opposed to only being able to use the `region()` function to obtain this?

src/core/reopt_inputs.jl
```diff
@@ -215,18 +239,26 @@ end
 Create data arrays associated with techs necessary to build the JuMP model.
 """
 function setup_tech_inputs(s::AbstractScenario)
+    #TODO: new boiler, steam turbine, ghp
+    #TODO: create om_cost_per_kwh in here as well as om_cost_per_kw?
```
adfarth, line 243: @hdunham  should we add this as an issue?

src/core/scenario.jl
```diff
@@ -111,22 +111,77 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
...
-        electric_utility = ElectricUtility(; outage_start_time_step = 1, outage_end_time_step = settings.time_steps_per_hour * 8760) 
+        electric_utility = ElectricUtility(; outage_start_time_step = 1, 
+                                            outage_end_time_step = settings.time_steps_per_hour * 8760, 
+                                            latitude=site.latitude, longitude=site.longitude, 
+                                            time_steps_per_hour=settings.time_steps_per_hour
+                                        ) 
     end
 
+    #Handle missing emissions inputs (due to failed lookup and not provided by user)
+    #TODO: condition below code on: if site.off_grid == false
```
adfarth, line 145: Can we do this conditioning on off_grid now? (For off-grid, should not throw an error if an emissions reduction target is set or CO2 included in objective and the `emissions_factor_series_lb_CO2_per_kwh` is missing.

src/core/scenario.jl
```diff
@@ -111,22 +111,77 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
...
+                throw(@error "To include health costs in the objective function, you must either enter custom health emissions factors or a site location within the continental U.S.")
+            else
+                setproperty!(electric_utility, 
+                    Symbol("emissions_factor_series_lb_$(emissions_type)_per_kwh"), 
+                    zeros(8760*settings.time_steps_per_hour)
+                )
+            end
+        end
+    end
+    #TODO: allow grid costs to be missing if site.off_grid == true
```
adfarth, line 167: Can we do this now?

src/core/site.jl
adfarth: @NLaws are you suggesting we use `throw(@error " ") `? If so I agree. And @hdunham I think the error should be "less than or equal to..."

src/results/chp.jl
```diff
@@ -111,6 +111,7 @@ function add_chp_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _n="")
     r["year_one_chp_fuel_cost_before_tax"] = round(value(m[:TotalCHPFuelCosts] / p.pwf_fuel["CHP"]), digits=3)                
 	r["lifecycle_chp_fuel_cost_after_tax"] = round(value(m[:TotalCHPFuelCosts]) * p.s.financial.offtaker_tax_pct, digits=3)
 	#Standby charges and hourly O&M
+	r["lifecycle_chp_standby_cost"] = round(value(m[Symbol("TotalCHPStandbyCharges")]) * p.s.financial.offtaker_tax_pct, digits=0)
```
adfarth, line 114: I had removed this output here because it is now included in results/financial.jl (in order to have the full LCC breakdown there). If you think it should also be repeated here, please add "after_tax" to the output name.

src/results/site.jl
```diff
@@ -0,0 +1,107 @@
...
+	r["year_one_emissions_tSO2"] = round(value(m[:EmissionsYr1_Total_LbsSO2] * TONNE_PER_LB), digits=2)
+	r["year_one_emissions_tPM25"] = round(value(m[:EmissionsYr1_Total_LbsPM25] * TONNE_PER_LB), digits=2)
+
+	r["year_one_emissions_from_fuelburn_tCO2"] = round(value(m[:yr1_emissions_onsite_fuel_lbs_CO2] * TONNE_PER_LB), digits=2)
+	r["year_one_emissions_from_fuelburn_tNOx"] = round(value(m[:yr1_emissions_onsite_fuel_lbs_NOx] * TONNE_PER_LB), digits=2)
+	r["year_one_emissions_from_fuelburn_tSO2"] = round(value(m[:yr1_emissions_onsite_fuel_lbs_SO2] * TONNE_PER_LB), digits=2)
+	r["year_one_emissions_from_fuelburn_tPM25"] = round(value(m[:yr1_emissions_onsite_fuel_lbs_PM25] * TONNE_PER_LB), digits=2)
+
+	# Lifecycle emissions results at Site level
+	if !isnothing(p.s.site.bau_grid_emissions_lb_CO2_per_year)
```
adfarth, line 90: Should this just be `bau_emissions_lb_CO2_per_year` instead of grid emissions? Presumably someone could set the grid emissions to zero and just look at fuel burn emissions.

src/REopt.jl
rathod-b: I believe DataFrames and CSV were only being used in one place: emissions_series() in electricutility.jl. I have converted that function to use DelimitedFiles. will push it shortly.

Not sure about JLD?

hdunham: JLD was from a previous iteration of the code for loading the emissions data. I removed all these packages from the project.

src/constraints/emissions_constraints.jl
hdunham: I do like the sound of that. Not the highest priority though, so will probably merge without that and change later when funds aren't so tight. I don't think we would change that in the input dict, just how they are saved in the structs.

src/constraints/emissions_constraints.jl
hdunham: Haha agreed, I copied the comment from the API emissions code

src/constraints/renewable_energy_constraints.jl
hdunham: Good to know! If statement is how it was done in the API version.

src/core/bau_inputs.jl
hdunham: yes

src/core/bau_inputs.jl
hdunham: I don't remember adding this TODO. These are the correct heat load types though, and no we don't save a list of them in s, so I'll delete. We could maybe do something similar to the tech category lists we create, but I don't see enough value to justify spending time on it right now.

src/core/electric_utility.jl
hdunham: I don't think so. The user provides lat/long, which determined the emissions region.

src/constraints/emissions_constraints.jl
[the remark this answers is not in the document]
NLaws: ok could you please put them directly into the model dict?

src/REopt.jl
rathod-b: We might need to keep JLD and HDF5 as they are being used to load EASIUR data in Financial.jl -> get_EASIUR2005( ), unless we now have a different method of loading EASIUR data?

I found some other remaining modules which I have removed such as PolygonInbounds, Shapefile, and Proj4

data/avert/avert_102008.cpg
```diff
@@ -0,0 +1 @@
+ISO-8859-1
```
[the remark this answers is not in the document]
rathod-b: Addressed in [3f0bbcc](https://github.com/NREL/REopt.jl/commit/3f0bbcc6aac2b1e0762377ac8a364bfc7bf868d4). Also removed some unused GIS files

src/results/site.jl
hdunham: good catch!

src/core/electric_utility.jl
hdunham: I didn't know this. I guess we do need to add emissions_region as a site input then.

src/core/site.jl
```diff
@@ -40,11 +40,25 @@ function Site(;
     roof_squarefeet::Union{Real, Nothing} = nothing,
     min_resil_time_steps::Int=0,
     mg_tech_sizes_equal_grid_sizes::Bool = true,
+    CO2_emissions_reduction_min_pct::Union{Float64, Nothing} = nothing,
+    CO2_emissions_reduction_max_pct::Union{Float64, Nothing} = nothing,
+    bau_emissions_lb_CO2_per_year::Union{Float64, Nothing} = nothing,
```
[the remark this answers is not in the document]
hdunham: Actually I wasn't intending this to be a possible user input. These fields should be in the Site struct but not in its constructor. I'm a bit wary of users being able to provide BAU emissions because if they use a different emissions dataset to calculate them then it wouldn't be an apples to apples comparison. But at the same time, it could definitely be useful if used correctly and would allow for emissions reduction constraints without running the BAU model. Maybe a compromise is allowing them as user inputs in the Julia package but not int he API?

src/results/chp.jl
hdunham: I hadn't intentionally added it back in (I think this is how the merge of our changes was resolved), but I do think it would be good to have the duplicate here.

src/core/reopt_inputs.jl
hdunham: issue created

test/test_with_xpress.jl
rathod-b: Could the long solve times be caused due to REopt version? I tested "Thermal loads" testset and got the following solve times:

- in Emissions branch: 56 seconds for first try, then 15 seconds for each attempt.
- in Develop branch: 45 seconds for first try, then 9 seconds for each attempt afterwards

would it makes sense to merge develop into emissions?

src/core/electric_utility.jl
hdunham: Amanda and I talked, and the API doesn't provide this option, so I won't include it in the Julia package now either.

src/core/heating_cooling_loads.jl
```diff
@@ -94,7 +95,8 @@ struct DomesticHotWaterLoad
         end
 
         new(
-            loads_kw
+            loads_kw,
+            (sum(loads_kw)/time_steps_per_hour)/KWH_PER_MMBTU
```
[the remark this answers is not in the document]
hdunham: No not really because loads_kw isn't an input, it is calculated based on fuel_loads_mmbtu_per_hour or reference profile, annual_mmbtu, and monthly_mmbtu. If the user provides both, then fuel_loads_mmbtu_per_hour is used, so we could add a warning about that. But if the user provides doe_reference_name etc, then that is used to calculated loads_kw, so this should be the same value as provided.

src/constraints/renewable_energy_constraints.jl
hdunham: reviewed and revised

src/core/generator.jl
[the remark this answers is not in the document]
NLaws: @hdunham do you have some un-pushed commits? The `Missing` type is still in the function argument types and this doc string.

src/core/reopt.jl
NLaws: @hdunham do you have this file modified locally? I do not see the `throw` statement.

src/core/electric_utility.jl
```diff
@@ -101,3 +196,121 @@ function ElectricUtility(;
         )
     end
 end
+
+function region(region_abbr::String)
```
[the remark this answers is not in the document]
NLaws: done in https://github.com/NREL/REopt.jl/pull/28/commits/2186a21cd6710d77a48a67956b96f6a6ef5ba32f

src/core/electric_utility.jl
NLaws: I changed the default emission region to `nothing` because that is what is expected in the `emissions_series` function. However, the downstream code, specifically the code in `add_yr1_emissions_calcs` will fail if any emission series is `nothing`. We need to fix this by either throwing an error, not modeling emissions with a warning, or setting the particular emission series to zeros with a warning.

src/core/financial.jl
NLaws: addressed in https://github.com/NREL/REopt.jl/pull/28/commits/c61f79a4263245f6662a777f05362186c258450d

src/core/reopt_inputs.jl
NLaws: addressed in https://github.com/NREL/REopt.jl/pull/28/commits/e0d97c036284e70ae76b909c43daf48696b9a895

src/core/generator.jl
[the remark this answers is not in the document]
rathod-b: If this is still an open issue, should emission factors per gal for all 4 pollutant types default to 0.0 with type Real for chp, gen, existinbboiler techs and we can call set_missing_emissions_factors_to_fuel_defaults right after calling tech structs?

src/core/financial.jl
```diff
@@ -134,8 +208,243 @@ struct Financial
...
+    end
+    if dollar_year != 2010
+        try
+            adj = get(GDP_deflator, dollar_year, nothing) / get(GDP_deflator, 2010, nothing)
+            for (k, v) in ret_map
+                setindex!(ret_map, v .* adj, k)
+            end
+        catch e
+            @warn "Dollar year must be between 1980 to 2010"
+            return false
```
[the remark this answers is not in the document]
rathod-b: The three `false` returns can be removed because the function is not user accessible, so the conditions which lead to false return will not be met with how the code is written at the moment. I will remove them locally and test them

src/constraints/generator_constraints.jl
rathod-b: I have updated these constraints locally to just apply to "Generator". I have also simplified the code to first check if p is of type REoptInputs. If it is, then set a local variable gen_size to existing kw + max_kw. if it is not REoptInputs type then assume it is MPC type.

src/core/bau_inputs.jl
hdunham: It's handled by the try/catches in the functions that call get_EASIUR2005(), i.e. easiur_costs() and easiur_escalation_rates(). False is kind of a weird thing to return when it fails though, so I changed them to nothing. I'm also changing the warnings to errors because that's what they are.

src/constraints/generator_constraints.jl
hdunham: Also, I think it's good practice to write t in p.techs.gen, even if currently there can only be one generator. Makes the code more robust to future changes.

src/core/electric_utility.jl
hdunham: I will update so that the AVERT emissions series is used when the provided array has an invalid length, and just give a warning.

src/core/electric_utility.jl
hdunham: I changed the field type to accept nothing

src/core/electric_utility.jl
```diff
@@ -67,27 +88,101 @@ function ElectricUtility(;
...
-    )
+        interconnection_limit_kw::Real = 1.0e9,
+        latitude::Real,
+        longitude::Real,
+        time_steps_per_hour::Int = 1
+        )
+
+        region_abbr, meters_to_region = region_abbreviation(latitude, longitude)
+        emissions_region = region(region_abbr)
+        emissions_series_dict = Dict{String,Array{Float64}}()
```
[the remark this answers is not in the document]
hdunham: Yes, originally I had the emissions_series function returning nothing if there was a lookup error, and I had Union{Nothing,Array{<:Real,1}} as the type for emissions_factor_series_lb_<pollutant>_per_kwh in ElectricUtility to accommodate this. Then, in Scenario, I was checking to see if CO2 emissions were nothing and the user had specified a CO2 reduction % or set include_climate_in_objective to true, or if health emissions were nothing and the user had set include_health_in_objective to true. We have to error in these cases because we can't do what the user is asking for. If the user didn't specify these, then I was setting the emissions to all zeros and allowing the optimization to proceed. It looks like recently this error handling code in scenario.jl was deleted by @rathod-b (commit 6f15ad3) and Nothing was removed from the emissions factors types by you (commit f8959c1). If I revert these changes from you two and add Nothing to the type of emissions_series_dict, does that sufficiently address your concerns?

src/core/financial.jl
```diff
@@ -62,7 +75,7 @@ function Financial(;
         end
     ```
 """
-struct Financial
+mutable struct Financial
```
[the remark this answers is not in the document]
hdunham: I made Financial mutable because it was necessary for code that @rathod-b has since commented out in commit 6f15ad3. In Scenario we were checking whether any health emissions cost inputs were missing and the user had set include_health_in_objective to true, because if so we need to throw an error. He moved this into the Financial struct by adding include_health_in_objective as an argument in the Financial constructor, which was going to be my approach as well. So, yes, Financial can go back to being immutable.

src/core/financial.jl
```diff
@@ -159,35 +159,35 @@ mutable struct Financial
         onsite_costs = easiur_costs(latitude, longitude, "onsite")
         escalation_rates = easiur_escalation_rates(latitude, longitude, om_cost_escalation_pct)
         if !isnothing(grid_costs)
-            if ismissing(NOx_grid_cost_per_tonne)
+            if NOx_grid_cost_per_tonne == 0.0
```
hdunham, line 166: Due to changes that @rathod-b has made since this commit, the Financial struct can indeed be made immutable, however the health emissions cost arguments in its constructor cannot have 0 as their default. With the change you've made here, if a user provides a value of 0 for any of the health emissions cost inputs then it will be overwritten. The default needs to be nothing so that we use EASIUR data only if the user didn't provide values. 
I am going to undo the changes made in this commit, except that I will now use nothing instead of missing and I will also leave the changes to the struct field types because handling of failed EASIUR lookup is now done in the Financial constructor (see https://github.com/NREL/REopt.jl/pull/28#discussion_r903296829).

src/core/utils.jl
NLaws: addressed in 7e198f0

src/core/scenario.jl
hdunham: see comment https://github.com/NREL/REopt.jl/pull/28/commits/926318dd0e86ca5b5522d9bc3ee6ad8e85e218f9#r903297193 for why not to make defaults zero

src/core/utils.jl
hdunham: I did this refactor to eliminate duplicate code, but I'm fine to revert it if you find it confusing.

src/constraints/generator_constraints.jl
NLaws: addressed in fb25e34

src/core/site.jl
```diff
@@ -40,11 +40,25 @@ function Site(;
...
+    renewable_electricity_min_pct::Real = 0.0,
+    renewable_electricity_max_pct::Union{Float64, Nothing} = nothing,
+    include_exported_elec_emissions_in_total::Bool = true,
+    include_exported_renewable_electricity_in_total::Bool = true,
     node::Int = 1, 
     )
 ```
 """
-struct Site
+mutable struct Site
```
NLaws, line 61: Why is Site now mutable? We should do what we can to keep it immutable.

test/test_with_xpress.jl
rathod-b: The testset solve times should be fixed per https://github.com/NREL/REopt.jl/commit/b3b3d490c1eea19a3e6611f4287e34f579a57c5e. along with some cleanup of error conditions. Downstream code adds RE electricity constraints if min re elec pct is nothing. But we were defaulting min re elec pct to 0.0 so it was never nothing, and hence always adding min re elec constraints.

src/core/chp.jl
```diff
@@ -87,7 +88,11 @@ can_net_meter::Bool = false
 can_wholesale::Bool = false
 can_export_beyond_nem_limit::Bool = false
 can_curtail::Bool = false
-# emissions_factor_lb_CO2_per_mmbtu::Float64
+fuel_renewable_energy_pct::Union{Nothing,Float64} = nothing
+emissions_factor_lb_CO2_per_mmbtu::Union{Nothing,Float64} = nothing
```
NLaws, line 92: The CHP doc string is incorrect for emissions inputs

src/core/site.jl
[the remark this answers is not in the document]
NLaws: it seems like BAU emissions should not allow the Nothing type, nor be part of Site. They are only needed in REoptInputs I think?  What are the cases that we need to calculate BAU emissions? Every run? Perhaps we can reduce overhead by adding a Emissions input struct that when present in the user inputs triggers all of the emissions capabilities?

src/mpc/structs.jl
hdunham: In the non-MPC case, if the emissions inputs aren't provided then AVERT data is used. In the MPC case, we don't want to do that lookup, and can't anyways because there is no lat/long. Best way of eliminating the MPCElectricUtility that I can think of at the moment would be to add nothing defaults for the lat/long inputs, and in that case don't look up AVERT data. It could theoretically lead to misuse of the constructor without lat/long in the non MPC case, but I think that risk is better than having to maintain separate structs.

src/results/electric_utility.jl
```diff
@@ -37,6 +37,16 @@ ElectricUtility results:
 - `year_one_energy_supplied_kwh` Total energy supplied from the grid in year one.
 - `year_one_to_load_series_kw` Vector of powers drawn from the grid to serve load in year one.
 - `year_one_to_battery_series_kw` Vector of powers drawn from the grid to charge the battery in year one.
+- `year_one_emissions_tCO2`
```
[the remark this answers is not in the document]
hdunham: @adfarth are the units of these tonnes or tons?

src/results/results.jl
rathod-b: Defaulted to 0.0 in https://github.com/NREL/REopt.jl/commit/7c7fba7d1e0da2e4b752e23f0c0bd5b3a58ba9fe

src/core/site.jl
hdunham: added "or equal to" and also changed all throw(@error "") to error("") for consistency

src/core/bau_inputs.jl
hdunham: Right. The outage is only considered in the emissions calculations when a single outage is modeled. This is noted in the doc strings of the renewable energy calc/constraints functions. I also just added a note about it to the ElectricUtility doc string for documentation at a higher level.

src/core/site.jl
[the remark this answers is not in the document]
rathod-b: Here is a snapshot of where major emissions inputs reside in REopt as-is. Adding a new emissions struct where we always run some functionality and conditionally (i.e. take different actions if off-grid is true etc.) run some functionality makes sense as well.
image

src/core/bau_inputs.jl
hdunham: done in 9f9c77a

hdunham: done in commit 9f9c77a

src/core/bau_inputs.jl
hdunham: made issue #87 for this

src/core/site.jl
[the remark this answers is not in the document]
hdunham: BAU emissions need to be nothing if there is no BAU scenario.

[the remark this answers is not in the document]
hdunham: I did consider the idea of creating a separate emissions struct, but opted to keep emissions inputs in their respective structs rather than overhauling the approach that @adfarth and Kathleen took in the API.

src/core/site.jl
rathod-b: Resolving this comment because its due to https://github.com/NREL/REopt.jl/pull/28/files#r890278659

src/core/chp.jl
rathod-b: Resolving. See updates in https://github.com/NREL/REopt.jl/commit/7c7fba7d1e0da2e4b752e23f0c0bd5b3a58ba9fe for resolution.

src/core/financial.jl
rathod-b: Sounds good, resolving.

src/core/scenario.jl
rathod-b: @hdunham @adfarth should EASIUR based grid_costs, onsite_costs and escalation_rates only be set if off_grid_flag is false?

src/core/scenario.jl
rathod-b: Addressed, this code has been moved into ElectricUtility struct. Resolving.

src/core/scenario.jl
rathod-b: This has been addressed in `emissions_series` where default returned is a vector of `zeros(8760*time_steps_per_hour)`

src/core/generator.jl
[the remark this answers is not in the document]
rathod-b: Addressed based on `generator.jl`

src/core/site.jl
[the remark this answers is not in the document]
rathod-b: I am tracking. there is a lot of results processing being done in `results/site.jl` for emissions. However the fields related to emissions in the Site struct appear to be set in the Site struct for BAU and optimal scenarios and aren't used downstream in emissions reduction or pct RE electricity calculations. Instead we are currently creating new keys in `add_site_results` function. So maybe we have some room to take out any fields in the Site structs which arent used downstream in calculations or model setting?

src/results/electric_utility.jl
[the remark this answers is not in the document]
adfarth: @hdunham all `t`s are tonnes !

src/results/site.jl
```diff
@@ -0,0 +1,107 @@
...
+"""
+function add_site_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _n="")
+	r = Dict{String, Any}()
+
+	# renewable elec
+	r["annual_renewable_electricity_kwh"] = round(value(m[:AnnualREEleckWh]), digits=2)
+	r["renewable_electricity_pct"] = round(value(m[:AnnualREEleckWh])/value(m[:AnnualEleckWh]), digits=6)
+
+	# total renewable 
+	add_re_tot_calcs(m,p)
```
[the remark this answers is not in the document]
rathod-b: This is done, resolving.

src/constraints/emissions_constraints.jl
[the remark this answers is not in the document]
hdunham: done in commit 2114cc92

src/results/electric_utility.jl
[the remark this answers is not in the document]
hdunham: changed to tonnes in commit f42b8cc

src/core/electric_utility.jl
hdunham: if emissions factor look up in emissions_series fails, we are now warning and returning zeros

src/core/site.jl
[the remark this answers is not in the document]
hdunham: True, it's totally unnecessary for us to be calculating BAU emissions before optimization for any pollutants other than CO2 because that's the only one with a reduction % constraint. I deleted all of that in 4fe60b2

src/results/results.jl
hdunham: I've now excluded it

src/core/scenario.jl
hdunham: I think we need to go back to returning nothing from emissions_series if it fails, because if that happens and the user requested an emissions reduction or emissions cost in the objective then we need to error. If the user didn't do either of those, then we can emissions factors to zeros. I also need to bring back the code I wrote to do this check that you commented out, but I will put it in the ElectricUtility struct instead of Scenario so that ElectricUtility can be immutable.

src/core/scenario.jl
hdunham: see https://github.com/NREL/REopt.jl/pull/28#discussion_r910483007

src/core/electric_utility.jl
hdunham: Oh I see, that's what you meant by breaking results/electric_utility.jl. In commit afd2191 I've made distance_to_emissions_region_meters = typemax(Int64) instead of nothing.

src/core/electric_utility.jl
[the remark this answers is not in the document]
hdunham: The commit didn't address that situation, it removed my code handling it. I've added it back in, but within the ElectricUtility constructor so as to avoid Nothing in the types.

src/core/scenario.jl
hdunham: done in commit 5a22af9

src/core/scenario.jl
hdunham: done in 7dd4fe7

src/core/utils.jl
hdunham: I can't fully remember but looks like it was because wholesale_rate is also allowed to be a scalar Real, so in this case you get an error trying to convert to Array{Real, 1}. I actually don't really understand what the purpose of doing these conversions here is because isn't that done by specifying constructor argument types?

src/core/scenario.jl
hdunham: done in 34d150c

src/core/scenario.jl
hdunham: in 9b613d4, not looking up grid costs if off grid

src/core/financial.jl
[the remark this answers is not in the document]
hdunham: Done in commit 1eda79b and 6b1dcdf. I also changed the defaults for health emissions cost inputs in the Financial constructor from 0.0 back to nothing. That way if the user provides values of 0.0 they won't be overwritten (commit 225c9f5).

PR #31, opened 2022-03-16
Fix outage time step bugs

hdunham:
When creating the CHP production factor, to check if an outage is being modeled, we're testing if the difference between outage start and end time steps is >= 1. This is incorrect because a difference of zero would just be a 1 tilmestep outage. I have changed the if statement to check if the outage start and end times are non zero (zero is not a valid time step so it represents no outage).
Also, when creating the time steps with and without grid inputs, we're checking if outage end > outage start. I have changed this to >= because again start and end can be equal.


NLaws:
@hdunham can you please change the base to develop? I tried and I get a 500 error. You might have to close this PR and open a new one.

NLaws:
@hdunham nevermind the last comment, github issues have resolved

PR #45, opened 2022-04-20
Switch inpoly2() to ArchGDAL

rathod-b:
Switch to using ArchGDAL's contains() method to find if a point is inside a polygon (i.e. determine the ashrae city for a site)


src/REopt.jl
```diff
@@ -61,6 +61,7 @@ import DelimitedFiles: readdlm
 const MOI = MathOptInterface
 using Shapefile
 using PolygonInbounds
```
NLaws, line 63: @rathod-b can we remove `PolygonInbounds` from the environment?

PR #46, opened 2022-04-23
update ASHRAE city look up

NLaws:
@rathod-b I had to revert merging this change into develop because the city look ups are failing. See the Actions tests: many cities are not found and it is reverting to the geometrically nearest city. Can you please investigate this issue? We can merge this change once the same cities are found as the previous method.


rathod-b:
It seems the ArchGDAL module is having trouble returning the city value when a match is made. I can print the corresponding city value but when I try to return the value it doesnt seem to return anything. I will find an alternative.

PR #54, opened 2022-05-09
Correct dvFuelUsage unit inconsistency

hdunham:
Decision variable dvFuelUsage is supposed to be in units of kWh, but for Generator techs it is in units of gallons. This PR changes the units of dvFuelUsage for Generator techs to kWh to be consistent with all other fuel burning techs and to avoid calculation errors this inconsistency would likely cause in the future.

### DIFF
--- a/src/REopt.jl
+++ b/src/REopt.jl
@@ -64,6 +64,13 @@ using ArchGDAL
 using Roots: fzero  # for IRR
 global hdl = nothing
 
+const M3_TO_GAL = 264.172  # [gal/m^3]
+const GAL_DIESEL_TO_KWH = 40.7  # [kWh/gal_diesel]
+const MMBTU_TO_KWH = 293.07107  # [kWh/mmbtu]
+const TONHOUR_TO_KWH_THERMAL = 3.51685
+const EXISTING_BOILER_EFFICIENCY = 0.8
+const TONNES_TO_LBS = 2204.62  # [lb/tonne]
+
 include("keys.jl")
 include("core/types.jl")
 include("core/utils.jl")
--- a/src/REopt.jl
+++ b/src/REopt.jl
@@ -64,12 +64,13 @@ using ArchGDAL
 using Roots: fzero  # for IRR
 global hdl = nothing
 
-const M3_TO_GAL = 264.172  # [gal/m^3]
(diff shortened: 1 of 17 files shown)


src/REopt.jl
```diff
@@ -64,9 +64,6 @@ using ArchGDAL
 using Roots: fzero  # for IRR
 global hdl = nothing
 
-const M3_TO_GAL = 264.172  # [gal/m^3]
```
NLaws, line 67: `const`s should be defined at the module level since they are global variables (and this way the order of the include statements does not impact which files can access these global variables). Please put these `const`s back here and move any other `const`s that you are aware of to this file.

src/constraints/generator_constraints.jl
```diff
@@ -89,6 +89,6 @@ function add_gen_constraints(m, p)
         m[:dvRatedProduction][t, ts] for t in p.techs.gen, ts in p.time_steps)
     )
     m[:TotalGenFuelCosts] = @expression(m, p.pwf_e *
-        sum(m[:dvFuelUsage][t,ts] * p.s.generator.fuel_cost_per_gallon for t in p.techs.gen, ts in p.time_steps)
+        sum(m[:dvFuelUsage][t,ts] * p.s.generator.fuel_cost_per_gallon / GAL_DIESEL_TO_KWH for t in p.techs.gen, ts in p.time_steps)
```
NLaws, line 92: should this be `* GAL_DIESEL_TO_KWH`? Maybe we should rename the `const` to be more clear how it should be used. Like `GAL_DIESEL_PER_KWH`.

NLaws (changes requested):

src/constraints/generator_constraints.jl
hdunham: Iff you have something in units of A then you can multiply it by A_TO_B to get units of B. If you want to go from B to A, or convert the denominator units from A to B (as is the case here), then you divide by A_TO_B. 
I think all the unit conversion consts used to be A_PER_B actually, but someone changed them to this naming convention. I don't really care either way as long as it's consistent, maybe slightly prefer A_PER_B.

src/REopt.jl
hdunham: Thanks, I was wondering if I should have consolidated them all here instead.

NLaws (approved):

PR #71, opened 2022-06-09
PV fixes

adfarth:
Previous Julia Pkg assumptions did not mirror API assumptions. 
This PR updates PV defaults to tilt=10 for rooftop, tilt = abs(lat) for ground mount, azimuth = 180 for northern lats, azimuth = 0 for southern lats.

Also fixed bug in Generator inputs to allow for time_steps_per_hour > 1


adf_tests/generator.json
```diff
@@ -0,0 +1,55 @@
+{
```
NLaws, line 1: this file should be in test/scenarios?

test/test_with_xpress.jl
```diff
@@ -1027,3 +1027,34 @@ end
 
 end
 
+@testset "PVspecs" begin
+    ## Scenario 1: Palmdale, CA; array-type = 0 (Ground-mount)
+    post_name = "pv.json" 
+    post = JSON.parsefile("./scenarios/$post_name")
+    scen = Scenario(post)
+    
+    print("\nlatitude: ", post["Site"]["latitude"])
```
NLaws, line 1036: please remove these print statements

test/test_with_xpress.jl
```diff
@@ -1027,3 +1027,34 @@ end
 
 end
 
+@testset "PVspecs" begin
```
NLaws, line 1030: since this new test does not require a solver I think that we should put it in the runtests.jl file, and that we should generally house tests that don't require a solver in runtests.jl (or a new file like unittests.jl that is `include`d in runtests.jl - in the future). Can you please move this test to runtests.jl and remove it from test_with_xpress & test_with_cplex ?

test/test_with_xpress.jl
```diff
@@ -1027,3 +1027,34 @@ end
...
+
+    @test scen.pvs[1].tilt ≈ 10
+
+    ## Scenario 3:Cape Town; array-type = 0 (ground)
+    post["Site"]["latitude"] = -33.974732
+    post["Site"]["longitude"] = 19.130050
+    post["PV"]["array_type"] = 0 
+    scen = Scenario(post)
+
+    @test scen.pvs[1].tilt ≈ abs(post["Site"]["latitude"])
```
NLaws, line 1056: It looks like we also need to test a user provided tilt value (and with your changes a user provided tilt would not be used because of the `tilt::Real= array_type == 1 ? 10 : abs(latitude)` in the PV constructor signature).

NLaws (changes requested):
Please see inline comments

test/test_with_xpress.jl
adfarth: addressed in https://github.com/NREL/REopt.jl/pull/71/commits/ed4eb816434cf961579a2b827c29525a5764c61c

test/test_with_xpress.jl
adfarth: addressed in https://github.com/NREL/REopt.jl/pull/71/commits/ed4eb816434cf961579a2b827c29525a5764c61c

src/core/pv.jl
```diff
@@ -126,11 +128,13 @@ struct PV <: AbstractTech
 
     function PV(;
         off_grid_flag::Bool = false,
-        tilt::Real,
-        array_type::Int=1,
-        module_type::Int=0,
+        latitude::Real,
+        longitude::Real,
```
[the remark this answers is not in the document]
adfarth: Good catch! Addressed in https://github.com/NREL/REopt.jl/pull/71/commits/8b35e5d34a0424cee6dec1c027e22c4681c5c1df

test/test_with_xpress.jl
adfarth: added test in https://github.com/NREL/REopt.jl/pull/71/commits/8b35e5d34a0424cee6dec1c027e22c4681c5c1df

NLaws (approved):
Thank you Amanda!

PR #73, opened 2022-06-13
Limited gen reliability

sericson0:


src/core/backup_reliability.jl
```diff
@@ -0,0 +1,82 @@
...
+# OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+# OF THE POSSIBILITY OF SUCH DAMAGE.
+# *********************************************************************************
+"""
+BackupReliability
+
+BackupReliability data struct with inner constructor:
+```julia
+function BackupReliability(;
+    gen::Float64 = 0.9998,
```
NLaws, line 36: what is `gen`, `gen_fts`,  and `gen_ftr`? Please change all shortened names and acronyms to self-explanatory names. Simple shortened names are acceptable, like `num` and `max`.

src/results/results.jl
```diff
@@ -105,6 +105,10 @@ function reopt_results(m::JuMP.AbstractModel, p::REoptInputs; _n="")
         add_flexible_hvac_results(m, p, d)
     end
 
+    if !isnothing(p.s.backup_reliability)
```
NLaws, line 108: This is always true and unnecessary. I think that we should let the user decide if they want to run the backup reliability calculation the same way we let them decide to run the outage simulator: that is they pass the results dictionary from `run_reopt` or `reopt_results` in to `simulate_outages`.

src/core/scenario.jl
```diff
@@ -46,6 +46,7 @@ struct Scenario <: AbstractScenario
     flexible_hvac::Union{FlexibleHVAC, Nothing}
     existing_chiller::Union{ExistingChiller, Nothing}
     absorption_chiller::Union{AbsorptionChiller, Nothing}
+    backup_reliability::BackupReliability
```
NLaws, line 49: Why do we need `BackupReliability` attached to the `Scenario`? I think that we should run the reliability simulation the same way that we run the outage simulator (see https://github.com/NREL/REopt.jl/blob/cece86212568ca6171e01469eab0bac9af6eb561/test/runtests.jl#L115 for example). Rather than create a new Scenario field we could put the default arguments to the backup reliability simulation in the function signature as keyword arguments (like `microgrid_only` in `simulate_outages`). What do you think?

src/outagesim/backup_reliability.jl
```diff
@@ -0,0 +1,925 @@
...
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED.
+# IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+# INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+# LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE
+# OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+# OF THE POSSIBILITY OF SUCH DAMAGE.
+# *********************************************************************************
+Numeric = Union{Int, Float64}
```
NLaws, line 30: Could you replace all `Numeric` instances with the built-in `Real` type?

src/outagesim/backup_reliability.jl
```diff
@@ -0,0 +1,925 @@
...
+
+"""
+    transition_prob(n::Vector{Int}, n_prime::Vector{Int}, p::Numeric)
```
NLaws, line 33: thank you for the great doc strings!

src/outagesim/backup_reliability.jl
```diff
@@ -0,0 +1,925 @@
...
+- `FTS::Numeric`: Failure to Start. The chance that a generator fails to successfully start and take load.
+
+# Examples
+```repl-julia
+julia> starting_probabilities(2, 0.99, 0.05)
+1×3 Matrix{Float64}:
+ 0.00354025  0.11192  0.88454
+```
+"""
+function starting_probabilities(N::Int, OA::Numeric, FTS::Numeric)::Matrix{Float64}
```
NLaws, line 97: please use descriptive names for variables. it is nice to keep them consistent from user input names (is `FTS` here the same as `gen_fts` provided by the user?)

src/outagesim/backup_reliability.jl
```diff
@@ -0,0 +1,925 @@
...
+
+
+#TODO allow for multiple generator types
+# function backup_reliability(d::Dict, p::REoptInputs)
+#     results = return_backup_reliability(d::Dict, p::REoptInputs)
+#     return process_reliability_results(results, length(p.time_steps), p.s.backup_reliability.max_outage_duration)
+# end
+
+
+# function simulate_outage(;init_time_step, diesel_kw, fuel_available, b, m, diesel_min_turndown, batt_kwh, batt_kw,
```
NLaws, line 578: please remove all the commented code (above and below this line)

src/results/backup_reliability.jl
```diff
@@ -0,0 +1,44 @@
...
+	add_backup_reliability_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict)
+
+Adds the Outages results to the dictionary passed back from `run_reopt` using the solved model `m` and the `REoptInputs`.
+Only added to results when multiple outages are modeled via the `ElectricUtility.outage_durations` input.
+
+!!! note
+	When modeling PV the name of the PV system is used for the output keys to allow for modeling multiple PV systems. The default PV name is `PV`.
+
+"""
+function add_backup_reliability_results(m, p, d::Dict)
```
NLaws, line 40: this function should be in outagesim/backup_reliability.jl and get exported under a different name, like `get_backup_reliability_results`. Also, it looks like `m` is unused, so the new function will take in `p` (please add type), `d`, and the other keyword args that are in the `BackupReliability` function.

src/outagesim/backup_reliability.jl
```diff
@@ -0,0 +1,925 @@
...
+#     return (critical_load .- solar_pv) .- chp_kw
+# end
+
+
+#_____________________________________________________________________________________________________________________
+#_____________________________________________________________________________________________________________________
+#_____________________________________________________________________________________________________________________
+
+function return_backup_reliability(d::Dict, p::REoptInputs)
+    #TODO add microgrid_only boolian to scenario
```
NLaws, line 444: Should this TODO and others in your PR be done before we merge this branch?

src/outagesim/backup_reliability.jl
```diff
@@ -0,0 +1,925 @@
...
+
+    else
+        net_critical_loads_kw = p.s.electric_load.critical_loads_kw .- pv_kw_ac_hourly
+        return [survival_with_battery(net_critical_loads_kw, starting_batt_soc_kwh, gen_oa, gen_fts, gen_ftr, num_gens, gen_capacity, batt_kwh, batt_kw,
+                                        num_battery_bins, max_outage_duration, batt_charge_efficiency, batt_discharge_efficiency, marginal_survival = true),
+                survival_with_battery(net_critical_loads_kw, starting_batt_soc_kwh, gen_oa, gen_fts, gen_ftr, num_gens, gen_capacity, batt_kwh, batt_kw,
+                                        num_battery_bins, max_outage_duration, batt_charge_efficiency, batt_discharge_efficiency, marginal_survival = false)] 
+
+    end
+
```
NLaws, line 535: please remove extra new lines (two new lines between functions/doc strings is great and one new line is nice for legibility in the code - no need to change the spacing you have in general, but there are 4 new lines here and a couple places that you have 2 new lines between code where 1 is sufficient.)

src/REopt.jl
```diff
@@ -61,6 +61,9 @@ import DelimitedFiles: readdlm
 const MOI = MathOptInterface
 using Shapefile
 using ArchGDAL
+#See if these are needed
```
NLaws, line 64: Please check if these two new dependencies are necessary

NLaws (changes requested):
I don't see why the BackupReliability struct should be created for every Scenario. Can we follow the design pattern for `simulate_outages` as I described in the comments? Please go through my comments and respond to each with an answer/opinion/thoughts. Thank you!

PR #75, opened 2022-06-14
Offgrid wind

adfarth:
This PR introduces the following changes: 
- Allow Wind tech to be included when `off_grid_flag` is true
- Add `operating_reserve_required_pct` to Wind struct and incorporate wind into operating reserve constraints
- Bug fix to constrain `dvCurtail` in `time_steps_without_grid` 
- Bug fix to report accurate wind `["year_one_to_load_series_kw"]` in results/wind.jl (was previously not accounting for curtailed wind)


src/core/techs.jl
```diff
@@ -140,7 +147,7 @@ function Techs(s::Scenario)
     end
 
     if "Wind" in all_techs
-        append!(techs_no_turndown, ["Wind"])
+        append!(techs_no_turndown, ["Wind"]) # TODO: can this get moved to the if s.wind.max_kw > 0 statement?
```
NLaws, line 150: regarding the TODO: yes it can if you want to move it

src/results/wind.jl
```diff
@@ -56,50 +56,43 @@ function add_wind_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _n="")
...
 	)
-	r["year_one_to_load_series_kw"] = round.(value.(prod_to_load), digits=3)
+	r["year_one_to_load_series_kw"] = round.(value.(WindToLoad), digits=3)
 
-	Year1GenProd = @expression(m,
-		p.hours_per_time_step * sum(m[:dvRatedProduction][t,ts] * p.production_factor[t, ts]
-			for ts in p.time_steps)
-	)
-	r["year_one_energy_produced_kwh"] = round(value(Year1GenProd), digits=0)
+	Year1WindProd = (sum(m[Symbol("dvRatedProduction"*_n)][t,ts] * p.production_factor[t, ts] for ts in p.time_steps) * p.hours_per_time_step)
```
NLaws, line 88: The expression `m[Symbol("dvRatedProduction"*_n)][t,ts] * p.production_factor[t, ts] for ts in p.time_steps` is in at least three places in these results. I think that we can save time in results processing (for all results) by avoiding redundant expressions and especially `JuMP.value` calls. @Bill-Becker has been looking into the time consumption in v1/2 of the API in the results processing. Bill: do you have any insights that could speed things up in the REopt.jl results? Should we reduce redundant `JuMP.value` calls and/or additional `@expression`s if we can?

src/core/techs.jl
adfarth: moved in https://github.com/NREL/REopt.jl/pull/75/commits/67e753736b25ac4f9549962b1dc2b53178821e63

test/Manifest.toml
```diff
@@ -1,6 +1,6 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.7.2"
+julia_version = "1.7.1"
```
adfarth, line 3: @NLaws  I think I accidentally changed this to 1.7.1 -- is this an issue and if so can I manually change it back?

test/test_with_xpress.jl
```diff
@@ -1025,6 +1027,29 @@ end
     # Test generator outputs
     @test typeof(r) == Model # this is true when the model is infeasible
 
+    ### Scenario 3: Indonesia. Wind (custom prod) and Generator only
+    m = Model(optimizer_with_attributes(Xpress.Optimizer, "MIPRELSTOP" => 0.01, "OUTPUTLOG" => 0))
+    post_name = "wind_intl_offgrid.json" 
+    post = JSON.parsefile("./scenarios/$post_name")
+    post["ElectricLoad"]["loads_kw"] = [10.0 for i in range(1,8760)]
+    scen = Scenario(post)
+    df = CSV.read("./data/example_wind_prod_factor_kw.csv", skipto=2, DataFrame)
```
NLaws, line 1035: please remove DataFrames and CSV from the test environment and use [readdlm](https://docs.julialang.org/en/v1/stdlib/DelimitedFiles/#DelimitedFiles.readdlm-Tuple{Any,%20AbstractChar}) from the standard library DelimitedFiles instead

test/Manifest.toml
NLaws: the Manifest.toml was changed to your julia_version when you `Pkg.add`ed DataFrames and CSV to the test environment. I left a review comment about removing those dependencies so you should be able to revert the test/Project.toml and test/Manifest.toml back to their original states. But in general I don't think that it will be an issue but we should try to all use the latest stable version of Julia.

test/test_with_xpress.jl
adfarth: removed in https://github.com/NREL/REopt.jl/pull/75/commits/840ea0ce8d3fa56c2ff5fa01b13d06ba8a85854d

src/core/prodfactor.jl
```diff
@@ -210,6 +209,10 @@ function prodfactor(wind::Wind, latitude::Real, longitude::Real, time_steps_per_
         @ccall hdl.ssc_data_set_array(wind_resource::Ptr{Cvoid}, "fields"::Cstring, 
             fields::Ptr{Cdouble}, length(fields)::Cint)::Cvoid
 
+        if size(resources)[1] == 1
```
NLaws, line 212: @rathod-b will you please add a comment in the code here explaining why we need this `if` statement?

adfarth: @rathod-b @NLaws should we consider this resolved?

src/results/wind.jl
adfarth: @Bill-Becker any thoughts here?

adfarth: addressed in https://github.com/NREL/REopt.jl/pull/75/commits/e64422d11776657f2c0b3b99f8956a9126050439

src/core/prodfactor.jl
```diff
@@ -99,8 +99,8 @@ function prodfactor(wind::Wind, latitude::Real, longitude::Real, time_steps_per_
 
     if all(length(a) > 0 for a in [wind.temperature_celsius, wind.pressure_atmospheres, wind.wind_direction_degrees,
                                    wind.wind_meters_per_sec])
-        push!(resources, [wind.temperature_celsius, wind.pressure_atmospheres, wind.wind_direction_degrees,
-                          wind.wind_meters_per_sec])
+        push!(resources, hcat([wind.temperature_celsius, wind.pressure_atmospheres, wind.wind_meters_per_sec, wind.wind_direction_degrees]...))
+        resources = hcat(resources...)
```
NLaws, line 103: it appears that in both of these lines some values are concatenated with `hcat` then immediately unpacked with `...`. Looks unnecessary.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # REopt Changelog
 
+## Develop
+- Allow Wind tech to be included when `off_grid_flag` is true
+- Add `operating_reserve_required_pct` to Wind struct and incorporate wind into operating reserve constraints
+- Bug fix to constrain dvCurtail in `time_steps_without_grid`
+- Bug fix to report accurate wind ["year_one_to_load_series_kw"] in results/wind.jl (was previously not accounting for curtailed wind)
+
 ## Develop
```
NLaws, line 9: note the double ## Develop headers (sure it will get resolved in merge to master branch)

NLaws (approved):
left a couple notes but looks good without any changes necessary

src/core/prodfactor.jl
rathod-b: Removed the surplus hcat when pushing wind resources to resources[] in https://github.com/NREL/REopt.jl/commit/c13a75042f1421ef0834058133cef6d0d98f229f

adfarth: Thanks Bhavesh!

PR #96, opened 2022-07-12
Limited Generator Reliability

sericson0:


Bill-Becker:
@sericson0 Can you add a test for the main function of `backup_reliability.jl`, like [this one does for outage_simulator](https://github.com/NREL/REopt.jl/blob/0f213370b6043cbe73d169f0197651c59fe5c2bb/test/runtests.jl#L115)?

Bill-Becker:
@sericson0 add a line or two to the CHANGELOG.md for these additions, under the top "#Develop" section (TBD a version).

src/outagesim/backup_reliability.jl
```diff
@@ -169,38 +172,38 @@ function get_maximum_generation(batt_kw::Real, gen_capacity::Real, bin_size::Rea
...
     end
     
     return generator_prod .+ max_battery_discharge
 end
 
 """
-    battery_bin_shift(excess_generation::Vector, bin_size::Real, batt_kw::Real, batt_charge_efficiency::Real, batt_discharge_efficiency::Real)
+    battery_bin_shift(excess_generation_kw::Vector, bin_size::Real, batt_kw::Real, batt_charge_efficiency::Real, batt_discharge_efficiency::Real)
 
 Return a vector of number of bins battery is shifted by
```
Bill-Becker, line 184: shifted by...?

src/outagesim/backup_reliability.jl
```diff
@@ -64,25 +98,59 @@ julia> markov_matrix(2, 0.1)
...
+    repeat(0:num_gen, inner = num_gen+1), fail_prob), num_gen+1, num_gen+1)
+    replace!(M, NaN => 0)
+    return M
+end
+
+"""
+    markov_matrix(num_gen_vec::Vector{Int}, fail_prob_vec::Vector{<:Real})::Matrix{Float64} 
+
+Markov Matrix for multiple generator types. 
+Return an prod(``num_gen_vec``.+1) by prod(``num_gen``.+1) matrix of transition probabilities of going from n (row) to n' (column) given probability ``fail_prob``
```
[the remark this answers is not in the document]
hdunham: Function description is very vague and confusing. 
Give a description for the arguments. 
Don't expect someone to immediately see why those are the matrix dimensions. 
Need to say more than rows denote starting generators. People would likely guess that row 0 is 0 gens. But what about row 1? Is it 1 gen of type 0 or type 1? Say how they increment. I think maybe that's what you're getting at with "Function iterates over first to last generator type in list." but it's not clear. The order that the function goes in doesn't matter, it's the ordering of the return value.

hdunham (commented):
I am pausing my review here and submitting to you what I have so far. Most all my notes are about the clarity of documentation. I think if you went and made that more clear throughout the file, it will address most of the additional comments I would probably end up making and will make it much less time consuming to review.

src/outagesim/backup_reliability.jl
```diff
@@ -91,14 +172,51 @@ julia> starting_probabilities(2, 0.99, 0.05)
  0.00354025  0.11192  0.88454
 ```
 """
-function starting_probabilities(N::Int, OA::Real, failure_to_start::Real)::Matrix{Float64} 
-    M = markov_matrix(N, (1-OA) + failure_to_start*OA) 
-    G = hcat(zeros(1, N), 1)
+function starting_probabilities(num_gen::Int, gen_operational_availability::Real, gen_failure_to_start::Real)::Matrix{Float64} 
+    M = markov_matrix(num_gen, (1-gen_operational_availability) + gen_failure_to_start*gen_operational_availability)
```
hdunham, line 176: Not ideal to calculate this whole matrix just to take the last row

src/outagesim/backup_reliability.jl
```diff
@@ -91,14 +172,51 @@ julia> starting_probabilities(2, 0.99, 0.05)
  0.00354025  0.11192  0.88454
 ```
 """
-function starting_probabilities(N::Int, OA::Real, failure_to_start::Real)::Matrix{Float64} 
-    M = markov_matrix(N, (1-OA) + failure_to_start*OA) 
-    G = hcat(zeros(1, N), 1)
+function starting_probabilities(num_gen::Int, gen_operational_availability::Real, gen_failure_to_start::Real)::Matrix{Float64} 
+    M = markov_matrix(num_gen, (1-gen_operational_availability) + gen_failure_to_start*gen_operational_availability) 
+    G = hcat(zeros(1, num_gen), 1)
+    return G * M
```
hdunham, line 178: Why not just index into matrix i.e. M[end,:] ? Also applies to multiple gen type version of function

src/outagesim/backup_reliability.jl
hdunham: This may not be important to address if solve times are already good.

src/outagesim/backup_reliability.jl
```diff
@@ -169,58 +323,130 @@ function get_maximum_generation(batt_kw::Real, gen_capacity::Real, bin_size::Rea
...
+    excess_generation_kw[excess_generation_kw .> 0] = excess_generation_kw[excess_generation_kw .> 0] .* batt_charge_efficiency
+    excess_generation_kw[excess_generation_kw .< 0] = excess_generation_kw[excess_generation_kw .< 0] ./ batt_discharge_efficiency
     #Battery cannot charge or discharge more than its capacity
-    excess_generation[excess_generation .> batt_kw] .= batt_kw
-    excess_generation[excess_generation .< -batt_kw] .= -batt_kw
-    shift = round.(excess_generation ./ bin_size)
+    excess_generation_kw[excess_generation_kw .> batt_kw] .= batt_kw
+    excess_generation_kw[excess_generation_kw .< -batt_kw] .= -batt_kw
+    shift = round.(excess_generation_kw ./ bin_size)
     # shift[is.nan(shift)] = 0
```
hdunham, line 414: can this be deleted?

src/outagesim/backup_reliability.jl
```diff
@@ -240,60 +466,69 @@ If ``marginal_survival`` = true then result is chance of surviving in given outa
...
+    generator_markov_matrix = markov_matrix(num_gen=num_generators, fail_prob=failure_to_run)
   
     #Get starting generator vector
-    starting_gens = starting_probabilities(num_generators, OA, failure_to_start) #initialize gen battery prob matrix
+    starting_gens = starting_probabilities(num_gen=num_generators, gen_operational_availability=gen_operational_availability, gen_failure_to_start=failure_to_start) #initialize gen battery prob matrix
 
-    #start loop
     for t  = 1:t_max
         gen_probs = starting_gens
         #
```
hdunham, line 526: were you intending to add something here?

hdunham: and line 532

src/outagesim/backup_reliability.jl
```diff
@@ -505,76 +765,115 @@ function backup_reliability_inputs(d::Dict, p::REoptInputs; r::Dict = Dict())::D
     end
```
hdunham, line 765: batt_charge_efficiency and batt_discharge_efficiency created but not used? (A few lines above, I can't put a comment there because those changes aren't in this PR.)

hdunham: Need to be added to the dict r2 returned, otherwise arg default is always used later on

src/outagesim/backup_reliability.jl
```diff
@@ -240,60 +466,69 @@ If ``marginal_survival`` = true then result is chance of surviving in given outa
 if ``marginal_survival`` = false then result is chance of surviving up to and including given outage hour.
```
hdunham: Instead of calling this function (and the with battery version) twice, with marginal_survival true and false, just make this function calculate the marginal version and calculate the other from that matrix. I think there's a function that does exactly that, returns a matrix where each element is the product of input matrix elements up to that index, but don't remember what it's called. Doing that should actually make an impact on speed because this is the part that takes longest I'm assuming.

src/outagesim/backup_reliability.jl
```diff
@@ -240,60 +466,69 @@ If ``marginal_survival`` = true then result is chance of surviving in given outa
...
         #
         for d in 1:max_duration
             survival = ones(1, length(generator_production))
             
             h = mod(t + d - 2, t_max) + 1 #determines index accounting for looping around year
-            net_gen = generator_production .- critical_load[h]
+            net_gen = generator_production .- critical_load_kw[h]
             survival[net_gen .< 0] .= 0
 
             gen_probs *= generator_markov_matrix #Update to account for generator failures
```
hdunham, line 540: Shouldn't this update not happen until the 2nd timestep because starting_probabilities() already calculated what it should be in the first tilmestep?

src/outagesim/backup_reliability.jl
```diff
@@ -417,14 +676,14 @@ function survival_with_battery(net_critical_load::Vector, starting_batt_soc_kwh:
             #update expected lost load for given outage start time and outage duration
```
hdunham, line 682: "remove probabilities which did not meet load"?

src/outagesim/backup_reliability.jl
```diff
@@ -616,8 +809,8 @@ Return dictionary of backup reliability results.
     -gen_operational_availability::Real = 0.9998 (Fraction of year generators not down for maintenance)
```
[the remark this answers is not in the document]
hdunham: Following up, the arguments of this function need to be only what the ERP tool would be able to pass to it in the json body of a POST. ERP runs are not linked to a REopt run. REopt results can initially populate the ERP tool fields, but those values can be changed and they are only a subset of the REopt output fields. So neither a full REopt results dict nor a REoptInputs struct should be args to this function. The ERP tool will just send a dict of the necessary inputs such as tech sizes, battery info, survival probs, etc. without differentiation between ones that may have come from the results of a REopt run. So this function should accept that, either as a dict or as keyword args.

src/outagesim/backup_reliability.jl
```diff
@@ -434,20 +697,21 @@ Return a dictionary of inputs required for backup reliability calculations.
     -gen_operational_availability::Real = 0.9998        Fraction of year generators not down for maintenance
     -gen_failure_to_start::Real = 0.0066                Chance of generator starting given outage
     -gen_failure_to_run::Real = 0.00157                 Chance of generator failing in each hour of outage
-    -num_gens::Int = 1                                  Number of generators. Will be determined by code if set to 0 and gen capacity > 0.1
-    -gen_capacity::Real = 0.0                           Backup generator capacity. Will be determined by REopt optimization if set less than 0.1
+    -num_gen::Int = 1                                  Number of generators. Will be determined by code if set to 0 and gen capacity > 0.1
+    -gen_capacity_kw::Real = 0.0                           Backup generator capacity. Will be determined by REopt optimization if set less than 0.1
```
hdunham, line 701: See comment about what's passed to backup_reliability(): https://github.com/NREL/REopt.jl/pull/96/files#r962108891
Reopt results are used to fill in fields initially, but full reopt results not passed to backup_reliability(). There should just be one gen size passed, not a reopt results one and user provided one.

src/outagesim/backup_reliability.jl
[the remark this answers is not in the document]
hdunham: @Bill-Becker could you confirm that the design I'm describing seems good to you?

src/outagesim/backup_reliability.jl
```diff
@@ -230,91 +462,99 @@ function shift_gen_battery_prob_matrix!(gen_battery_prob_matrix::Matrix, shift_v
...
-- `failure_to_run::Real`: hourly Failure to Run probability. failure_to_run is 1/MTTF (mean time to failure). 
-- `num_generators::Int`: number of generators in microgrid.
-- `gen_capacity::Real`: size of generator.
+- `critical_load_kw::Vector`: 8760 vector of system critical loads. 
+- `gen_operational_availability::Union{Real, Vector{<:Real}}`: Operational Availability of backup generators.
+- `failure_to_start::Union{Real, Vector{<:Real}}`: probability of generator Failure to Start and support load. 
+- `failure_to_run::Union{Real, Vector{<:Real}}`: hourly Failure to Run probability. failure_to_run is 1/MTTF (mean time to failure). 
+- `num_generators::Union{Int, Vector{Int}}`: number of generators in microgrid.
+- `gen_capacity_kw::Union{Real, Vector{<:Real}}`: size of generator.
 - `max_duration::Int`: maximum outage duration in hours.
```
hdunham, line 481: Is only 8760 supported so timesteps=hours? If so, that should be documented more clearly.

src/outagesim/backup_reliability.jl
```diff
@@ -434,20 +697,21 @@ Return a dictionary of inputs required for backup reliability calculations.
     -gen_operational_availability::Real = 0.9998        Fraction of year generators not down for maintenance
     -gen_failure_to_start::Real = 0.0066                Chance of generator starting given outage
     -gen_failure_to_run::Real = 0.00157                 Chance of generator failing in each hour of outage
-    -num_gens::Int = 1                                  Number of generators. Will be determined by code if set to 0 and gen capacity > 0.1
-    -gen_capacity::Real = 0.0                           Backup generator capacity. Will be determined by REopt optimization if set less than 0.1
+    -num_gen::Int = 1                                  Number of generators. Will be determined by code if set to 0 and gen capacity > 0.1
+    -gen_capacity_kw::Real = 0.0                           Backup generator capacity. Will be determined by REopt optimization if set less than 0.1
     -num_battery_bins::Int = 100                        Internal value for modeling battery
     -max_outage_duration::Int = 96                      Maximum outage hour modeled
     -microgrid_only::Bool = false                       Determines how generator, PV, and battery act during islanded mode
```
hdunham, line 704: In what was does it determine this?

src/outagesim/backup_reliability.jl
```diff
@@ -434,20 +697,21 @@ Return a dictionary of inputs required for backup reliability calculations.
...
 ```
 """
-function backup_reliability_inputs(d::Dict, p::REoptInputs; r::Dict = Dict())::Dict
+function backup_reliability_inputs(;d::Dict, p::REoptInputs, r::Dict = Dict())::Dict
```
hdunham, line 707: This function should determine just the subset of ERP inputs that come from reopt inputs/results. It will be used to populate the ERP fields initially after a reopt run, at which point the user hasn't provided any ERP specific inputs yet.

src/outagesim/backup_reliability.jl
```diff
@@ -505,76 +769,108 @@ function backup_reliability_inputs(d::Dict, p::REoptInputs; r::Dict = Dict())::D
...
+process_reliability_results(results::Array)::Dict
+
+Return dictionary of processed backup reliability results.
 
-function process_reliability_results(results)
+# Arguments
+- `results::Array`: results from function return_backup_reliability. 
+"""
+function process_reliability_results(results::Array)::Dict
     if results == []
```
hdunham, line 874: isempty()?

src/outagesim/backup_reliability.jl
hdunham, line 701: Why is there this 0.1 cut off? Why not make the default nothing, then use gen_capacity_kw if not nothing and the REopt results value otherwise?

src/outagesim/backup_reliability.jl
```diff
@@ -434,20 +697,21 @@ Return a dictionary of inputs required for backup reliability calculations.
...
     -num_battery_bins::Int = 100                        Internal value for modeling battery
     -max_outage_duration::Int = 96                      Maximum outage hour modeled
     -microgrid_only::Bool = false                       Determines how generator, PV, and battery act during islanded mode
 ```
 """
-function backup_reliability_inputs(d::Dict, p::REoptInputs; r::Dict = Dict())::Dict
+function backup_reliability_inputs(;d::Dict, p::REoptInputs, r::Dict = Dict())::Dict
     zero_array = zeros(length(p.time_steps))
     r2 = dictkeys_tosymbols(r)
     critical_loads_kw = p.s.electric_load.critical_loads_kw
```
hdunham, line 710: need to account for critical_loads_kw_is_net

src/outagesim/backup_reliability.jl
```diff
@@ -434,20 +697,21 @@ Return a dictionary of inputs required for backup reliability calculations.
...
 
+    #TODO Change CHP to meet capacity not load
     if "CHP" in keys(d)
-        chp_generation =  get(d["CHP"], "year_one_electric_production_series_kw", zero_array)
+        chp_generation =  get(d["CHP"], "size_kw", 0)
         critical_loads_kw .-= chp_generation
```
hdunham, line 715: @Bill-Becker was the ERP supposed to have an option to change CHP size?

src/outagesim/backup_reliability.jl
hdunham, line 701: In addition to gen, ERP user also needs to be able to specify PV and CHP size, not just taken from REopt results. Also, when you make the changes to this function that I requested in #r962108891 PV and CHP prod series inputs will need to be normalized to size!

src/outagesim/backup_reliability.jl
```diff
@@ -505,76 +769,108 @@ function backup_reliability_inputs(d::Dict, p::REoptInputs; r::Dict = Dict())::D
...
+Return dictionary of processed backup reliability results.
 
-function process_reliability_results(results)
+# Arguments
+- `results::Array`: results from function return_backup_reliability. 
+"""
+function process_reliability_results(results::Array)::Dict
     if results == []
         marginal_duration_means = []
         marginal_duration_mins = []
```
hdunham, line 876: marginal_final_resilience doesn't seem like a useful output, or am I missing something?

src/outagesim/backup_reliability.jl
sericson0: Applied Change

src/outagesim/backup_reliability.jl
sericson0: Removed the dangling #'s

src/outagesim/backup_reliability.jl
```diff
@@ -505,76 +769,108 @@ function backup_reliability_inputs(d::Dict, p::REoptInputs; r::Dict = Dict())::D
...
-    gen_capacity = get(r, "gen_capacity", 0)
-    num_gens = get(r, "num_gens", 1)
-
-    if gen_capacity < 0.1
-        if num_gens <= 1
-            gen_capacity = diesel_kw
-            num_gens = 1
-        else
-            gen_capacity = diesel_kw / num_gens
+    #If num_gen is zero then either set to 1 or base on ceiling(diesel_kw / gen_capacity_kw)
```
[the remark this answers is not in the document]
sericson0: Original discussion was that PV could only be used in island mode if there is also a battery to smooth output. I.e., PV gets disconnected in island mode if there is not a battery

src/outagesim/backup_reliability.jl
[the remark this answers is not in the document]
Bill-Becker: Sounds good to me!

src/outagesim/backup_reliability.jl
```diff
@@ -28,30 +28,64 @@
...
 end
 
 """
-    markov_matrix(N::Int, p::Real)
+transition_prob(start_gen_vec::Vector{Vector{Int}}, end_gen_vec::Vector{Vector{Int}}, fail_prob_vec::Vector{<:Real})::Vector{Float64}
 
-Return an ``N``+1 by ``N``+1 matrix of transition probabilities of going from n (row) to n' (column) given probability ``p``
+Transition probability for multiple generator types. 
+Return the probability of going from i_t to j_t generators for each of generator type t given a failure rate of 
+``fail_prob`` for each i,j,t in vector of vectors ``start_gen_vec`` and ``end_gen_vec``.
```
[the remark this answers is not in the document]
sericson0: Updated descriptions of function per comments

src/outagesim/backup_reliability.jl
sericson0: Updated function per recommendation

src/outagesim/backup_reliability.jl
sericson0: Deleted

src/outagesim/backup_reliability.jl
hdunham: @Bill-Becker could you provide your opinion on what would be correct here?

src/outagesim/backup_reliability.jl
```diff
@@ -692,22 +692,25 @@ Return a dictionary of inputs required for backup reliability calculations.
...
     -generator_failure_to_run::Real = 0.00157                 Chance of generator failing in each hour of outage
-    -num_genenerators::Int = 1                                  Number of generators. Will be determined by code if set to 0 and gen capacity > 0.1
+    -num_generators::Int = 1                                  Number of generators. Will be determined by code if set to 0 and gen capacity > 0.1
     -generator_size_kw::Real = 0.0                           Backup generator capacity. Will be determined by REopt optimization if set less than 0.1
-    -num_battery_bins::Int = 100                        Internal value for modeling battery
+    -num_battery_bins::Int = 101                        Internal value for modeling battery
     -max_outage_duration::Int = 96                      Maximum outage hour modeled
     -microgrid_only::Bool = false                       Boolian to check if only microgrid upgraded technologg runs during grid outage 
 ```
 """
```
hdunham, line 701: I think all the ERP inputs should be options to include in the Dict arg r

src/outagesim/backup_reliability.jl
hdunham: Also the loop writes over the same variable for each b in p.s.storage.types.elec

src/outagesim/backup_reliability.jl
```diff
@@ -803,37 +806,44 @@ Return a dictionary of inputs required for backup reliability calculations.
...
-    #diesel_size_kw
-    #diesel_size_mg_kw
+    -battery_year_one_soc_series_pct           Battery state of charge in each hour (if not input then defaults to battery size)
     -generator_failure_to_start::Real = 0.0066                Chance of generator starting given outage
     -generator_failure_to_run::Real = 0.00157                 Chance of generator failing in each hour of outage
-    -num_genenerators::Int = 1                                  Number of generators. Will be determined by code if set to 0 and gen capacity > 0.1
+    -num_generators::Int = 1                                  Number of generators. Will be determined by code if set to 0 and gen capacity > 0.1
     -generator_size_kw::Real = 0.0                           Backup generator capacity. Will be determined by REopt optimization if set less than 0.1
-    -num_battery_bins::Int = 100                        Internal value for modeling battery
+    -num_battery_bins::Int = 101                        Internal value for modeling battery
```
hdunham: why did you change this default?

src/outagesim/backup_reliability.jl
```diff
@@ -803,37 +806,44 @@ Return a dictionary of inputs required for backup reliability calculations.
...
 ```
 """
 function backup_reliability_inputs(;r::Dict)::Dict
-    r2 = dictkeys_tosymbols(r)
-    critical_loads_kw = r["critical_loads_kw"]
-    zero_array = zeros(length(critical_loads_kw))
+    
+    
+    r2 = Dict()
+    r2[:generator_operational_availability] = get(r, "generator_operational_availability", 0.9998)
```
hdunham, line 828: why are you doing these individual gets instead of dictkeys_tosymbols like it was before? There are already defaults for args in return_backup_reliability, so its not great to duplicate those defaults here and thus never use the ones in the return_backup_reliability function signature

src/outagesim/backup_reliability.jl
```diff
@@ -396,59 +655,62 @@ function survival_with_battery(net_critical_load::Vector, starting_batt_soc_kwh:
...
-            shift_gen_battery_prob_matrix!(gen_battery_prob_matrix, battery_bin_shift(excess_generation, bin_size, batt_kw, batt_charge_efficiency, batt_discharge_efficiency))
+            shift_gen_battery_prob_matrix!(gen_battery_prob_matrix, battery_bin_shift(excess_generation_kw, bin_size, battery_size_kw, battery_charge_efficiency, battery_discharge_efficiency))
         end
     end
     return survival_probability_matrix
 end
 
 """
-    backup_reliability_inputs(d::Dict, p::REoptInputs; r::Dict)::Dict
+    backup_reliability_reopt_inputs(;d::Dict, p::REoptInputs, r::Dict)::Dict
```
hdunham, line 685: Why not just call backup_reliability_inputs in backup_reliability_reopt_inputs instead of duplicating code? 
Or, even better, stop backup_reliability_reopt_inputs before it gets to the part that's a responsibility of backup_reliability_inputs. Then for the two backup_reliability functions, the one with 3 arg can just call backup_reliability_reopt_inputs and pass what is returned to the 1 arg backup_reliability function.

src/outagesim/backup_reliability.jl
```diff
@@ -591,40 +993,78 @@ function process_reliability_results(results)
...
+#diesel_size_mg_kw
+-generator_failure_to_start::Real = 0.0066                Chance of generator starting given outage
+-generator_failure_to_run::Real = 0.00157                 Chance of generator failing in each hour of outage
+-num_generators::Int = 1                                  Number of generators. Will be determined by code if set to 0 and gen capacity > 0.1
+-generator_size_kw::Real = 0.0                           Backup generator capacity. Will be determined by REopt optimization if set less than 0.1
+-num_battery_bins::Int = 100                        Internal value for discretely modeling battery state of charge
+-max_outage_duration::Int = 96                      Maximum outage hour modeled
+
+"""
+function backup_reliability(r::Dict)
```
hdunham, line 1066: this function should catch exceptions and set an "error" field in the returned dictionary

src/outagesim/backup_reliability.jl
Bill-Becker: I honestly have not navigated my way around this math to understand it; would Sean be more appropriate to answer that, or is there a debate here between you two :-)?

src/outagesim/backup_reliability.jl
```diff
@@ -396,59 +655,62 @@ function survival_with_battery(net_critical_load::Vector, starting_batt_soc_kwh:
...
 
+    #TODO Change CHP to meet capacity not load
     if "CHP" in keys(d)
-        chp_generation =  get(d["CHP"], "year_one_electric_production_series_kw", zero_array)
-        critical_loads_kw .-= chp_generation
+        chp_generation =  get(d["CHP"], "size_kw", 0)
+        net_critical_loads_kw .-= chp_generation
     end
 
     microgrid_only = get(r, "microgrid_only", false)
```
hdunham, line 716: Need to handle multiple PVs. d["PV"] could be a list of dictionaries instead of a dictionary

src/outagesim/backup_reliability.jl
```diff
@@ -505,77 +767,217 @@ function backup_reliability_inputs(d::Dict, p::REoptInputs; r::Dict = Dict())::D
...
     #If gen capacity is 0 then base on diesel_kw
-    #If num_gens is zero then either set to 1 or base on ceiling(diesel_kw / gen_capacity)
-    gen_capacity = get(r, "gen_capacity", 0)
-    num_gens = get(r, "num_gens", 1)
-
-    if gen_capacity < 0.1
-        if num_gens <= 1
-            gen_capacity = diesel_kw
-            num_gens = 1
+    #If num_generators is zero then either set to 1 or base on ceiling(diesel_kw / generator_size_kw)
```
hdunham, line 772: If the user provides a reopt run and a generator_size_kw to override that but not num_generators, should num_generators be set to 1 or reopt results gen size divided by generator_size_kw?
@Bill-Becker

src/outagesim/backup_reliability.jl
Bill-Becker: @hdunham and we would just aggregate/sum the PV production for the ERP? Alternatively to handling this within backup_reliability.jl, since we are not modeling multiple PV's explicitly in ERP, we could make this a requirement for the ERP **endpoint** instead of these functions - i.e. process the REopt inputs to handle/aggregate multiple PV's. What if the REopt outputs have multiple PVs but then the ERP user changes the PV size? How do we normalize the PV production to scale to different sizes? If this gets too tricky, I'd be fine only allowing one PV to be handled, such as the first one from REopt if there are multiple. We don't allow multiple PV's from the UI.

src/outagesim/backup_reliability.jl
hdunham: I think the former, but just wanted to check

src/outagesim/backup_reliability.jl
```diff
@@ -505,77 +769,217 @@ function backup_reliability_inputs(d::Dict, p::REoptInputs; r::Dict = Dict())::D
...
+function backup_reliability_inputs(;r::Dict)::Dict
+    
+    
+    r2 = Dict()
+    r2[:generator_operational_availability] = get(r, "generator_operational_availability", 0.9998)
+    r2[:generator_failure_to_start] = get(r, "generator_failure_to_start", 0.0066)
+    r2[:generator_failure_to_run] = get(r, "generator_failure_to_run", 0.00157)
+    r2[:num_generators] = get(r, "num_generators", 1)
+    r2[:generator_size_kw] = get(r, "generator_size_kw", 0.0)
+    r2[:max_outage_duration] = get(r, "num_battery_bins", 96)
```
hdunham, line 835: incorrect assignment of battery bins to outage duration

src/outagesim/backup_reliability.jl
```diff
@@ -505,77 +769,217 @@ function backup_reliability_inputs(d::Dict, p::REoptInputs; r::Dict = Dict())::D
...
+        net_critical_loads_kw .-= r["chp_capacity"]
+    end
+
+    microgrid_only = get(r, "microgrid_only", false)
+
+    pv_kw_ac_hourly = zero_array
+    if "pv_size_kw" in keys(r)
+        pv_kw_ac_hourly = r["pv_size_kw"] .* r["pv_production_factor_series"]
+    end
+    if microgrid_only && !Bool(get(r, "pv_migrogrid_upgraded", false))
```
hdunham, line 850: I don't think pv_kw_ac_hourly should be set to zero in this case if the user provided a pv_size_kw to overwrite the REopt size. Actually I don't really see the need for these microgrid only inputs at all. Seems like an unnecessary complication. If the user doesn't want to include PV in the calculations, they can just provide pv_size_kw = 0.

hdunham: I could see microgrid_only being used in backup_reliability_reopt_inputs along with the REopt decision variables of whether PV and battery are upgraded, but it doesn't make sense to have in backup_reliability_inputs

src/outagesim/backup_reliability.jl
```diff
@@ -591,40 +995,78 @@ function process_reliability_results(results)
...
-    -gen_operational_availability::Real = 0.9998 (Fraction of year generators not down for maintenance)
-    -gen_failure_to_start::Real = 0.0066 (Chance of generator starting given outage)
-    -gen_failure_to_run::Real = 0.00157 (Chance of generator failing in each hour of outage)
-    -num_gens::Int = 1  (Number of generators. Will be determined by code if set to 0 and gen capacity > 0.1)
-    -gen_capacity::Real = 0.0 (Backup generator capacity. Will be determined by REopt optimization if set less than 0.1)
-    -num_battery_bins::Int = 100 (Internal value for modeling battery)
+    -generator_operational_availability::Real = 0.9998 (Fraction of year generators not down for maintenance)
+    -generator_failure_to_start::Real = 0.0066 (Chance of generator starting given outage)
+    -generator_failure_to_run::Real = 0.00157 (Chance of generator failing in each hour of outage)
+    -num_generators::Int = 1  (Number of generators. Will be determined by code if set to 0 and gen capacity > 0.1)
```
hdunham, line 1025: These args can be vectors too right? In the case of multiple gen types

src/outagesim/backup_reliability.jl
```diff
@@ -591,40 +995,78 @@ function process_reliability_results(results)
...
+-pv_production_factor_series::Array        PV production factor per time step (required if pv_size_kw in dictionary)
+-pv_migrogrid_upgraded::Bool               If true then PV runs during outage if microgrid_only = TRUE (defaults to false)
+-battery_size_kw::Real                     Battery capacity. If no battery installed then PV disconnects from system during outage
+-battery_size_kwh::Real                    Battery energy storage capacity
+-charge_efficiency::Real                   Battery charge efficiency
+-discharge_efficiency::Real                Battery discharge efficiency
+battery_year_one_soc_series_pct
+#diesel_size_kw
+#diesel_size_mg_kw
+-generator_failure_to_start::Real = 0.0066                Chance of generator starting given outage
```
hdunham, line 1060: These gen args can be vectors too right? In the case of multiple gen types

src/outagesim/backup_reliability.jl
```diff
@@ -396,106 +655,111 @@ function survival_with_battery(net_critical_load::Vector, starting_batt_soc_kwh:
...
             init_soc = []
         end
 
-        starting_batt_soc_kwh = init_soc .* batt_kwh
+        starting_battery_soc_kwh = init_soc .* battery_size_kwh
 
-        #Only adds PV generation if there is also a battery
-        critical_loads_kw .-= pv_kw_ac_hourly
-        r2[:starting_batt_soc_kwh] = starting_batt_soc_kwh
+        #Only subtracts PV generation if there is also a battery
```
hdunham, line 754: Should put this assumption in backup_reliability and backup_reliability_inputs function documentation (@sericson0) and API Django input help text (me)

test/runtests.jl
```diff
@@ -189,6 +189,12 @@ else  # run HiGHS tests
      
     end
 
+    @testset "Backup Generator Reliability" begin
+        reliability_inputs = JSON.parsefile("./scenarios/backup_reliability_inputs.json")
+        reliability_results = backup_reliability(reliability_inputs)
```
hdunham, line 194: test other version of function as well

src/outagesim/backup_reliability.jl
hdunham: @sericson0 please address

test/runtests.jl
```diff
@@ -189,6 +189,16 @@ else  # run HiGHS tests
...
     end
 
+    @testset "Backup Generator Reliability" begin
+        p = REoptInputs("./scenarios/backup_reliability_inputs.json")
+        reopt_results = JSON.parsefile("./scenarios/backup_reliability_reopt_results.json")
+        reliability_inputs = Dict("gen_operational_availability" => 0.9998, "gen_failure_to_start" => 0.0066, "gen_failure_to_run" => 0.00157, "num_gen" => 2,
+        "gen_capacity_kw" => 200.0, "num_battery_bins" => 101, "max_outage_duration" => 96)
+
+        reliability_results = backup_reliability(reopt_results, p, reliability_inputs)
+        @test reliability_results["cumulative_duration_means"][96] ≈ 0.741 atol=0.01
```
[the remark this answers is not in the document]
hdunham: Simple verifiable tests have been added but we are still waiting on the more comprehensive validation that Jeff requested

src/outagesim/backup_reliability.jl
sericson0: Just added additional description. Let me know if it needs additional text to address

src/outagesim/backup_reliability.jl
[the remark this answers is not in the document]
hdunham: @sericson0 my comments don't appear to be addressed

src/outagesim/backup_reliability.jl
```diff
@@ -64,25 +98,59 @@ julia> markov_matrix(2, 0.1)
...
+ 0.1   0.0     0.225   0.0  0.0     0.675
+ 0.01  0.81    0.045   0.0  0.0     0.135
+ 0.0   0.25    0.0     0.0  0.75    0.0
+ 0.9   0.025   0.0     0.0  0.075   0.0
+ 0.18  0.0025  0.2025  0.0  0.0075  0.6075
+```
+"""
+function markov_matrix(num_gen_vec::Vector{Int}, fail_prob_vec::Vector{<:Real})::Matrix{Float64} 
+
+    num_gen_working = reshape(collect(Iterators.product((0:g for g in num_gen_vec)...)), :, 1)
```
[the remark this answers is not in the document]
hdunham: adding myself and resolving

src/outagesim/backup_reliability.jl
hdunham: Sorry I didn't notice this before, but why does this produce a vector instead of the multiple generator outputs being summed and the battery shifted by the that total?

src/outagesim/backup_reliability.jl
hdunham: @sericson0 I'm not seeing how my comment was addressed

src/outagesim/backup_reliability.jl
hdunham: @sericson0 this doesn't appear to be addressed yet

src/outagesim/backup_reliability.jl
hdunham: @sericson0 this is not yet addressed

src/outagesim/backup_reliability.jl
hdunham: the way it's being done is correct because starting_probabilities uses generator_operational_availability and generator_failure_to_start, while this loop uses generator_failure_to_run

src/outagesim/backup_reliability.jl
hdunham: @sericson0 can you address this? You replaced most "hours" with "time_steps", but I still see one, as well as documentation that says certain vectors must be 8760s.

src/outagesim/backup_reliability.jl
```diff
@@ -505,126 +768,325 @@ function backup_reliability_inputs(d::Dict, p::REoptInputs; r::Dict = Dict())::D
...
+    generator_size_kw::Union{Real, Vector{<:Real}} = 0.0, 
+    num_battery_bins::Int = 101,
+    max_outage_duration::Int = 96,
+    battery_size_kw::Real = 0.0,
+    battery_size_kwh::Real = 0.0,
+    battery_charge_efficiency::Real = 0.948, 
+    battery_discharge_efficiency::Real = 0.948,
+    kwargs...)::Array
+ 
+    total_gen_cap = sum(generator_size_kw)
```
hdunham, line 938: unused variable

src/outagesim/backup_reliability.jl
sericson0: Pushed an update to make replacements

src/outagesim/backup_reliability.jl
sericson0: Was an artifact of a previous version. Just pushed an update which removes it

src/outagesim/backup_reliability.jl
hdunham: I'm still seeing references to 8760

src/outagesim/backup_reliability.jl
sericson0: Was talking with Bill about this today that the naming of Marginal versus Cumulative is not ideal. The cumulative means that the system survives up to and including the outage while the marginal means the system can survive in the given outage hour. Unfortunately the cumulative survivability is not the product of the marginal survivals in each hour. I definitely agree that it is inefficient to run the code twice if there were a way to calculate in 1 go, just not sure how to do it.

src/outagesim/backup_reliability.jl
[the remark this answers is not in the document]
sericson0: Shoot, made the changes for the Markov matrix one and forgot to do it for the transition probability as well. Pushing update now

src/outagesim/backup_reliability.jl
sericson0: From a conversation we had I thought the comment was to add the [end, :] to the Markov matrix instead of the original way of calculating M, then multiplying by G to get the end row? Apologies if I misunderstood. 

Or were you thinking of doing the calculation without creating the Markov matrix to begin with? May be slightly better but we already have the function for making the Markov matrix, the matrix is pretty small and does not take much time to create, and only gets made once.

src/outagesim/backup_reliability.jl
sericson0: Added the battery charge and discharge efficiencies to r2. Not sure how best to address the problem of having multiple storage types. Would it make sense to only allow a single storage type?

hdunham: resolving because only allowing one elec storage type is good

src/outagesim/backup_reliability.jl
hdunham: Agreed not that important. Resolving.

Bill-Becker:
@hdunham I started to merge develop into this branch but then aborted because there were some conflicting changes I believe you made where I didn't know how to merge with develop. The conflicts are likely caused by the recent big name change of results (year_one -> annual). Can you work on merging that?

Bill-Becker (approved):
It seems like we're at a good point to merge stuff in and make a new PR with the next round of updates (Wind, fuel/RE, etc).

Bill-Becker:
@hdunham can you merge this into develop before our meeting today? We can then work together on merge into master and registering the new version to use in the erp API, and merging that.

PR #100, opened 2022-07-13
update version number for registration, update AVERT to 2021 data, minor other changes to emissions implementation

hdunham:
This PR makes a few changes/additions to the previous PR for emissions capabilities (#28). 
    - update to use 2021 AVERT data, which also has different regions, and update tests accordingly
    - make ExistingBoiler immutable again (the changes to allow this were in #28)
    - update struct comments and field ordering for more consistency
    - change name lifecycle_emissions_cost_CO2 to lifecycle_emissions_cost_climate, move lifecycle_emissions_cost_climate and lifecycle_emissions_cost_health from site to financial results, and update tests accordingly
    - update REopt version number in Project.toml in order for new version to be registered

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "REopt"
 uuid = "d36ad4e8-d74a-4f7a-ace1-eaea049febf6"
-authors = ["Nick Laws <nick.laws@nrel.gov>", "Hallie Dunham <hallie.dunham@nrel.gov>", "Bill Becker <william.becker@nrel.gov>", "Bhavesh Rathod <bhavesh.rathod@nrel.gov>", "Alex Zolan <alexander.aolan@nrel.gov>", "Amanda Farthing <amanda.farthing@nrel.gov>"]
+authors = ["Nick Laws", "Hallie Dunham <hallie.dunham@nrel.gov>", "Bill Becker <william.becker@nrel.gov>", "Bhavesh Rathod <bhavesh.rathod@nrel.gov>", "Alex Zolan <alexander.aolan@nrel.gov>", "Amanda Farthing <amanda.farthing@nrel.gov>"]
 version = "0.17.0"
 
 [deps]


hdunham:
Still working on getting the 2021 AVERT data working but GitHub won't let me convert this to a draft.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "REopt"
 uuid = "d36ad4e8-d74a-4f7a-ace1-eaea049febf6"
-authors = ["Nick Laws <nick.laws@nrel.gov>"]
-version = "0.16.2"
+authors = ["Nick Laws <nick.laws@nrel.gov>", "Hallie Dunham <hallie.dunham@nrel.gov>", "Bill Becker <william.becker@nrel.gov>", "Bhavesh Rathod <william.becker@nrel.gov>", "Alex Zolan <alexander.aolan@nrel.gov>", "Amanda Farthing <amanda.farthing@nrel.gov>"]
```
rathod-b, line 3: Seems like my email is incorrect. For emails, should we move Nick's email out or towards the end of this list now?

src/results/financial.jl
```diff
@@ -53,6 +53,8 @@
 - `om_and_replacement_present_cost_after_tax` Present value of all O&M and replacement costs, after tax.
 - `developer_om_and_replacement_present_cost_after_tax` Present value of all O&M and replacement costs incurred by developer, after tax.
 - `offgrid_microgrid_lcoe_dollars_per_kwh` Levelized cost of electricity for modeled off-grid system.
+- `lifecycle_emissions_cost_climate` LCC component if Settings input include_climate_in_objective is true. Present value of CO2 emissions cost over the analysis period.
```
rathod-b, line 56: Should these docstrings be updated to maybe say 'Present value of CO2 emissions cost over the analysis period. Minimized LCC component if Settings input include_climate_in_objective is true' since we minimize climate and health costs depending on include_climate_in_objective and include_health_in_objective?

It seems like we are always adding these two fields to the model.

adfarth:
> No description provided.

@hdunham would you mind updating the description of this PR? will help with review!

Project.toml
adfarth: updated your email in https://github.com/NREL/REopt.jl/pull/100/commits/12498dd5c145826d281f4f659365cfda48c5484a. In my opinion it's fine (and most accurate currently) to leave Nick's name at the front, but we should probably remove his email because it's no longer active.

src/results/financial.jl
adfarth: Second proposal: 
"Present value of CO2 emissions cost over the analysis period. LCC component (i.e., included in LCC) only if Settings input include_climate_in_objective is true."

adfarth (approved):
Looks good! Noting here that the "failing" tests are just duplicate tests (for the commit, duplicate of the PR tests) that Hallie cancelled.

Bill-Becker (approved):

PR #101, opened 2022-07-19
Avoid triggering duplicate GitHub workflows

hdunham:
When pushing to a branch that's in a PR, only trigger tests on the push not on the PR sync also.


adfarth (approved):
Looks good and will save lots of testing time!

PR #106, opened 2022-07-26
Update electric_storage.jl and doc strings

rathod-b:
Update default degradation parameters per updated parameters. 

Also update several doc strings for input descriptions.


adfarth:
I piggy backed on this PR to include a very minor update to the wind doc string https://github.com/NREL/REopt.jl/pull/106/commits/3acc26a6ce0e1578c484801e14ee464ac132f9cd

src/constraints/battery_degradation.jl
```diff
@@ -196,11 +254,15 @@ function add_degradation(m, p; b="ElectricStorage")
         )
         # add augmentation cost to objective
         # maintenance_cost_per_kwh must have length == length(days) - 1, i.e. starts on day 2
+
+        # create salvage value expression for objective, all 0s since we have no salvage value
+        s = zeros(length(months))
+        @expression(m, salv_value, sum(s[mth] for mth in months))
```
zolanaj, line 260: why not just `@expression(m, salv_value, 0.0)`?

src/results/electric_storage.jl
```diff
@@ -64,6 +64,7 @@ function add_electric_storage_results(m::JuMP.AbstractModel, p::REoptInputs, d::
                 r["replacement_month"] = Int(value(
                     sum(mth * m[:bmth][mth] for mth in 1:p.s.financial.analysis_years*12)
                 ))
+                r["residual_value"] = value(m[:salv_value])
             end
```
zolanaj, line 68: Should we include alternative values for these if the strategy is `"augmentation"` or `p.s.storage.attr[b].model_degradation` is False?  Or do we only output these for the degradation model?  Thinking of how this should be implemented in the API endpoint.

zolanaj (changes requested):
@rathod-b Looks good, thanks for these updates:  A few comments beyond the in-line ones: 
- The docs for the degradation model still need to be updated; there are TODOs in the code but we might want that as an open issue just so we don't forget it.
- It would be good to make a more robust collection of tests.  We can use the new file  degradation_tests.json for this since it's not currently used.  We should include a test case for replacement (which we already have, but please add an assertion for the number of replacements or a new degradation-specific measure) and a test case for augmentation (which we don't have yet).

Let me know if you want help with any of these.  Thanks!
Alex

src/constraints/battery_degradation.jl
```diff
@@ -196,11 +186,12 @@ function add_degradation(m, p; b="ElectricStorage")
...
+
+        add_to_expression!(m[:TotalStorageCapCosts], p.third_party_factor*(m[:degr_cost]))
+
     else
         @error "Battery maintenance strategy $strategy is not supported. Choose from augmentation and replacement."
     end
-
-    @objective(m, Min, m[:Costs] + m[:degr_cost])
     
     # NOTE adding to Costs expression does not modify the objective function
```
rathod-b, line 196: @zolanaj does the note on this line mean anything? Is `add_to_expression!` the same as adding degradation costs to `objective` function? When I test with adding degradation cost to objective, I get no replacements for a scenario which gives replacements otherwise.

src/constraints/battery_degradation.jl
```diff
@@ -114,77 +118,63 @@ function add_degradation(m, p; b="ElectricStorage")
...
-         ------|----|------|----|-----------------|->  d_0p8
-              N/5  N/4    N/3  N/2                N = 365*analysis_years
+        ElectricStorage.degradation.maintenance_strategy = \"replacement\". 
+        Not all solvers support indicators and some are slow with integers."
+        
+        @variable(m, soh_indicator[months], Bin) # track SOH levels, 1 if SOH >= 80%, 0 otherwise
+        @variable(m, bmth[months], Bin) # track which month SOH indicator drops to < 80%
+        @variable(m, 0 <= bmth_BkWh[months]) # track the kwh replacement value
+
+        M = 10*maximum(p.s.electric_load.loads_kw)  # the big M
```
rathod-b, line 130: Note: Current big-M. This value seems to be giving good consistent results. Rationale is that a battery wont be larger than the peak load so 10x peak load is large enough. Even this value might be larger than expected. @zolanaj should we sweep across a range of big M values (i.e. 0.1x, 1.0x, 10x peak load) to fine tune this parameter?

src/constraints/battery_degradation.jl
zolanaj: @rathod-b I believe this comment is intended to say that if you change `m[:Costs]` after this line, then it doesn't change the objective and the expression and objective function won't match anymore.

Good to know on the SOC objective; I agree that should be removed if we're running degradation.

src/constraints/battery_degradation.jl
zolanaj: @rathod-b I checked the API and we never introduced a practical maximum size for batteries because they didn't have integer variables involved prior to the introduction of degradation.  

One heuristic would be power = 2x peak load, and energy = 12x peak load (i.e., allowing you to charge for half the day, and discharge for the other half of the day).  This precludes long-term storage systems, but we can probably include a user setting for long-term storage.

Related: I'll check our issues for importing big-M's from the API Julia code and add that issue if it's not there.

zolanaj: big-M issue is here: https://github.com/NREL/REopt.jl/issues/114

zolanaj: Actually, a system that has oversized PV to power the site load might need more.  just do 24x peak load for both power and energy, that should be enough.

src/core/energy_storage/electric_storage.jl
```diff
@@ -151,9 +151,9 @@ The following shows how one would use the degradation model in REopt via the [Sc
 Note that not all of the above inputs are necessary. When not providing `calendar_fade_coefficient` for example the default value will be used.
 """
 Base.@kwdef mutable struct Degradation
-    calendar_fade_coefficient::Real = 2.46E-03
-    cycle_fade_coefficient::Real = 7.82E-05
-    time_exponent::Real = 0.5
+    calendar_fade_coefficient::Real = 2.55E-03
+    cycle_fade_coefficient::Real = 9.83E-05
+    time_exponent::Real = 0.42
```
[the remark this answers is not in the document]
rathod-b: Yes, this is intentional update of default degradation parameters per battery SOH modeling done by Paul Gasper. Let me confirm on how to cite this source.

src/constraints/battery_degradation.jl
rathod-b: addressed in https://github.com/NREL/REopt.jl/commit/1b2cfff04bc79fb7a51c06462116e70eeaa0576e

src/results/electric_storage.jl
rathod-b: Good question. Currently these fields are only in the output if degradation is true. These fields can be handled in the API by creating nullable and blankable output fields so API would catch them if REopt were to return them.

test/scenarios/degradation_tests.json
```diff
@@ -0,0 +1,18 @@
+{
```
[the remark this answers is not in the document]
rathod-b: Will remove this JSON file in next push

src/constraints/battery_degradation.jl
rathod-b: I got rid of this expression. If we have to include salvage value in the objective, we can do it via `reopt.jl` file for replacement strategy only.

src/constraints/battery_degradation.jl
rathod-b: Updated big-M to be set based on max_kwh. Is it normal for changes in big-M to change the results? I see slight differences  in electric storage results.

test/test_with_xpress.jl
```diff
@@ -400,6 +400,44 @@ check to make sure that PV does NOT export unless the site load is met first for
               if results["PV"]["year_one_to_grid_series_kw"][i] > 0)
 end
 
+@testset "Battery degradation replacement strategy" begin
```
rathod-b, line 403: @zolanaj I added this test to test the outputs as well as model expression values. Since solve time is not an issue I am sticking with a simple LargeOffice load profile.

I had to change the result JSON tests to inequalities because we get different results on github action than those when testing this scenario by itself.

rathod-b:
> @rathod-b Looks good, thanks for these updates: A few comments beyond the in-line ones:
> 
> * The docs for the degradation model still need to be updated; there are TODOs in the code but we might want that as an open issue just so we don't forget it.
> * It would be good to make a more robust collection of tests.  We can use the new file  degradation_tests.json for this since it's not currently used.  We should include a test case for replacement (which we already have, but please add an assertion for the number of replacements or a new degradation-specific measure) and a test case for augmentation (which we don't have yet).
> 
> Let me know if you want help with any of these. Thanks! Alex

Hi Alex, thanks for your comments on the PR! Its been a while and I am picking this back up. I have removed the many TODOs from the code documentation which were initially added as reminders for me. If you see any more of them, do let me know.

I have also updated the replacement strategy testset to validate some degradation specific parameters to ensure the results make sense and DVs arent switching between 0 and 1 as they were when we were testing this back in Fall 2022. Let me know if there are any comments on that.

Lastly, I have an open question for you. At present, we add the SOC incentive to the objective function in `run_reopt.jl` without any consideration for BESS degradation. I was wondering if we wanted to handle that programmatically, or handle it via a warning to a user and have then set the add_soc_incentive to false when running degradation?

rathod-b:
> > @rathod-b Looks good, thanks for these updates: A few comments beyond the in-line ones:
> > 
> > * The docs for the degradation model still need to be updated; there are TODOs in the code but we might want that as an open issue just so we don't forget it.
> > * It would be good to make a more robust collection of tests.  We can use the new file  degradation_tests.json for this since it's not currently used.  We should include a test case for replacement (which we already have, but please add an assertion for the number of replacements or a new degradation-specific measure) and a test case for augmentation (which we don't have yet).
> > 
> > Let me know if you want help with any of these. Thanks! Alex
> 
> Hi Alex, thanks for your comments on the PR! Its been a while and I am picking this back up. I have removed the many TODOs from the code documentation which were initially added as reminders for me. If you see any more of them, do let me know.
> 
> I have also updated the replacement strategy testset to validate some degradation specific parameters to ensure the results make sense and DVs arent switching between 0 and 1 as they were when we were testing this back in Fall 2022. Let me know if there are any comments on that.
> 
> Lastly, I have an open question for you. At present, we add the SOC incentive to the objective function in `run_reopt.jl` without any consideration for BESS degradation. I was wondering if we wanted to handle that programmatically, or handle it via a warning to a user and have then set the add_soc_incentive to false when running degradation?

I could also use some help in fixing the rendering of a formula in documentation. Line 145 in `electric_storage.jl` has the broken formula for residual value.

src/constraints/battery_degradation.jl
```diff
@@ -118,131 +118,62 @@ function add_degradation(m, p; b="ElectricStorage")
...
-        @constraint(m, [mth in months], bmth_BkWh[mth] >= m[:dvStorageEnergy][b] - M*(1-bmth[mth]))
-
-        # add replacment cost to objective
+        # create replacement cost expression for objective
         @expression(m, degr_cost, sum(c[mth] * bmth_BkWh[mth] for mth in months))
 
-        # create salvage value expression for objective
-        @expression(m, salv_value, sum(s[mth] * bmth_BkWh[mth] for mth in months))
+        # create residual value expression for objective
+        @expression(m, residual_value, sum(s[mth] * bmth_BkWh[mth] for mth in months))
```
zolanaj: is maintenance_cost_per_kwh actually a replacement cost?  we may want to rename it for clarity (so itn's not confused with O&M costs).

test/test_with_xpress.jl
zolanaj: @rathod-b some values like the SOH indicator at the end are going to be stable while a lot of these others will change with solver settings.  We can construct a more unit-test-like load profile to incentivize a certain behavior and make the solution more stable (something like a flat load with hourly cost that oscillates from very expensive to free every other period.)

zolanaj:
> > > @rathod-b Looks good, thanks for these updates: A few comments beyond the in-line ones:
> > > 
> > > * The docs for the degradation model still need to be updated; there are TODOs in the code but we might want that as an open issue just so we don't forget it.
> > > * It would be good to make a more robust collection of tests.  We can use the new file  degradation_tests.json for this since it's not currently used.  We should include a test case for replacement (which we already have, but please add an assertion for the number of replacements or a new degradation-specific measure) and a test case for augmentation (which we don't have yet).
> > > 
> > > Let me know if you want help with any of these. Thanks! Alex
> > 
> > 
> > Hi Alex, thanks for your comments on the PR! Its been a while and I am picking this back up. I have removed the many TODOs from the code documentation which were initially added as reminders for me. If you see any more of them, do let me know.
> > I have also updated the replacement strategy testset to validate some degradation specific parameters to ensure the results make sense and DVs arent switching between 0 and 1 as they were when we were testing this back in Fall 2022. Let me know if there are any comments on that.
> > Lastly, I have an open question for you. At present, we add the SOC incentive to the objective function in `run_reopt.jl` without any consideration for BESS degradation. I was wondering if we wanted to handle that programmatically, or handle it via a warning to a user and have then set the add_soc_incentive to false when running degradation?
> 
> I could also use some help in fixing the rendering of a formula in documentation. Line 145 in `electric_storage.jl` has the broken formula for residual value.



> Lastly, I have an open question for you. At present, we add the SOC incentive to the objective function in `run_reopt.jl` without any consideration for BESS degradation. I was wondering if we wanted to handle that programmatically, or handle it via a warning to a user and have then set the add_soc_incentive to false when running degradation?

A warning should be fine here - these different objectives can be handled by the solver.

zolanaj (approved):
@rathod-b I made the degradation test a bit simpler and did some renaming of parameters and variables for style and clarity.  If you agree with these and tests pass, I think this can be merged.  (The HiGHS tests are timing out but I think everything else is passing.)

rathod-b:
Relevant changes from this PR were merged into develop via #443. This PR is being closed.

PR #111, opened 2022-08-03
allow user to specify emissions_region (overrides lat/long if using AVERT data) and emissions factor type conversion

hdunham:
- Allow user to specify emissions_region as an ElectricUtility input, which overrides lat/long if looking up AVERT data when emissions factors not provided by user.
- Add emissions_factor_series_lb_<pollutant>_per_kwh inputs to list of inputs that dictkeys_tosymbols() tries to convert to type Array{Real}. Due to serialization, when list inputs come from the API, they are of type Array{Any} so must be converted to match type required by the constructors they are passed to.
- Remove unused test input JSONs.


src/core/electric_utility.jl
```diff
@@ -43,10 +43,11 @@
     outage_time_steps::Union{Nothing, UnitRange} = isempty(outage_durations) ? nothing : 1:maximum(outage_durations),
     scenarios::Union{Nothing, UnitRange} = isempty(outage_durations) ? nothing : 1:length(outage_durations),
     # Emissions and renewable energy inputs:
-    emissions_factor_series_lb_CO2_per_kwh::Union{Real,Array{<:Real,1}} = Float64[],
-    emissions_factor_series_lb_NOx_per_kwh::Union{Real,Array{<:Real,1}} = Float64[],
-    emissions_factor_series_lb_SO2_per_kwh::Union{Real,Array{<:Real,1}} = Float64[],
-    emissions_factor_series_lb_PM25_per_kwh::Union{Real,Array{<:Real,1}} = Float64[],
+    emissions_region::String = "", # AVERT emissions region. Default is based on location, or can be overriden by providing region here.
+    emissions_factor_series_lb_CO2_per_kwh::Union{Real,Array{<:Real,1}} = Float64[], # can be scalar or timeseries (aligned with time_steps_per_hour)
```
adfarth, line 47: @hdunham  I added these comments--can you review?

adfarth (approved):
@hdunham -- all looks good to me! I will approve but I think it would be good to add a test for this change if you have time.

src/core/electric_utility.jl
hdunham: great thanks!

hdunham:
> @hdunham -- all looks good to me! I will approve but I think it would be good to add a test for this change if you have time. (There is also a conflict with the CHANGELOG that needs to be addressed before merging)

I resolved the CHANGELOG conflict and modified the existing emissions tests to use emissions_region in one of the scenarios.

PR #117, opened 2022-08-05
GhpGhx package separation

Bill-Becker:
This makes GhpGhx a **conditional** dependency of REopt, so that REopt can be loaded without first doing "add https://github.com/NREL/GhpGhx.jl" and then "using GhpGhx" in the users active Julia environment. In order to run GhpGhx functions to size the GHE and dispatch, you must first do those two commands. Before this PR, the user **must** run those two commands just to load REopt.

The Requires.jl package (added here to the REopt environment) @requires macro detects if the user has loaded the GhpGhx package to the running Julia environment, and it loads the package within REopt to allow those GhpGhx function to work in `scenario.jl`. A `try-catch` wrapper was put around the GhpGhx function calls in `scenario.jl` to throw a useful error if they try to run GhpGhx without first loading the package.

This PR changes the GHP test which previously included the function calls to GhpGhx; the GHP test now relies on a *previously* run GhpGhx model and uses the `ghpghx_response` from that run. I could not get the `Pkg.test` environment (for Git workflows CI, and eventually the Julia registry (I think) to run the test with the conditionally loaded GhpGhx (I tried many different ways!) package. However, I did also add a test for using multiple GHP options (inputting a list of 2 dicts for ghpghx_responses) which then tests the expected REopt choice between those two options.

Added to `docs` for how to load the GhpGhx module, by doing those two commands.


CHANGELOG.md
```diff
@@ -4,9 +4,10 @@
...
-    - The `GHP` size and dispatch of the different `GHP` options is pre-determined by the `GhpGhx.jl` package, so the REopt model just chooses one or none of the GHP options with a binary decision variable.
+- Add geothermal heat pump (`GHP`), also known as ground-source heat pump (GSHP), to the REopt model for serving heating and cooling loads (typically the benefits include electrifying the heating load and improving the efficiency of cooling).
+    - The unregistered `GhpGhx` package (https://github.com/NREL/GhpGhx.jl) is a "conditional" dependency of REopt by using the Requires.jl package, and this package sizes the ground heat exchanger (GHE) and gets the hourly electric consumption of the `GHP` for the specified heating and cooling loads that it serves.
+    - The `GhpGhx` module calls for sizing the GHE can only be done if you first "add https://github.com/NREL/GhpGhx.jl" to the environment and then load the package by "using GhxGhx" before running REopt with `GHP`.
```
zolanaj: Also, maybe say:

first add https://github.com/NREL/GhpGhx.jl" to the environment **from the package manager** and then...

zolanaj (approved):
The tests I performed on this were as follows: 

(1) Run non-GHP scenario, no GhpGhx.jl installed, run successful [x]
(2) Run GHP test scenario that reads in the GHP response, no GhpGhx.jl installed, run successful [x]
(3) Run GHP test scenario without populating GHP properties from ghpghx_response.json and ghp_cop_map_custom.csv, no GhpGhx.jl installed, get the LoadError that you need the library [x]
(4) Run GHP test scenario without populating GHP properties from ghpghx_response.json and ghp_cop_map_custom.csv, with GhpGhx.jl installed, run successful [x]

So this looks great, thanks @Bill-Becker!  I made a couple of minor changes to the readme and docs to clarify the location of the install guide and the capcabilities if you don't install GhpGhx.jl.  If these look good, please feel free to merge!

Bill-Becker:
Thanks a lot @zolanaj. Great that you tested and it worked as you expected. I'm glad I could merge this before I head out until Friday next week.

Issue #14, opened 2022-02-10
transfer AbsorptionChiller models from API to REopt.jl


zolanaj:

Bill-Becker:
@zolanaj Did the default costs and COP logic for `AbsorptionChiller` get integrated when you added `AbsorptionChiller`? See starting [here in the API](https://github.com/NREL/REopt_API/blob/4b1a154ab82380f39ee16d12b6047e279f67a944/reo/src/techs.py#L556). I'm not seeing this.

PR #126, opened 2022-09-11
Abs chill mods

Bill-Becker:
After `AbsorptionChiller` was uncommented from @testset "Heat and cool energy balance", the first test for "cooling production from techs equals the cooling load plus TES losses" is failing. The test is showing that there is excessive amounts of TES losses (TES charge energy minus TES discharge energy), and the techs are not supplying that extra loss. When `AbsorptionChiller` is not included in the test, the model is not showing these excessive losses, so somehow the `AbsorptionChiller` is allowing unserved TES losses.

Note that I changed some variable naming, and the Hot and Cold TES sizes for this test, but that is not consequential to the test result because it's an energy balance test.

### DIFF
--- a/src/core/heating_cooling_loads.jl
+++ b/src/core/heating_cooling_loads.jl
@@ -339,21 +339,25 @@ struct CoolingLoad
                 or the site_electric_load_profile along with one of [per_time_step_fractions_of_electric_load, monthly_fractions_of_electric_load, annual_fraction_of_electric_load].")
         end
 
-        # Now that either the cooling electric loads_kw or the cooling thermal loads_kw_thermal is known, update existing_chiller_cop if it was not input
-        if isnothing(existing_chiller_cop)
-            existing_chiller_cop = get_existing_chiller_default_cop(;existing_chiller_max_thermal_factor_on_peak_load=existing_chiller_max_thermal_factor_on_peak_load,
-                                loads_kw=loads_kw, loads_kw_thermal=loads_kw_thermal)
-        end
-
         if isnothing(loads_kw_thermal)  # have to convert electric loads_kw to thermal load
-            if !isnothing(annual_tonhour) || !isempty(monthly_tonhour)
+            if (!isnothing(annual_tonhour) || !isempty(monthly_tonhour)) && isnothing(existing_chiller_cop)
                 # cop_unknown_thermal (4.55) was used to convert thermal to electric in BuiltInCoolingLoad, so need to use the same here to convert back
                 #  in order to preserve the input tonhour amounts - however, the updated/actual existing_chiller_cop is still assigned based on user input or actual max ton load conditional defaults
                 chiller_cop = get_existing_chiller_default_cop()
+            elseif (!isempty(doe_reference_name) || !isempty(blended_doe_reference_names)) || isnothing(existing_chiller_cop)
+                # Generated loads_kw (electric) above based on the building's default fraction of electric profile
+                chiller_cop = get_existing_chiller_default_cop(;existing_chiller_max_thermal_factor_on_peak_load=existing_chiller_max_thermal_factor_on_peak_load, 
+                                        loads_kw=loads_kw)
             else
                 chiller_cop = existing_chiller_cop
             end
             loads_kw_thermal = chiller_cop * loads_kw
+        end
+
+        # Now that cooling thermal loads_kw_thermal is known, update existing_chiller_cop if it was not input
+        if isnothing(existing_chiller_cop)
+            existing_chiller_cop = get_existing_chiller_default_cop(;existing_chiller_max_thermal_factor_on_peak_load=existing_chiller_max_thermal_factor_on_peak_load, 
+                                        loads_kw_thermal=loads_kw_thermal)
         end
 
         if length(loads_kw_thermal) < 8760*time_steps_per_hour
--- a/src/core/reopt_inputs.jl
+++ b/src/core/reopt_inputs.jl
@@ -651,6 +651,11 @@ function setup_absorption_chiller_inputs(s::AbstractScenario, max_sizes, min_siz
     end
 
     cop["AbsorptionChiller"] = s.absorption_chiller.cop_electric
+    if isnothing(s.chp)
+        thermal_factor = 1.0
+    else
+        thermal_factor = s.chp.cooling_thermal_factor
+    end    
     thermal_cop["AbsorptionChiller"] = s.absorption_chiller.cop_thermal * thermal_factor
     om_cost_per_kw["AbsorptionChiller"] = s.absorption_chiller.om_cost_per_kw
     return nothing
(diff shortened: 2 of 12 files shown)


zolanaj:
@Bill-Becker looks like the missing defaults you mentioned in a separate comment in https://github.com/NREL/REopt.jl/issues/14 is at least part of the problem.  Let me look into this today and follow up?

Bill-Becker:
@zolanaj sounds good, thanks!

src/results/existing_chiller.jl
```diff
@@ -39,7 +39,7 @@ function add_existing_chiller_results(m::JuMP.AbstractModel, p::REoptInputs, d::
     r = Dict{String, Any}()
 
 	@expression(m, ELECCHLtoTES[ts in p.time_steps],
-		sum(m[:dvProductionToStorage][b,"ExistingChiller",ts] for b in p.s.storage.types.cold, t in p.techs.cooling)
```
Bill-Becker, line 42: @zolanaj This was the issue. Just a results thing. When both ExistingChiller and AbsorptionChiller are included in techs.cooling, this ELECCHLtoTES expression was double counted.

zolanaj: @Bill-Becker that'll definitely do it, thanks for catching this!

zolanaj: @Bill-Becker is this still a draft, or can I review more thoroughly?

Bill-Becker: I've made the changes I wanted to make, and then I wasn't sure if we wanted to just leave this PR as a placeholder for adding the default cost/cop data and logic. If you want to address that later, this is ready to review.

zolanaj: @Bill-Becker thanks, I'll review today in that case, so that we can get the bug fix out (the lack of defaults and cost curves make it a simpler cost model for the Absorption chiller, but it's functional, and we have the issue in place to make sure that feature is added. I'll assign the issue to myself, mark this ready for review, and follow up later today.

src/core/heating_cooling_loads.jl
```diff
@@ -339,23 +339,27 @@ struct CoolingLoad
...
-                                loads_kw=loads_kw, loads_kw_thermal=loads_kw_thermal)
-        end
-
         if isnothing(loads_kw_thermal)  # have to convert electric loads_kw to thermal load
-            if !isnothing(annual_tonhour) || !isempty(monthly_tonhour)
+            if (!isnothing(annual_tonhour) || !isempty(monthly_tonhour)) && isnothing(existing_chiller_cop)
                 # cop_unknown_thermal (4.55) was used to convert thermal to electric in BuiltInCoolingLoad, so need to use the same here to convert back
                 #  in order to preserve the input tonhour amounts - however, the updated/actual existing_chiller_cop is still assigned based on user input or actual max ton load conditional defaults
                 chiller_cop = get_existing_chiller_default_cop()
+            elseif (!isempty(doe_reference_name) || !isempty(blended_doe_reference_names)) || isnothing(existing_chiller_cop)
```
zolanaj, line 347: minor formatting nit: with serial 'or' statements, you can remove the parentheses wrapping around the first two bools

src/core/scenario.jl
```diff
@@ -80,6 +80,7 @@ All values of `d` are expected to be `Dicts` except for `PV` and `GHP`, which ca
     handle conversion of Vector of Vectors from JSON to a Matrix in Julia).
 """
 function Scenario(d::Dict; flex_hvac_from_json=false)
+    d = deepcopy(d)
```
zolanaj, line 83: do we need this, and if so why?

test/test_with_xpress.jl
```diff
@@ -685,54 +688,53 @@ end
...
     @test round(total_thermal_mmbtu_calculated, digits=0) ≈ total_thermal_expected atol=8.0  
 
-    # Test CHP.cooling_thermal_factor = 0.8, AbsorptionChiller.chiller_cop = 0.7 (from test_cold_POST.json)
-    # absorpchl_heat_in_kwh = results["AbsorptionChiller"]["year_one_absorp_chl_thermal_consumption_mmbtu"] * 1.0E6 / 3412.0
-    # absorpchl_cool_out_kwh = results["AbsorptionChiller"]["year_one_absorp_chl_thermal_production_tonhr"] * REopt.KWH_THERMAL_PER_TONHOUR
-    #absorpchl_cop = absorpchl_cool_out_kwh / absorpchl_heat_in_kwh
+    # Test CHP["cooling_thermal_factor"] = 0.8, AbsorptionChiller["cop_thermal"] = 0.7 (from inputs .json)
+    absorpchl_heat_in_kwh = results["AbsorptionChiller"]["year_one_thermal_consumption_mmbtu"] * REopt.KWH_PER_MMBTU
+    absorpchl_cool_out_kwh = results["AbsorptionChiller"]["year_one_thermal_production_tonhour"] * REopt.KWH_THERMAL_PER_TONHOUR
+    absorpchl_cop = absorpchl_cool_out_kwh / absorpchl_heat_in_kwh
```
zolanaj, line 736: For near-future reference: I may add a check here for thermal load balancing between hot/cold thermal load met and the thermal loads we will populate in results.

src/core/reopt_inputs.jl
```diff
@@ -650,8 +650,16 @@ function setup_absorption_chiller_inputs(s::AbstractScenario, max_sizes, min_siz
...
     end
 
-    cop["AbsorptionChiller"] = s.absorption_chiller.chiller_elec_cop
-    thermal_cop["AbsorptionChiller"] = s.absorption_chiller.chiller_cop
+    cop["AbsorptionChiller"] = s.absorption_chiller.cop_electric
+    if isnothing(s.chp)
+        thermal_factor = 1.0
+    elseif s.chp.cooling_thermal_factor == 0.0
+        error("The CHP cooling_thermal_factor is 0.0 which implies that CHP cannot serve AbsorptionChiller. If you
+            want to model CHP and AbsorptionChiller, you must specify a cooling_thermal_factor greater than 0.0")
```
zolanaj, line 658: Do we want to support a case with a CHP system that can meet heating load but not serve the absorption chiller (which could be served by the boiler), in which case we'd switch `error` with `warn`?

zolanaj (approved):
Thanks @Bill-Becker for these mods and fixes!  I have a couple small questions here but nothing that requires seeing this again before merging.

src/core/scenario.jl
Bill-Becker: It's not absolutely necessary, but this function actually mutates the inputs dict `d`; this deepcopy avoids that mutation. I think it's cleaner in a workflow where you define that inputs_dict and expect it to be preserved for later use (e.g. running again with the same or slightly modified inputs_dict, or comparing the raw inputs with processed inputs).

src/core/reopt_inputs.jl
Bill-Becker: I agree that should/could be a possibility, but that would be strange to put in a CHP system and not have it be able to serve the absorption chiller.

Bill-Becker:
@zolanaj thanks for the review and adding the CHANGELOG.md stuff. I'm going to merge.

PR #146, opened 2022-11-04
Update Absorption Chiller Defaults

zolanaj:
- Adds default values for the absorption chiller which depend on size and heat transfer medium (i.e., hot water or steam)
- Exports absorption chiller defaults function for use in API (mimics that of CHP and SteamTurbine technologies, which also have stored defaults)
- Tests added to check absorption chiller default values

Addresses Issue https://github.com/NREL/REopt.jl/issues/127


src/core/absorption_chiller.jl
```diff
@@ -28,61 +28,217 @@
...
-            max_kw,
-            installed_cost_per_kw,
-            om_cost_per_kw
+return a Dict{String, Any} by selecting the appropriate values from 
+data/chp/absorption_chiller_defaults.json, which contains values based on heat transfer medium (hot_water_or_steam)
+such as:
+- "installed_cost_per_ton"
+- "om_cost_per_ton"
+- "cop_thermal"
+- "tech_sizes_for_cost_curve"
```
Bill-Becker, line 140: Are we modeling a cost curve (y-intercept + slope) for AbsorptionChiller, or just linearly interpolating between the cost data points between two adjacent sizes? If not, I suggest we rename to something like tech_sizes_for_cost_**data**, since it's different than the CHP tech_sizes_for_cost_curve.

src/core/absorption_chiller.jl
```diff
@@ -28,61 +28,217 @@
...
+        if !isnothing(chp_prime_mover)
+            if chp_prime_mover == "combustion_engine"
+                hot_water_or_steam = "steam"
+            else  #if chp_prime mover is blank or is anything but "combustion engine" then assume hot water
+                hot_water_or_steam = "hot_water"
+            end
+        elseif !isnothing(existing_boiler)
+            hot_water_or_steam = existing_boiler.production_type
+        else
+            @error "Invalid argument for `hot_water_or_steam`; must be `hot_water` or `steam`"
```
Bill-Becker, line 182: Let's default to "hot_water" if no other data is given to inform this.

src/core/absorption_chiller.jl
```diff
@@ -28,61 +28,217 @@
...
+    for (k, v) in d
+        if typeof(v) <: AbstractVector{Any}
+            d[k] = convert(Vector{Float64}, v)  # JSON.parsefile makes things Vector{Any}
+        end
+    end
+
+    absorp_chl = AbsorptionChiller(; dictkeys_tosymbols(d)...)
+
+    #check for 0.0 max size, return nothing if so
+    if absorp_chl.max_ton == 0.0
```
Bill-Becker, line 95: Let's not require max_ton, and instead default to a big number, to be consistent with other techs which just require the tech name to be in the inputs.

src/core/absorption_chiller.jl
```diff
@@ -28,61 +28,217 @@
...
+- "installed_cost_per_ton"
+- "om_cost_per_ton"
+- "cop_thermal"
+- "tech_sizes_for_cost_curve"
+
+Unlike CHP, the AbsorptionChiller tech sizes inform a single rate for installed_cost_per_ton that uses max_ton as input;
+there is no piecewise linear cost curve for the AbsorptionChiller technology.
+
+Inputs: 
+max_ton::Float64 -- maximum size of the absorption chiller technology, in tons
```
Bill-Becker, line 146: The defaults for AbsorptionChiller should actually be based on the max cooling load in tons, not the input value of max_ton for AbsorptionChiller. So we'll have to feed that in from the CoolingLoad. I mention in another comment that I think the max_ton value here should actually default to a big number like the other techs.

Bill-Becker (changes requested):
@zolanaj A couple of requests from the review comments. The main two are:

1. The tonnage/size from which to get defaults should be from the max cooling load ton, not the input max_ton for AbsorptionChiller.
2. The max_ton should default to a big number, so just the AbsorptionChiller tech input should be required to model it. I think this is more consistent with the other techs.

I also changed the name of `hot_water_or_steam` to be more explicit: `thermal_consumption_hot_water_or_steam` in a commit that I pushed.

src/core/absorption_chiller.jl
```diff
@@ -28,61 +28,217 @@
 # OF THE POSSIBILITY OF SUCH DAMAGE.
 # *********************************************************************************
 
+heat_transfer_mediums = ["steam", "hot_water"]
+chp_prime_movers = ["recip_engine", "micro_turbine", "combustion_turbine", "fuel_cell"]
```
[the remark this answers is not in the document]
zolanaj: In the commit 298eac373f9feec60c770487aa4089131721aece this is addressed; I also made a const for hot_water_or_steam since it's used by CHP too.  When the CHP work is merged, I'll update that side as well (I'm trying to avoid merge conflicts for now since CHP development is happening in parallel)

src/core/absorption_chiller.jl
zolanaj: It's the latter; I implemented the change in e2da3a96bf9a619170852a07942f6087bcc215df

src/core/absorption_chiller.jl
zolanaj: Makes sense to me!  Implemented in cd21cee4bbae3f40a6b1dbe60ac65d79240ea9f5

src/core/absorption_chiller.jl
zolanaj: I've addressed this in 298eac373f9feec60c770487aa4089131721aece and left a TODO to employ the new const BIG_NUMBER elsewhere, since it's used for a few technologies right now.

src/core/absorption_chiller.jl
zolanaj: Thanks for catching this!  The bulk of this is covered in fe94dee33392ed199b4a97dd1c5009f5dfefbfb4 and other adjacent updates are in cb45aef5411d96cc4d3a97849319176180972e81 and 5faeaed52d3a7e9532277aa96481b34b90f48768.  I updated the relevant tests as well in a966e817b2c61ad25eb620a3ff122b7b93890fdf but will wait on them passing before I send back to you for another review.

zolanaj:
@Bill-Becker tests are now passing and this PR is ready for your next round of review.  Note that I have a couple of TODOs regarding the const values that were added and need to be referenced in other technologies that are developed in other branches right now.

Bill-Becker:
@zolanaj I made a small bug fix for `combustion_turbine` naming (was called `combustion_engine`), updated the default data values for O&M for steam (this was erroneously copied in the **API** from the hot_water values, so it was wrong in the API to begin with and not consistent with the documentation in the User Manual), and then updated the test for "Thermal Energy Storage + Absorption Chiller" to check for consistency between `get_absorption_chiller_defaults` and inputs->`Scenario` struct. I'm good with this as is, so I'll approve, but let me know if you don't agree with any of the changes I made.

Bill-Becker (approved):

Bill-Becker:
I also just changed the base branch into which this branch will merge to be `develop` instead of `master`.

zolanaj:
@Bill-Becker thanks for these updates - and bravo on finding the defaults bug from documentation to API!

I'm good with all these updates, and will merge now.

PR #152, opened 2022-11-22
EV

Bill-Becker:
Include a number of `ElectricVehicle`s that have most attributes of `ElectricStorage` with additional attributes such as:
- Availability profile on-site vs off-site
- State of charge when the EV is first timestep it is on-site
- Minimum required state of charge of the EV the timestep before it leaves off-site

Define the `EVSupplyEquipment` (EVSE is a common industry term to describe the charging station and related controls) which includes the electric capacity (kW) and cost of discrete options to represent different classes/levels of charging stations, such as level 1, 2, and 3. The ability for the EV to discharge back to the site load (V2G) is also defined here.


src/core/energy_storage/electric_vehicle.jl
```diff
@@ -0,0 +1,342 @@
...
+- Add soc_used_off_site options to be Scalar, 2x (weekday/weekend), or 7x (each day of week) Values of 0.0 to (1.0 - min_soc)
+
+
+"""
+Base.@kwdef mutable struct ElectricVehicle
+    energy_capacity_kwh::Float64 = NaN
+    max_c_rate::Float64 = 1.0
+    ev_on_site_start_end::Array{Array{Int64, 1}, 1} = [[0,0],[0,0],[0,0],[0,0]]  # This should **maybe** be an array of arrays (meant to be array of tuples, could convert)
+    ev_on_site_series::Union{Array{Int64, 1}, Nothing} = nothing
+    soc_used_off_site::Array{Array{Float64, 1}, 1} = [[0.2,0.8],[0.2,0.8],[0.2,0.8],[0.2,0.8]]
```
Bill-Becker, line 62: @rathod-b what do the two values in the inner arrays mean? I was thinking there's just one "SOC used" value for each quarter. The ev_on_site_start_end have two values because you need the start and end timestep.

rathod-b: The pair of values in each array can represent the typical incoming EV SOC and desired outbound EV SOC for each quarter. I was thinking that the outbound SOC can be different based on quarters, allowing us to model higher desired SOCs in the summer and winter due to HVAC use and comparatively lower desired SOCs in Fall/Spring. Hopefully this makes sense. Maybe I should rename this fieldname to `ev_arrive_depart_soc`?

Bill-Becker: Ok, I thought that might be the case. We currently we also have `leaving_next_time_step_soc_min` and `back_on_site_time_step_soc_drained`  which are placeholders for custom hourly/15-minute data that would be a separate way to enter this data (along with `ev_on_site_series`, but that might be redundant if we have the first two series data). I think you already know this, so no need to respond.

PR #153, opened 2022-11-30
MPC multinode

NLaws:
a basic start (no generators, no outages)

### Added
- multi-node MPC modeling capability
- more MPC outputs (e.g. Costs, ElectricStorage.to_load_series_kw)
### Fixed
- PV results for all multi-node scenarios
- MPC objective definition w/o ElectricStorage


Bill-Becker:
@NLaws Thanks for your industry/external contribution! I hope things are going well at the new gig. I will definitely get to this before the new year, but I'm not sure how much before. I'd like to learn-up on multi-node anyway, and I'll soon be (finally) working with MPC on a project too.

Bill-Becker (approved):
@NLaws This looks good to me as a start for MPC multi-node! Note that I removed the seemingly redundant `lower_bounds()` function in `reopt_multinode.jl` because as you noted and omitted in `mpc/model_multinode.jl` we shouldn't need that with the `lower_bound=0` argument in the @variable creation.

PR #162, opened 2022-12-23
ThermalStorage Update to Water Property with CoolProp and Round Size

Bill-Becker:
This PR:

1. Updates the water properties for converting the Hot/Cold ThermalStorage size in kWh thermal to gallons of storage to use CoolProp instead of an assumed static value. The CoolProp water property calls use the average water temperature input (or default) and an assumed water pressure (atmospheric) - liquid water properties are much stronger function of temperature than pressure.
2. Rounds the `ThermalStorage.size_gal` to 0 digits instead of many.
3. Add standard function for converting temperature in degrees Fahrenheit to Kelvin.

### DIFF
--- a/src/core/energy_storage/thermal_storage.jl
+++ b/src/core/energy_storage/thermal_storage.jl
@@ -145,7 +145,6 @@ struct ThermalStorage <: AbstractThermalStorage
     function ThermalStorage(s::AbstractThermalStorageDefaults, f::Financial, time_steps_per_hour::Int)
          
         kwh_per_gal = get_kwh_per_gal(s.hot_water_temp_degF, s.cool_water_temp_degF)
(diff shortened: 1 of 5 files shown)


zolanaj (approved):
Thank you @Bill-Becker, looks great!  I ran `get_kwh_per_gal`  for water at 20 degrees C with a +/- 20 degree delta (cold side 48, hot side 88) and got a match to 4 significant digits versus using the old method with hard-coded values for 20 degrees C water and a 40-degree delta so I think we're good.

I made one small addition to the docstrings of `get_kwh_per_gal` to reference the CoolProp API so that there's context for the call to the arguments in `PropsSI`.  This can be merged (feel free to revert to change as well if you prefer).

PR #173, opened 2023-01-24
Update descriptions and fix missing Bau result

adfarth:
Update descriptions of inputs and outputs

Add missing annual_emissions_tonnes_SO2 to BAU results


hdunham:
I think there are some more inputs that are calculated in the structs and the user generally shouldn't provide, e.g. outage_time_steps and scenarios in electric_utility.jl

adfarth:
> I think there are some more inputs that are calculated in the structs and the user generally shouldn't provide, e.g. outage_time_steps and scenarios in electric_utility.jl

@hdunham Good call. I removed these and one or two others in https://github.com/NREL/REopt.jl/pull/173/commits/8c5602d9bb56aec31d6dc300e4f706f6799a4ae3. I also tagged you for review if you don't mind!

src/core/absorption_chiller.jl
```diff
@@ -35,15 +35,15 @@
...
-    min_ton::Float64 = 0.0,
-    max_ton::Float64 = BIG_NUMBER,
-    cop_thermal::Union{Float64, Nothing} = nothing,
-    cop_electric::Float64 = 14.1,
-    om_cost_per_ton::Union{Float64, Nothing} = nothing,
-    macrs_option_years::Float64 = 0,
-    macrs_bonus_fraction::Float64 = 0
+    installed_cost_per_ton::Union{Float64, Nothing} = nothing # Thermal power-based cost of absorption chiller (3.5 to 1 ton to kwt)
+    om_cost_per_ton::Union{Float64, Nothing} = nothing # Yearly fixed O&M cost on a thermal power (ton) basis
+    om_cost_per_kw::Float64 = NaN,
```
hdunham, line 40: why do we have both per ton and per kw? looks like the per kw one just gets overwritten. in that case it shouldn't be a user input right?

adfarth: Agreed, it just gets overwritten based on the `om_cost_per_ton` input. I removed this from the user inputs in https://github.com/NREL/REopt.jl/pull/173/commits/8c36dbc025d18cb6604de27c06135fd7f1e10d63

hdunham (approved):

PR #181, opened 2023-02-02
Wind acres

adfarth:
### Added 
- Constraint on wind sizing based on Site.land_acres and associated test and warnings 
- New Wind input acres_per_kw, defaults to 0.03

### Fixed
- Naming of outputs in markdown text and missing outputs in markdown text


src/core/electric_load.jl
```diff
@@ -109,7 +109,7 @@ mutable struct ElectricLoad  # mutable to adjust (critical_)loads_kw based off o
         off_grid_flag::Bool = false,
         loads_kw::Array{<:Real,1} = Real[],
         path_to_csv::String = "",
-        year::Int = 2020, # used in ElectricTariff to align rate schedule with weekdays/weekends
+        year::Int = 2020, # used in ElectricTariff to align rate schedule with weekdays/weekends # TODO: why don't we default this to 2017 to avoid warning each time?
```
adfarth, line 112: @Bill-Becker Do you happen to know why we default the ElectricLoad year to 2020, but then always change it to 2017 and warn the user? Why not just default to 2017?

test/test_with_xpress.jl
```diff
@@ -661,6 +662,18 @@ end
 
     TODO: will these discrepancies be addressed once NMIL binaries are added?
     =#
+
+    m = Model(optimizer_with_attributes(Xpress.Optimizer, "OUTPUTLOG" => 0))
+    d["Site"]["land_acres"] = 60 # = 2 MW (with 0.03 acres/kW)
+    results = run_reopt(m, d)
+    @test results["Wind"]["size_kw"] == 2000.0 # Wind should be constrained by land_acres
```
Bill-Becker, line 669: I assume you discussed this with some others on the team to implement the max sizing based on the 1.5MW threshold, but just chiming in late here. This would really be two 1000 kW wind turbines or some combo of a larger 1.5MW and a smaller 500 kW turbine, right? Would both of those combos not require the sizing constraint because there's only a max of 1x 1.5MW turbines? I'm wondering if we want to somehow increase the max size based on how many (N) 1.5 MW turbines can fit in the land, where N >= 2 and some integer. So in this case, we should be able to go up to 3000 kW before we get into 2x 1.5MW unit territory before we'd be violating the "turbines greater than 1.5MW need X amount of land". So the simple implementation could be "round up to the nearest N 1.5MW turbines, and max size = N * 1.5MW". I realize this is a little more aggressive, and possibly starting to violate the sizing limitations if the result is closer to the limit, but I like the idea of identifying more first and then deciding to back off. YOU CAN TOTALLY SHUT MY IDEA DOWN BY SAYING THIS METHODOLOGY WAS ALREADY AGREED ON!!! :-)

src/core/electric_load.jl
adfarth: @Bill-Becker I ended up implementing a conditional default ElectricLoad "year" value based on whether or not the user is providing their own data or using a CRB. I think it makes sense to assume by default that user-provided data would be for the latest year available (2022). What do you think of this change? https://github.com/NREL/REopt.jl/pull/181/commits/a36195027e7b5a31978e19b00c357644e2f50e3d

test/test_with_xpress.jl
adfarth: @Bill-Becker I've just mirrored what we currently do in the v2 API here, and what we have documented in the User Manual. I like your idea, and can try to see if anyone currently on the team was in on the original decision and might have thoughts.

Bill-Becker: Ooh, I didn't realize we did this same thing in V2. I thought this was a new capability for some reason. I'm in favor of mirroring!

Bill-Becker (approved):

PR #183, opened 2023-02-02
allow for creation of emissions_profile endpoint in API

hdunham:


Bill-Becker (approved):
@hdunham I did a quick test and it looks good to me. I agree we should have a docstring for any exported functions, so I'd suggest we make one before merging in, even if it's very simplified/high-level.

Bill-Becker:
@hdunham do you know if this is something Indu is waiting on for beta UI?

PR #186, opened 2023-02-10
GHP fixes and for adding to API V3

Bill-Becker:
The changes made in this PR were found to be required as I was [adding GHP to the (future) v3 of the REopt_API](https://github.com/NREL/REopt.jl/pull/186) and testing that. I uncovered some bugs for GHP in REopt.jl, as well as some things needed for v3 API integration. My local Julia version is 1.8, so I updated the REopt environment with that version, added to the compat section, and updated the CI workflow. I also added two dependencies for data loading from .csv. While I *could* have found a way to use DelimitedFiles.readdlm with some custom parsing, I've encountered the need to load in data from various sources in the past and I think things will be easier and cleaner with CSV and DataFrames, and we may start to do a lot more complex data loading for things like FlexLoads in the future. Other than those two big REopt.jl package changes, the others are GHP-specific.

- Update Julia to `v1.8`
- Add `CSV` and `DataFrames` packages to `REopt` dependencies
- In `Scenario`, change `heating_thermal_load_reduction_with_ghp_kw` to `space_heating_thermal_load_reduction_with_ghp_kw` to be more explicit
- Add and modify the `GHP` results to align with the existing/expected results from the v2 REopt_API


test/test_with_xpress.jl
```diff
@@ -1194,7 +1194,13 @@ end
...
     ghp_option_chosen = results["GHP"]["ghp_option_chosen"]
     @test ghp_option_chosen == 2
-    
+
+    # Test GHP heating and cooling load reduced
+    hot_load_reduced_mmbtu = sum(results["GHP"]["space_heating_thermal_load_reduction_with_ghp_mmbtu_per_hour"])
+    cold_load_reduced_tonhour = sum(results["GHP"]["cooling_thermal_load_reduction_with_ghp_ton"])
+    @test hot_load_reduced_mmbtu ≈ 1440.00 atol=0.1
+    @test cold_load_reduced_tonhour ≈ 761382.78 atol=0.1
+
```
zolanaj, line 1203: I added these results tests for the additions to GHP results.

zolanaj (approved):
Looks good, thanks @Bill-Becker!  I added a couple small tests - feel free to merge if you agree with the add-ons (assuming tests pass).

PR #190, opened 2023-02-25
Outage bug fix

rathod-b:
This PR is to fix three issues in REopt.jl
1. Add a stacktrace indicator in response JSON
2. Correct the year one coincident peak cost calculation to be before-tax instead of after-tax
3. Allow outage simulator to be run at sub-hourly time interval


Bill-Becker (approved):
I see this is still in draft form, but if it's ready for review, then I approve! Feel free to add this PR reference to the list of PRs for the current open `master` <- `develop` PR.

rathod-b:
> I see this is still in draft form, but if it's ready for review, then I approve! Feel free to add this PR reference to the list of PRs for the current open `master` <- `develop` PR.

Hi Bill, before making this PR ready to merge I wanted to get some input on the data type of `has_stacktrace.` Initially I had it as a boolean but the tests were failing when they should have been passing. So I switched it to a String to get the tests to pass. However, String type for this field means downstream in the API, it will have to be implemented as a TextField instead of BoolField. Should I keep `has_stacktrace` as type String or try to get it to work as type Bool?

Bill-Becker:
@rathod-b seems like we should try to get it to be boolean to be consistent with the rest of true/false fields, but I'm realizing maybe we don't have any boolean results, only inputs. Either way, seems like there needs to be a type conversion on the Julia side.

rathod-b:
`has_stacktrace` is now of type bool and tests are also passing.

Bill-Becker:
> `has_stacktrace` is now of type bool and tests are also passing.

Great, what was the issue? I don't see any added type conversion or anything.

rathod-b:
> > `has_stacktrace` is now of type bool and tests are also passing.
> 
> Great, what was the issue? I don't see any added type conversion or anything.

It was human error on my part. I was using single '=' instead of 2 '==' signs.

PR #195, opened 2023-03-06
debug steamturbine and require input for Boiler

rathod-b:
- Require fuel_cost_per_mmbtu as an input for Boiler tech
- Correct how steam turbine defaults are determined


rathod-b:
I just realized that this PR was created back in March when we were last working on adding these techs into REopt_API. It should have some relevant changes.

Bill-Becker (approved):

PR #201, opened 2023-03-21
Add CHP to stochastic/multi- outages, some CHP tweaks for UI V3

Bill-Becker:
- Add `CHP` `FuelUsed` and `FuelCost` modeling/tracking for stochastic/multi-outages
- Add `CHP` outputs for stochastic/multi-outages
- Add conditional logic to make `CHP.min_allowable_kw` 25% of `max_kw` if there is a conflicting relationship 
- Iterate on calculating `CHP` heuristic size based on average heating load which is also used to set `max_kw` if not given: once `size_class` is determined, recalculate using the efficiency numbers for that `size_class`.
- Fix non-handling of cost-curve/segmented techs in stochastic outages
- Fix issues with `simulated_load.jl` monthly heating energy input to return the heating load profile

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ Classify the change according to the following categories:
 - Iterate on calculating `CHP` heuristic size based on average heating load which is also used to set `max_kw` if not given: once `size_class` is determined, recalculate using the efficiency numbers for that `size_class`.
 ### Fixed
 - Fix non-handling of cost-curve/segmented techs in stochastic outages
+- Fix issues with `simulated_load.jl` monthly heating energy input to return the heating load profile
 
 ## v0.28.1
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,10 +24,17 @@ Classify the change according to the following categories:
     ### Removed
 
 ## Develop 3/22/23
+### Added
+- Add `CHP` `FuelUsed` and `FuelCost` modeling/tracking for stochastic/multi-outages
+- Add `CHP` outputs for stochastic/multi-outages
 ### Changed
 - Made outages output names not dynamic to allow integration into API
 - Add missing units to outages results field names: **unserved_load_series_kw**, **unserved_load_per_outage_kwh**, **generator_fuel_used_per_outage_gal**
 - Default `Financial` field **microgrid_upgrade_cost_fraction** to 0
+- Add conditional logic to make `CHP.min_allowable_kw` 25% of `max_kw` if there is a conflicting relationship 
+- Iterate on calculating `CHP` heuristic size based on average heating load which is also used to set `max_kw` if not given: once `size_class` is determined, recalculate using the efficiency numbers for that `size_class`.
+### Fixed
+- Fix non-handling of cost-curve/segmented techs in stochastic outages
 
 ## v0.28.1
 ### Added
--- a/src/constraints/outage_constraints.jl
+++ b/src/constraints/outage_constraints.jl
@@ -175,10 +175,10 @@ function add_MG_CHP_fuel_burn_constraints(m, p; _n="")
     if fuel_burn_intercept > 1.0E-7
         #Constraint (1c1): Total Fuel burn for CHP **with** y-intercept fuel burn and supplementary firing
         @constraint(m, MGCHPFuelBurnCon[t in p.techs.chp, s in p.s.electric_utility.scenarios, tz in p.s.electric_utility.outage_start_time_steps],
-            m[Symbol("dvMGFuelUsed"*_n)][t,s,tz]  == p.hours_per_time_step * 
+            m[Symbol("dvMGFuelUsed"*_n)][t,s,tz]  == p.hours_per_time_step * (
                 m[Symbol("dvMGCHPFuelBurnYIntercept"*_n)][s,tz] +
                 sum(p.production_factor[t,tz+ts-1] * fuel_burn_slope * m[Symbol("dvMGRatedProduction"*_n)][t,s,tz,ts]
-                    for ts in 1:p.s.electric_utility.outage_durations[s])
+                    for ts in 1:p.s.electric_utility.outage_durations[s]))
         )
 
         #Constraint (1d): Y-intercept fuel burn for CHP across the scenario outage time steps
--- a/src/constraints/outage_constraints.jl
+++ b/src/constraints/outage_constraints.jl
@@ -161,7 +161,6 @@ function add_MG_Gen_fuel_burn_constraints(m,p)
         sum( m[:dvMGGenMaxFuelCost][s] * p.s.electric_utility.outage_probabilities[s] for s in p.s.electric_utility.scenarios )
     )
 
-    m[:ExpectedMGFuelUsed] += ExpectedMGGenFuelUsed
     m[:ExpectedMGFuelCost] += ExpectedMGGenFuelCost
 end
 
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -24,10 +24,18 @@ Classify the change according to the following categories:
...
 
 ## Develop 3/22/23
+### Added
+- Add `CHP` `FuelUsed` and `FuelCost` modeling/tracking for stochastic/multi-outages
+- Add `CHP` outputs for stochastic/multi-outages
 ### Changed
 - Made outages output names not dynamic to allow integration into API
 - Add missing units to outages results field names: **unserved_load_series_kw**, **unserved_load_per_outage_kwh**, **generator_fuel_used_per_outage_gal**
 - Default `Financial` field **microgrid_upgrade_cost_fraction** to 0
+- Add conditional logic to make `CHP.min_allowable_kw` 25% of `max_kw` if there is a conflicting relationship
```
hdunham, line 34: does this need to be added to inputs_with_defaults_from_chp in API?

src/constraints/outage_constraints.jl
```diff
@@ -56,11 +56,22 @@ function add_outage_cost_constraints(m,p)
...
+            m[:binMGTechUsed][t] => {m[:dvMGTechUpgradeCost][t] >= p.s.financial.microgrid_upgrade_cost_fraction * p.third_party_factor *
+                                    p.cap_cost_slope[t] * m[:dvMGsize][t]}
+        )
+    end
 
-    @constraint(m, [t in p.techs.elec],
-        m[:binMGTechUsed][t] => {m[:dvMGTechUpgradeCost][t] >= p.s.financial.microgrid_upgrade_cost_fraction * p.third_party_factor *
-		                         p.cap_cost_slope[t] * m[:dvMGsize][t]}
-    )
+    if !isempty(p.techs.segmented)
```
hdunham, line 67: can there be non-elec segmented techs?

src/constraints/outage_constraints.jl
```diff
@@ -143,14 +154,73 @@ function add_MG_fuel_burn_constraints(m,p)
...
+    
+    @constraint(m, [s in p.s.electric_utility.scenarios, tz in p.s.electric_utility.outage_start_time_steps],
+        m[:dvMGCHPMaxFuelCost][s] >= sum( MGCHPFuelCost[t, s, tz] for t in p.techs.chp )
+    )
+    
+    @expression(m, ExpectedMGCHPFuelCost,
+        sum( m[:dvMGCHPMaxFuelCost][s] * p.s.electric_utility.outage_probabilities[s] for s in p.s.electric_utility.scenarios )
+    )
+
+    m[:ExpectedMGFuelUsed] += ExpectedMGCHPFuelUsed
```
hdunham, line 221: How can gen and chp fuel be added together? Aren't they different units?

src/constraints/outage_constraints.jl
Bill-Becker: Nope, I just checked and AbsorptionChiller and Boiler, the only non-elec techs which have a capital cost, do not have any attribute that can make them segmented (cost curve or incentive inputs).

hdunham: ok, lets just create an issue to make this robust to that possibility in case it happens later

src/constraints/outage_constraints.jl
Bill-Becker: Shoot yes, that's not right. It looks like we're not really using ExpectedMGFuelUsed - think we need it?

src/constraints/outage_constraints.jl
```diff
@@ -143,14 +154,73 @@ function add_MG_fuel_burn_constraints(m,p)
...
+    if fuel_burn_intercept > 1.0E-7
+        #Constraint (1c1): Total Fuel burn for CHP **with** y-intercept fuel burn and supplementary firing
+        @constraint(m, MGCHPFuelBurnCon[t in p.techs.chp, s in p.s.electric_utility.scenarios, tz in p.s.electric_utility.outage_start_time_steps],
+            m[Symbol("dvMGFuelUsed"*_n)][t,s,tz]  == p.hours_per_time_step * 
+                m[Symbol("dvMGCHPFuelBurnYIntercept"*_n)][s,tz] +
+                sum(p.production_factor[t,tz+ts-1] * fuel_burn_slope * m[Symbol("dvMGRatedProduction"*_n)][t,s,tz,ts]
+                    for ts in 1:p.s.electric_utility.outage_durations[s])
+        )
+
+        #Constraint (1d): Y-intercept fuel burn for CHP across the scenario outage time steps
```
hdunham, line 185: I'm noticing that this CHP fuel burn intercept usage is different than for generator (in outages and not). The CHP fuel_burn_intercept is multiplied by dvSize to get dvMGCHPFuelBurnYIntercept, which seems correct to me based on how fuel_burn_intercept is calculated. But it's making me think that this is handled incorrectly for generator, probably a result of changing the gen fuel inputs to efficiencies. Gen slope and intercept are calculated the same way as for CHP (plus unit conversion to gal) so that intercept value needs to be multiplied by size, however it is used the way that the old input fuel_intercept_gal_per_hr (which didn't need to be multiplied by size) was used. Kinda confusing, am I making sense?

src/constraints/outage_constraints.jl
```diff
@@ -143,14 +154,73 @@ function add_MG_fuel_burn_constraints(m,p)
...
+    
+    @constraint(m, [s in p.s.electric_utility.scenarios, tz in p.s.electric_utility.outage_start_time_steps],
+        m[:dvMGCHPMaxFuelUsage][s] >= sum( m[:dvMGFuelUsed][t, s, tz] for t in p.techs.chp )
+    )
+    
+    @expression(m, ExpectedMGCHPFuelUsed, 
+        sum( m[:dvMGCHPMaxFuelUsage][s] * p.s.electric_utility.outage_probabilities[s] for s in p.s.electric_utility.scenarios )
+    )
+
+    # fuel cost = kWh * $/kWh
```
hdunham, line 208: Seems weird that CHP dvMGFuelUsed units would be kwh but generator gal. Something to fix when we generalize generator to use other fuels. Made an issue.

src/constraints/outage_constraints.jl
hdunham: Or maybe it's just late and I can't think straight because wouldn't the generator fuel input change have broken tests if this was the case?

src/constraints/outage_constraints.jl
hdunham: No we just need the separate gen and CHP expected mg fuel used

src/constraints/outage_constraints.jl
Bill-Becker: Hmm, yeah I think you're right. I'll take a look.

src/constraints/outage_constraints.jl
Bill-Becker: Yeah, I agree it's inconsistent but I'm also hesitant to keep adding issues for things that aren't necessarily issues right now. They are piling up (50+) and I'm concerned it's just getting to be a big pile of stuff we'll maybe never do.

src/constraints/outage_constraints.jl
Bill-Becker: I removed the sum/total expression.

hdunham (approved):
generator fuel burn intercept to be fixed in another PR

PR #202, opened 2023-03-22
Make outages output names not dynamic, add units to outages results, microgrid_upgrade_cost_fraction default to 0

hdunham:
### Changed
- Made outages output names not dynamic to allow integration into API
- Add missing units to outages results field names: **unserved_load_series_kw**, **unserved_load_per_outage_kwh**, **generator_fuel_used_per_outage_gal**
- Default `Financial` field **microgrid_upgrade_cost_fraction** to 0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,12 @@ Classify the change according to the following categories:
     ### Deprecated
     ### Removed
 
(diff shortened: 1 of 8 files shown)


Bill-Becker (approved):
Not worried about that pesky Mac test run.

PR #209, opened 2023-03-30
Make generator higher heating value an input

hdunham:
### Added
- `Generator` input **fuel_higher_heating_value_kwh_per_gal**, which defaults to the existing constant KWH_PER_GAL_DIESEL


Bill-Becker (approved):
Approved, but do you think we should remove the const KWH_PER_GAL_DIESEL in REopt.jl to avoid confusion, or is that used elsewhere that warrants it still being a global constant?

hdunham:
> Approved, but do you think we should remove the const KWH_PER_GAL_DIESEL in REopt.jl to avoid confusion, or is that used elsewhere that warrants it still being a global constant?

I'm still using it for the default value

PR #210, opened 2023-03-30
Change Replacements Treatment

### LINKED ISSUE #69: Review tax considerations (make replacements depreciable, check for double-counting O&M tax expensing in Third Party Financing calculations)

### LINKED ISSUE #212: Treat replacements as capital costs (apply ITC and MACRS), not as tax deductible costs

adfarth:
- `ElectricStorage` and `Generator` have `replacement_year` inputs. It is assumed that these two technologies can be replaced within the analysis period. 
- Previously, replacement costs were treated as tax deductible. Neither ITC nor MACRS were applied to these costs. 
- Now, we are treating replacement costs are depreciable and are applying the ITC and MACRS to these costs. 
- The following new inputs have been added to `ElectricStorage` and `Generator`:
-- `replace_macrs_option_years`
-- `replace_macrs_bonus_fraction`
-- `replace_total_itc_fraction` 
- For Generators, we assume no ITC or MACRS incentives for the upfront cost or replacements
- For ElectricStorage replacements, we assume that the MACRS bonus depreciation has phased out, that storage qualifies for 5-year depreciation (2025 and onward), and that the ITC remains at 30%. 
- This is all captured in the `effective_cost()` calculation 

TODO: 
- [x] Update CHANGELOG
- [ ] Add tests
- [ ] Update reporting of results (esp. in financial.jl and proforma.jl)
- [ ] Update existing tests once this branch is fully tested 
- [ ] Consider changing input names to "replacement" not just "replace"? 
- [ ] Also update docs?
- [ ] Check if generator replacement costs apply to total capacity or just new

Note! I haven't looked into tax treatment assumptions built into replacement costs with battery degradation modeling.


src/core/generator.jl
```diff
@@ -48,7 +48,7 @@
     can_export_beyond_nem_limit = false,
     can_curtail::Bool = false,
     macrs_option_years::Int = 0,
-    macrs_bonus_fraction::Real = 1.0,
```
adfarth, line 51: I'm not sure why we previously defaulted this to 1. It got set to 0 anyway (in cost_curve.jl) since macrs_option_years = 0.

src/core/utils.jl
```diff
@@ -162,6 +175,63 @@ function effective_cost(;
     return round(cap_cost_slope, digits=4)
 end
 
+function replacement_effective_cost(;
```
adfarth, line 178: I made this a separate function so we can just call this to get replacement costs in the reporting of results.

PR #214, opened 2023-04-06
Create and export easiur_data function for the API to use

hdunham:
Created and exported easiur_data function (which returns health emissions costs and escalations) for the API to be able to call for it's easiur_costs endpoint

### DIFF
--- a/src/core/financial.jl
+++ b/src/core/financial.jl
@@ -471,3 +471,22 @@ Convert Geodetic (lon, lat) to LCP (x, y) in CAMx 148x112 grid
 function g2l(lon::Real, lat::Real; datum::String="NAD83")
     return l2g(lon, lat, inverse=true, datum=datum)
 end
+
+#TODO: add docstring
+function easiur_costs(; latitude, longitude, inflation)
+    response_dict = Dict{String, Any}(
+        "units_costs" => "US dollars per metric ton",
+        "description_costs" => "Health costs of emissions from the grid and on-site fuel burn, as reported by the EASIUR model.",
+        "units_escalation" => "nominal annual fraction",
+        "description_escalation" => "Annual nominal escalation rate of public health costs of emissions.",
+    )
+    grid_costs = easiur_costs(latitude, longitude, "grid")
+    onsite_costs = easiur_costs(latitude, longitude, "grid")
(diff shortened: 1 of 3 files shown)


src/core/financial.jl
```diff
@@ -471,3 +471,22 @@ Convert Geodetic (lon, lat) to LCP (x, y) in CAMx 148x112 grid
 function g2l(lon::Real, lat::Real; datum::String="NAD83")
     return l2g(lon, lat, inverse=true, datum=datum)
 end
+
+#TODO: add docstring
+function easiur_data(; latitude, longitude, inflation)
```
rathod-b, line 476: Can we add data types for these input arguments?

src/core/financial.jl
```diff
@@ -471,3 +471,22 @@ Convert Geodetic (lon, lat) to LCP (x, y) in CAMx 148x112 grid
...
+#TODO: add docstring
+function easiur_data(; latitude::Real, longitude::Real, inflation::Real)
+    response_dict = Dict{String, Any}(
+        "units_costs" => "US dollars per metric ton",
+        "description_costs" => "Health costs of emissions from the grid and on-site fuel burn, as reported by the EASIUR model.",
+        "units_escalation" => "nominal annual fraction",
+        "description_escalation" => "Annual nominal escalation rate of public health costs of emissions.",
+    )
+    grid_costs = easiur_costs(latitude, longitude, "grid")
+    onsite_costs = easiur_costs(latitude, longitude, "grid")
```
rathod-b, line 484: And the third argument here should be "onsite" so `type` argument for `get_EASIUR2005()` can be set to either area or p150?

rathod-b: Everything else looks good! I can approve after resolving above comment. One issue we might need to make is for handling what happens when site coordinates are outside the US. I accidentally ran a test case and forgot the minus sign in `longitude`. The function errored out as following. 

image

hdunham: Addressed both of these. Errors thrown in this function are caught in the API, but I made the error more informative.

hdunham: Updated to add the error to the response dictionary instead

rathod-b (approved):
It works! Maybe we should consider the error handling style/format in the future if we have a discussion on standardizing error messages.

Issue #216, opened 2023-04-12
Add Wind to simulate_outages


zolanaj:
When running the outage simulator's `simulate_outages` method that uses the results dictionary and REoptInputs (and the optional Bool `microgrid_only`) as input, the wind production `wind_ac_hourly` isn't calculated and is instead passed through as an empty list that is later converted to a vector of zeros: https://github.com/NREL/REopt.jl/blob/20fa1047653322aec851bccd879a23519ab30cb8/src/outagesim/outage_simulator.jl#L306

As a result, Wind does not account for resilience measures when running the outage simulator using this method.

PR #220, opened 2023-04-18
Fix wind time series input to outage simulator

zolanaj:
- Fixes a bug in which Wind time series output was excluded from the outage simulator; instead of calculating production, an empty list was passed into `simulate_outages()`.  
- Adds a test of multiple outages that includes wind as a technology.


zolanaj:
Linked issue: https://github.com/NREL/REopt.jl/issues/216

hdunham (approved):
Looks good. I pushed one little efficiency change. Just making sure, you're aware that the test you updated doesn't run simulate_outages() right?

zolanaj:
@hdunham thank you!  Yes, I kept simulate_outages() out of the test suite because the test had a really long runtime whe I did include it and I didn't want to burden the test suite.  I'll try to craft a test case for simulate_outages that includes all the technologies and still is quick (something like a unit test where production factors are 1 in one period and zero o.w.) in the future but will merge this for now.

PR #222, opened 2023-04-21
PVWatts API Fix

Bill-Becker:
Issue:
PVWatts API connection periodically hangs for a long time (15+ minutes) when using REopt.jl which calls the PVWatts API on our Rancher servers.

Fix: 
Add a timeout so that if it does not connect within 10 seconds, it will retry and we've seen it work on the first retry.


hdunham (approved):
Approving. Only comment is I don't think we need the added @info because there's the error message or "PVWatts success" info right below it.

PR #224, opened 2023-04-24
Handle matrix type urdb demand fields

Bill-Becker:
When calling REopt.jl from a python environment using PyJulia and PyCall, some urdb_response fields get converted from a list-of-lists to a matrix type, when REopt.jl expects an array type. This PR adds checks on the type for two urdb_response fields and converts them to an array if needed.


brtietz (approved):
This works well with data structures delivered by PyJulia. Thanks for implementing!

PR #228, opened 2023-05-05
Update LinDistFlow

NLaws:
The LinDistFlow API has significantly changed. This PR updates the methods for integrating REopt with the LinDistFlow model.

Also:
- remove test_with_cplex.jl (b/c I don't think anyone used it except me?)
- mv LinDistFlow test to test_with_xpress (from test_with_cplex)
- ignore test/Manifest.toml


NLaws:
welp, SCIP doesn't work on Windows. I will have to get back to this later.

PR #236, opened 2023-06-29
Hot fix for rates with non standard units

adfarth:


src/core/urdb.jl
```diff
@@ -292,10 +292,12 @@ function parse_urdb_energy_costs(d::Dict, year::Int; time_steps_per_hour=1, bigM
                     end
                     if non_kwh_units
                         rate = rate_average
```
adfarth, line 294: @rathod-b `rate_average` already includes adjustments

rathod-b (approved):
Good catch!

PR #245, opened 2023-08-08
Existing chiller cop api setup

zolanaj:
- Changes the inputs to `get_existing_chiller_cop` from vectors to single numbers (or nothing)
- Exports the function `get_existing_chiller_cop` for consistency with other functions that we call from the API (future todo: see if we need this)


zolanaj:
@Bill-Becker the updated tests from this PR passed locally - just waiting on the remaining tests before merging.

Bill-Becker (approved):

PR #252, opened 2023-08-21
add cooling_defaults to cooling responses in simulated_load

zolanaj:
Intended to address https://github.com/NREL/REopt_API/issues/491


zolanaj:
Closing this as issue https://github.com/NREL/REopt_API/issues/491 will be resolved without adding cooling defaults to the cooling load response in `simulated_load`, and the monthly inputs fix is addressed in new PR https://github.com/NREL/REopt.jl/pull/258.

PR #256, opened 2023-08-28
Add electric heater tech

zolanaj:
This adds an electric heater technology to the code base.  A couple of notes:
- The default costs assume a resistive heater with a COP of one.
- We included all default data in a .json in the data/ folder so it can be queried by the API, hopefully allowing for defaults to be stored in one place and called through http.jl when this is added to the API at a later time.
- The COP field uses `heating_cop` in the place of just `cop` in anticipation of the future addition of an air-sourced heat pump and other hybrid heating/cooling technologies that might use both heating and cooling (and the desire to keep the COP references in a single set).


zolanaj:
closing and will restart using a fresh branch after merging obsolete (pre-reset) commits from develop

PR #266, opened 2023-09-20
Add Outages results, Make Outages load balance == instead of >=

Bill-Becker:
- Add the critical load in the Outages results for all modeled outages
- Add Wind results for outages, just like PV - it was using Wind in the load balance, but not reporting in the results
- Add battery soc series results for all modeled outages
- Make the outage load balance `supply == load` instead of `supply >= load`

Added a test which covers all of those additions. I confirmed that before I did `supply == load`, the supply from Wind to load and battery to load was much greater than the critical load.


Bill-Becker:
@hdunham can you try to run the test that I added locally? It passes locally for me but is failing on GitHub Actions and I can't track down the Error.

Bill-Becker:
> @hdunham can you try to run the test that I added locally? It passes locally for me but is failing on GitHub Actions and I can't track down the Error.

Nevermind - just a variable name change that I had stored locally so it didn't error locally.

Bill-Becker:
@hdunham this is just failing from the Mac tests

src/results/outages.jl
```diff
@@ -12,6 +12,11 @@
 - `pv_to_storage_series_kw` Array of PV power sent to the battery in every outage modeled.
 - `pv_curtailed_series_kw` Array of PV curtailed in every outage modeled.
 - `pv_to_load_series_kw` Array of PV power used to meet load in every outage modeled.
+- `wind_microgrid_size_kw` Optimal microgrid PV capacity. Note that the name `PV` can change based on user provided `PV.name`.
```
hdunham, line 15: Update these descriptions to refer to Wind not PV

src/results/outages.jl
```diff
@@ -67,13 +74,24 @@ function add_outage_results(m, p, d::Dict)
 	r["unserved_load_per_outage_kwh"] = round.(unserved_load_per_outage, digits=2)
 	r["storage_microgrid_upgrade_cost"] = value(m[:dvMGStorageUpgradeCost])
 	r["microgrid_upgrade_capital_cost"] = r["storage_microgrid_upgrade_cost"]
+    _n = ""
```
hdunham, line 77: If multinode doesn't work with this multiple outage modeling, why bother with doing this and then possibly giving the illusion that they are compatible, rather than just doing m[:dvStorageEnergy]?

src/results/outages.jl
```diff
@@ -67,13 +74,24 @@ function add_outage_results(m, p, d::Dict)
...
 	if !isempty(p.s.storage.types.elec) && Bool(round(value(m[:binMGStorageUsed]), digits=0))
 		r["storage_discharge_series_kw"] = value.(m[:dvMGDischargeFromStorage]).data
+        electric_storage_energy_capacity_kwh = round(value(m[Symbol("dvStorageEnergy"*_n)]["ElectricStorage"]), digits=2)
```
hdunham, line 80: More robust to use p.s.storage.types.elec instead of hardcoding "ElectricStorage" key

src/results/outages.jl
```diff
@@ -67,13 +74,24 @@ function add_outage_results(m, p, d::Dict)
...
+        r["soc_series_fraction"] = round.(value.(m[:dvMGStoredEnergy][:,:,1:end]).data ./ electric_storage_energy_capacity_kwh, digits=3)
 	else
-		r["storage_discharge_series_kw"] = []
+		r["storage_discharge_series_kw"] = value.(m[:dvMGDischargeFromStorage]).data
```
hdunham, line 83: Should this be = []? The same line is in the if part above, so if not then move out of the if else?
Also should soc_series_fraction = [] on the next line instead of zeros array? I think that's generally what we do when a tech isn't included right?

src/results/outages.jl
```diff
@@ -67,13 +74,24 @@ function add_outage_results(m, p, d::Dict)
...
+        r["soc_series_fraction"] = round.(value.(m[:dvMGStoredEnergy][:,:,1:end]).data, digits=3)  # Will be zeros
 	end
+    
+    r["critical_loads_per_outage_series_kw"] = Matrix{Array{Float64}}(undef, S, T)
```
hdunham, line 87: I'd suggest doing Array{Float64,3} instead. This would match other Outages outputs like unserved_load_series_kw and the tech dispatch series that are 3D arrays. If you want Matrix{Array} so that unused elements in the 3rd dim aren't included, then the other outputs should be changed to match this one.

test/test_with_xpress.jl
```diff
@@ -498,6 +498,25 @@ end
...
+    input_data = JSON.parsefile("./scenarios/wind_outages.json")
+    s = Scenario(input_data)
+    inputs = REoptInputs(s)
+    m1 = Model(optimizer_with_attributes(Xpress.Optimizer, "MIPRELSTOP" => 0.01, "OUTPUTLOG" => 0))
+    m2 = Model(optimizer_with_attributes(Xpress.Optimizer, "MIPRELSTOP" => 0.01, "OUTPUTLOG" => 0))
+    results = run_reopt([m1,m2], inputs)
+
+    # Check that supply-to-load is equal to critical load during outages, including wind
+    supply_to_load = results["Outages"]["storage_discharge_series_kw"] .+ results["Outages"]["wind_to_load_series_kw"]
+    supply_to_load = [supply_to_load[:,:,i][1] for i in eachindex(supply_to_load)]
```
hdunham, line 511: Each index loops through all dims of supply_to_load, but you're using i in the last dim. Did this not throw an error?
Are you trying to test just the series corresponding t scenario 1, start time 1?

src/results/outages.jl
Bill-Becker: Addressed in c6b049b515b7e07c4bd58fb4b9b31041ee31e1a7

src/results/outages.jl
Bill-Becker: Sure, address in d24c2e2b348ce9d6245913b25a4d95854f9df48c.

src/results/outages.jl
Bill-Becker: Addressed in 9275b0de798fa66d1e77efbc194ca8f6879926f7.

src/results/outages.jl
Bill-Becker: Yep, addressed in 6e3e85a75d1a6f6ce5f6d3df40b10801e20beaed.

test/test_with_xpress.jl
Bill-Becker: There's only one outage start time and duration in the scenario.

src/results/outages.jl
Bill-Becker: Ok, addressed in abdf8b68ea76a975440460a6a7b142d275b436f9.

hdunham (approved):

PR #272, opened 2023-09-26
Centralized ghp, hybrid ghp, ghp residual value

lixiangk1:
@Bill-Becker Just wanted to get your eyes on this as soon as possible since Dan is out Friday. Bhavesh and I are still adding tests. This branch contains all the changes associated with centralized GHP, hybrid GHP, and calculating residual value for GHP. One area we especially wanted you to look at is all the changes around using the `cost_curve` function. Dan was a bit concerned about units, but we had that discussion about how we're really just co-opting that function to deal with incentives and not because we're modeling a GHP cost curve... so just want to make sure all the changes still fit within how you originally intended to use the `cost_curve` function for GHP.


src/core/reopt.jl
```diff
@@ -144,7 +144,7 @@ function run_reopt(ms::AbstractArray{T, 1}, p::REoptInputs) where T <: JuMP.Abst
 		Threads.@threads for i = 1:2
 			rs[i] = run_reopt(inputs[i])
 		end
-		if typeof(rs[1]) <: Dict && typeof(rs[2]) <: Dict
+		if typeof(rs[1]) <: Dict && typeof(rs[2]) <: Dict && rs[1]["status"] != "error" && rs[2]["status"] != "error"
```
Bill-Becker, line 147: Does this fix the TODO comment below?

Bill-Becker (commented):
All looks good - but I'll take another look once the tests are added. I assume you'll take the same approach to using a stored ghpghx_response.json for a hybrid run to avoid running GhpGhx.jl in the test because of the "conditional loading" of GhpGhx.jl package into REopt.jl, like I did (I couldn't find a way to run that on GitHub workflows due to GhpGhx.jl not being registered).

src/core/reopt.jl
rathod-b: Depends on what the anticipated functionality was in the TODO. Right now, if rs[1] is optimal but rs[2] errors out, then both rs[1] and rs[2] are returned where rs[2] will contain the true error. we can add a @warn before returning `rs` to alert the user that they will receive a vector of Dict elements.

rathod-b: We'd have to handle infeasible models on line 529. instead of returning `m` we might want to return a simple dictionary with termination status? maybe invoke the error handler here.

Bill-Becker: Not trying to add work here, just wondering what the change was. Feel free to proceed as you'd like even if it doesn't address the TODO.

rathod-b: We might have resolved the TODO to some extent... we throw an error if models are infeasible, which returns a dictionary that contains model infeasibility warning as well as an error text.

rathod-b:
> All looks good - but I'll take another look once the tests are added. I assume you'll take the same approach to using a stored ghpghx_response.json for a hybrid run to avoid running GhpGhx.jl in the test because of the "conditional loading" of GhpGhx.jl package into REopt.jl, like I did (I couldn't find a way to run that on GitHub workflows due to GhpGhx.jl not being registered).

Test associated with Hybrid GHP has been added and is passing.

lixiangk1:
@Bill-Becker The central plant test is in there too now, and I've put in the workaround for incentives we discussed yesterday. This branch should be ready to merge unless you find issues.

src/core/reopt_inputs.jl
```diff
@@ -974,26 +981,43 @@ function setup_ghp_inputs(s::AbstractScenario, time_steps, time_steps_without_gr
...
-                end
+            if option.heat_pump_configuration == "WSHP"
+                fixed_cost, variable_cost = get_ghp_installed_cost(option, s.financial, option.heatpump_capacity_ton)
+                ghp_installed_cost[i] = fixed_cost + variable_cost
+
+            elseif option.heat_pump_configuration == "WWHP"
+                temp = option.installed_cost_per_kw
+                option.installed_cost_per_kw = option.wwhp_heating_pump_installed_cost_curve
+                fixed_cost_heating, variable_cost_heating = get_ghp_installed_cost(option, s.financial, option.wwhp_heating_pump_capacity_ton)
+                ghp_installed_cost_heating = 0.5 * fixed_cost_heating + variable_cost_heating
```
Bill-Becker, line 999: What is the 0.5 from?

Bill-Becker (approved):
Approving but please answer the couple of questions and/or consider replacing hard-coded values with existing parameters/functions to understand how they were derived.

src/core/reopt_inputs.jl
lixiangk1: This is just to avoid double counting the "fixed cost", since GHX/aux unit costs are the same in the cost curve for both the heating heat pump and cooling heat pump. The current workaround to make incentives work mean these values aren't used since there's no y-intercept.

test/test_with_xpress.jl
```diff
@@ -1244,6 +1245,68 @@ end
...
+    cooling_hp_cost = input_data_wwhp["GHP"]["installed_cost_wwhp_cooling_pump_per_ton"] * 
+                        input_data_wwhp["GHP"]["heatpump_capacity_sizing_factor_on_peak_load"] *
+                        results_wwhp["GHP"]["ghpghx_chosen_outputs"]["peak_cooling_heatpump_thermal_ton"]
+
+    ghx_cost = input_data_wwhp["GHP"]["installed_cost_ghx_per_ft"] * 
+                results_wwhp["GHP"]["ghpghx_chosen_outputs"]["number_of_boreholes"] * 
+                results_wwhp["GHP"]["ghpghx_chosen_outputs"]["length_boreholes_ft"]
+
+    # CAPEX reduction factor for 30% ITC, 5-year MACRS, assuming 26% tax rate and 8.3% discount
+    capex_reduction_factor = 0.455005797
```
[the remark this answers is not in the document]
lixiangk1: This was calculated in a spreadsheet Dan made to cross check REopt calculations. We can make it more explicit how this value was derived in the next iteration of GHP fixes.

PR #281, opened 2023-10-02
Add GHP to proforma metrics, add thermal BAU outputs

rathod-b:


test/test_with_xpress.jl
```diff
@@ -1245,11 +1245,44 @@ end
...
+    
+    calculated_om_costs = inputs.s.ghp_option_list[1].building_sqft*
+    inputs.s.ghp_option_list[1].om_cost_per_sqft_year * inputs.third_party_factor * inputs.pwf_om
+
+    @test results["Financial"]["lifecycle_om_costs_before_tax"] = calculated_om_costs atol=0.1
+
+    calc_om_cost_after_tax = calculated_om_costs*(1-inputs.s.financial.owner_tax_rate_fraction)
+    @test results["Financial"]["lifecycle_om_costs_after_tax"] = calc_om_cost_after_tax atol=0.1
+
+    @test r["lifecycle_capital_costs_plus_om_after_tax"] = results["Financial"]["initial_capital_costs"]*0.7 + calc_om_cost_after_tax atol=5
```
rathod-b, line 1280: For some reason, this and subsequent tests arent adding up. There is a difference of ~54,000 between the two numbers. Any idea what could be causing this?

Bill-Becker (approved):

src/results/proforma.jl
```diff
@@ -359,6 +364,72 @@ function update_metrics(m::Metrics, p::REoptInputs, tech::AbstractTech, tech_nam
     nothing
 end
 
+function update_ghp_metrics(m::REopt.Metrics, p::REoptInputs, tech::REopt.AbstractTech, tech_name::String, results::Dict, third_party::Bool)
```
Bill-Becker, line 367: @rathod-b I updated this to handle Central GHP "WWHP" with the two different heat pump sizes. I just made `new_kw` == average of two sizes as a workaround even though they should be handled separately. Punt on that!

rathod-b: Makes sense!

PR #283, opened 2023-10-04
Update CHP defaults to 2023 values

zolanaj:
Updates the CHP Defaults to the 2023 values (and revised tests to match), with the following caveats: 
- The microturbine does not have a hot-water efficiency value by size class; they have steam, and steam with supplementary firing - nothing else.  Hot water efficiencies are left alone by size class right now, but perhaps it's best to make them zero instead?
- No updates have been made for the "fuel_cell" prime mover as we don't have revised data for that (yet).


Bill-Becker:
@zolanaj let's discuss the microturbine hot water vs steam efficiency by size class. Ping me when you're available.

zolanaj:
@Bill-Becker I think this is ready for number checking; please see the Teams folder for the revised worksheets.  I isolated all the inflation adjustments (including fuel cells) into commit https://github.com/NREL/REopt.jl/pull/283/commits/93efebdb0f50bc83a2d5e51489f3c4c2410ba381 so it should be easy to revert.

Do you want me to work on the project or manifest .toml files to get xpresstests building properly, or is there a feature branch I can merge?

Bill-Becker:
@zolanaj All numbers cross-checked, and the one interesting one that I agree with is the change from 50% to 25% for minimum loading for recips, as it says in the PDF.

PR #289, opened 2023-10-10
Update chp defaults 2023

Bill-Becker:
Updates the CHP Defaults to the 2023 values (and revised tests to match), with the following caveats: 
- The microturbine does not have a hot-water efficiency value by size class; they have steam, and steam with supplementary firing - nothing else.  Hot water efficiencies are left alone by size class right now, but perhaps it's best to make them zero instead?
- No updates have been made for the "fuel_cell" prime mover as we don't have revised data for that (yet).


Bill-Becker:
@zolanaj looks like all the tests passed once we got the dependency thing fixed. Oscar actually took initiative to propose an alternative solution in this PR https://github.com/NREL/REopt.jl/pull/292, where we don't actually have the Manifest.toml file in the repo, and we can fix it using the [compat] section of the test/Project.toml. Sounds like that might be best-practice to not have the Manifest.toml in the package.

What else do we need to do before merging this?

zolanaj (approved):
@Bill-Becker with the changelog updated, I think we can merge to develop (but please let me know if you want me to put more details in the changelog like there were for the other default values update).  After we test the web tool (I'll let Erika know about the change in number of size classes once we merge), and update the User Manual, we should be ready to move to master.

zolanaj:
> @zolanaj looks like all the tests passed once we got the dependency thing fixed. Oscar actually took initiative to propose an alternative solution in this PR #292, where we don't actually have the Manifest.toml file in the repo, and we can fix it using the [compat] section of the test/Project.toml. Sounds like that might be best-practice to not have the Manifest.toml in the package.

I think Oscar's fix is safe to merge assuming tests pass - I agree it's worth following best practices (and only managing test/Project.toml should be lower maintenance).

PR #291, opened 2023-10-11
Update CHP defaults, and update Julia dependencies

Bill-Becker:
PRs:
- https://github.com/NREL/REopt.jl/pull/290
- https://github.com/NREL/REopt.jl/pull/289


Bill-Becker:
@zolanaj I'm seeing an issue with Oscar's strategy with the GhpGhx package which is only conditionally loaded with the Requires package. I vaguely remember something about this when adding GHP. I'm going to use my strategy for now, since it's working, and I will work toward removing the Manifest.toml with the [compat] section of the test/Project.toml in the near future.

zolanaj (approved):

PR #294, opened 2023-10-12
Electric heater

zolanaj:
This adds an electric heater technology to the code base. A couple of notes:

- The default costs assume a resistive heater with a COP of one.
- We included all default data in a .json in the data/ folder so it can be queried by the API, hopefully allowing for defaults to be stored in one place and called through http.jl when this is added to the API at a later time.


zolanaj:
@Bill-Becker updating to develop addressed the bug I ran into, so this is ready to review!

Bill-Becker (approved):
My "no comments" approval is not an indication that I didn't review closely! I had a few thoughts/questions as I went through it, but those ended up being more preference/nit comments. I like the straightforward/understandable approach you took for applying defaults. I also understand that An did some testing - maybe beyond the test that you added - so that is comforting as well.

Bill-Becker:
Something that came to mind: for a lot of applications of this, I anticipate clients will want to see the results if they electrify their **specific** heating load/process (cost and emissions impact). So it would be great if we had the ability to "force in" both size **and dispatch** of the `ElectricHeater` to serve one or more of the heating loads (e.g. even if it is not cost optimal to operate in a given hour). We could incorporate this with the disaggregation of heating loads and adding a process heating load effort.

zolanaj:
> Something that came to mind: for a lot of applications of this, I anticipate clients will want to see the results if they electrify their **specific** heating load/process (cost and emissions impact). So it would be great if we had the ability to "force in" both size **and dispatch** of the `ElectricHeater` to serve one or more of the heating loads (e.g. even if it is not cost optimal to operate in a given hour). We could incorporate this with the disaggregation of heating loads and adding a process heating load effort.

Thanks! I added issue https://github.com/NREL/REopt.jl/issues/335, but kept it more general than just the `ElectricHeater` as the forced-dispatch feature may be useful for other techs too (I'm thinking of ASHP in particular).

PR #296, opened 2023-10-13
Add adjustments to defaults for CHP prime generator

zolanaj:
- Adds new attribute "is_electric_only" to CHP technology (if true, the tech is a prime generator)
- Adjusts capital and O&M costs for prime generators, and reduces ITC credit to 0 for combustion turbines and reciprocating engines
- New tests added for O&M, ITC default adjustments


zolanaj:
@Bill-Becker waiting for tests to pass but if they do this is ready for review as I'm getting the behavior we want and tests pass locally.

src/core/chp.jl
```diff
@@ -162,7 +164,8 @@ function CHP(d::Dict;
         :thermal_efficiency_full_load => chp.thermal_efficiency_full_load,
```
Bill-Becker, line 164: Should we assign `thermal_efficiency_full_load` and `_half_load` to zero if `is_electric_only=true`?

Bill-Becker (changes requested):
Just one requested change to assign thermal efficiency to zero if is_electric_only.

src/core/chp.jl
zolanaj: I agree; this is updated in df24331b80aaeb12af07a8f601a328ead6acfcc8.

In case we get conflicting user inputs (thermal_efficiency_full_load > 0.0 and is_electric_only == true, for example), I set up a warning and change efficiency to zero in e362ae13c40eee29f3ea0799477e3779a4372a8c as well.

Bill-Becker: I was adding a test to make sure the thermal_efficiency_full_load is being set to zero when is_electric_only == true, just under the tests you added, but I'm getting NaN for thermal_efficiency_full_load.

Bill-Becker: My change lets the CHP defaults processing first assign the thermal_efficiency value (non-zero), and then relies on your later checks if >0 then assign to zero. Before, this was setting the value for thermal_efficiency_full_load to 0.0 in custom_chp_inputs, and then when it checked for a value of NaN, it never updated the chp.thermal_efficiency_full_load (because it was 0.0) so it kept it at the default NaN (when not input).

Bill-Becker (approved):
Assuming the tests, and extra tests pass, approved!

Bill-Becker:
@zolanaj I just thought of the need to get Prime Generator defaults from the /chp_defaults endpoint, and therefore the need to be able to pass "is_electric_only" or some other flag like "prime_generator" to the CHP defaults function to get the 75% costs back. I'm not sure how to best deal with ITC - I know that's handled differently in the web tool code, so maybe we just have her code in and display that 0% ITC for combustion_turbine and recip_engine without returning those from the /chp_defaults endpoint.

zolanaj:
> @zolanaj I just thought of the need to get Prime Generator defaults from the /chp_defaults endpoint, and therefore the need to be able to pass "is_electric_only" or some other flag like "prime_generator" to the CHP defaults function to get the 75% costs back. I'm not sure how to best deal with ITC - I know that's handled differently in the web tool code, so maybe we just have her code in and display that 0% ITC for combustion_turbine and recip_engine without returning those from the /chp_defaults endpoint.

We should be covered as far as updates here, as `is_electric_only` is now an optional input to `get_chp_defaults_prime_mover_size_class`, the Julia function which we call in the API via http.jl:  https://github.com/NREL/REopt.jl/blob/78645aa21ef3b944e132c5737fb0345585050073/src/core/chp.jl#L356C1-L364C78

Testing the API locally when hooked up to this branch, the `chp_defaults` endpoint accepts `is_electric_only` as an input and outputs `federal_itc_fraction` in the default inputs part of the response.  I haven't thoroughly tested this to get the full range of outcomes though.

PR #312, opened 2023-11-06
Ignore CHP unavailability during multiple outages

Bill-Becker:
Currently,  inconsistent with the deterministic single outage scenario which ignores CHP unavailability during the outage, the stochastic multiple outages was not ignoring CHP unavailability. This PR changes that to ignore CHP unavailability during the outages, while preserving the unavailability periods for the 8760 grid-tied operation. This allows consistency in the economic dispatch between non-outages and outages scenarios.


Bill-Becker:
Github has queued the workflows/actions runner, but all tests pass locally for me so I'm going to merge.

PR #315, opened 2023-11-07
Update AVERT_2021_NOx_lb_per_kwh.csv

adfarth:
Previously incorrect NOx emissions profiles (they were the same as SO2)


adfarth:
> Looks good, can you add a line or two (and possibly a link) on where the NOx emissions profile came from for documentation purposes? Was there any data transformation?

added a bit of information in latest changelog commit! More info is in user manual

PR #345, opened 2024-01-18
Avoid modifying LCC with Obj Func Incentives

adfarth:
- We previously had some of the objective function modifiers within the "Cost" equation and some were added on using `add_to_expression!( Objective,  ...` 
- Some, but not all, of these were then subtracted out from the LCC in `results/financial.jl`
- This PR reverts to how these modifiers were previously included to avoid modifying `Costs` (instead using `@objective(m, Min, m[:Costs] + m[:ObjectiveIncentives] )`)
- The only objective fn incentive that's still removed from the lcc in `results/financial.jl` is `0.0001 * value(m[Symbol("MinChargeAdder"*_n)])` since this is contained within the `TotalElecBill`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ Classify the change according to the following categories:
 
 ## Develop 2024-01-16
 ### Fixed
-- In `reopt.jl`, group objective function incentives and avoid directly modifying m[:Costs]. Previously, some of these were incorrectly included in the reported **Financial.lcc**
+- In `reopt.jl`, group objective function incentives (into **ObjectivePenalties**) and avoid directly modifying m[:Costs]. Previously, some of these were incorrectly included in the reported **Financial.lcc**. 
 
 ## v0.39.1
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ Classify the change according to the following categories:
     ### Deprecated
     ### Removed
 
+## Develop 2024-01-16
+### Fixed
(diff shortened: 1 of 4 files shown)


Bill-Becker (commented):
Looks good. A couple of things:

1. Did you run the test_with_xpress.jl locally? (Did they pass?)
2. Thoughts on the "Incentives" part of "ObjectiveIncentives" being confusing with actual incentives like tax/production incentives? Maybe there is a standard terminology in linear programming for this type of thing ("Bias", "Steer", "Influence"...?). If you can't find anything, I'm Ok with keeping "Incentives", but my first reaction was "why is incentives not included in the Costs?.

adfarth:
> Looks good. A couple of things:
> 
> 1. Did you run the test_with_xpress.jl locally? (Did they pass?)
> 2. Thoughts on the "Incentives" part of "ObjectiveIncentives" being confusing with actual incentives like tax/production incentives? Maybe there is a standard terminology in linear programming for this type of thing ("Bias", "Steer", "Influence"...?). If you can't find anything, I'm Ok with keeping "Incentives", but my first reaction was "why is incentives not included in the Costs?.

I have now run test_with_xpress.jl locally. One test failed due to the LCC changing with this PR. Fixed in https://github.com/NREL/REopt.jl/pull/345/commits/4271de024a847ebe4ebdcb93c677b4b57f2edd63.

I changed "ObjectiveIncentives" to "ObjectivePenalties"

Bill-Becker (approved):

PR #347, opened 2024-02-06
Adding a battery storage cost constant

toddleif:
This code adds a battery storage cost constant


src/core/energy_storage/electric_storage.jl
```diff
@@ -148,10 +148,13 @@ end
     can_grid_charge::Bool = off_grid_flag ? false : true
     installed_cost_per_kw::Real = 910.0
     installed_cost_per_kwh::Real = 455.0
+    installed_cost_constant::Real = 0.0
     replace_cost_per_kw::Real = 715.0
     replace_cost_per_kwh::Real = 318.0
+    replace_cost_constant::Real = 0.0
     inverter_replacement_year::Int = 10
     battery_replacement_year::Int = 10
+    cost_constant_replacement_year::Int = 10
```
adfarth, line 157: @toddleif Can you explain the reasoning for making `cost_constant_replacement_year` a separate input, rather than just assuming the cost constant is associated with either the inverter/BOS or the battery packs?

src/core/energy_storage/electric_storage.jl
```diff
@@ -148,10 +148,13 @@ end
     can_grid_charge::Bool = off_grid_flag ? false : true
     installed_cost_per_kw::Real = 910.0
     installed_cost_per_kwh::Real = 455.0
+    installed_cost_constant::Real = 0.0
```
adfarth, line 151: We haven't done a great job of adding help text in the battery inputs, but could you add some explanatory text for the cost_constant inputs? You could also add text for the per_kwh and per_kw inputs as well while you're at it :) These could either be the same as the API input help text, or a bit more explanatory than those (closer to the web tool tooltips)

PR #348, opened 2024-02-06
Battery storage cost constant version 2

toddleif:
This new code adds a battery storage cost constant. Note that this is version 2 and there is another branch called "storage-cost-constants." An error was made in creating that feature branch, so a second version was created.

### DIFF
--- a/src/core/energy_storage/electric_storage.jl
+++ b/src/core/energy_storage/electric_storage.jl
@@ -148,10 +148,13 @@ end
     can_grid_charge::Bool = off_grid_flag ? false : true
     installed_cost_per_kw::Real = 910.0
     installed_cost_per_kwh::Real = 455.0
+    installed_cost_constant::Real = 0.0
     replace_cost_per_kw::Real = 715.0
     replace_cost_per_kwh::Real = 318.0
+    replace_cost_constant::Real = 0.0
     inverter_replacement_year::Int = 10
     battery_replacement_year::Int = 10
+    cost_constant_replacement_year::Int = 10
     macrs_option_years::Int = 7
     macrs_bonus_fraction::Float64 = 0.8
     macrs_itc_reduction::Float64 = 0.5
@@ -180,10 +183,13 @@ Base.@kwdef struct ElectricStorageDefaults
     can_grid_charge::Bool = off_grid_flag ? false : true
     installed_cost_per_kw::Real = 910.0
     installed_cost_per_kwh::Real = 455.0
+    installed_cost_constant::Real = 0.0
     replace_cost_per_kw::Real = 715.0
     replace_cost_per_kwh::Real = 318.0
+    replace_cost_constant::Real = 0.0
     inverter_replacement_year::Int = 10
     battery_replacement_year::Int = 10
+    cost_constant_replacement_year::Int = 10
     macrs_option_years::Int = 7
     macrs_bonus_fraction::Float64 = 0.8
     macrs_itc_reduction::Float64 = 0.5
@@ -218,10 +224,13 @@ struct ElectricStorage <: AbstractElectricStorage
     can_grid_charge::Bool
     installed_cost_per_kw::Real
     installed_cost_per_kwh::Real
+    installed_cost_constant::Real
     replace_cost_per_kw::Real
     replace_cost_per_kwh::Real
+    replace_cost_constant::Real
     inverter_replacement_year::Int
     battery_replacement_year::Int
+    cost_constant_replacement_year::Int
     macrs_option_years::Int
     macrs_bonus_fraction::Float64
     macrs_itc_reduction::Float64
@@ -233,6 +242,7 @@ struct ElectricStorage <: AbstractElectricStorage
     grid_charge_efficiency::Float64
     net_present_cost_per_kw::Real
     net_present_cost_per_kwh::Real
+    net_present_cost_cost_constant::Real
     model_degradation::Bool
     degradation::Degradation
     minimum_avg_soc_fraction::Float64
@@ -274,6 +284,19 @@ struct ElectricStorage <: AbstractElectricStorage
 
         net_present_cost_per_kwh -= s.total_rebate_per_kwh
 
+        net_present_cost_cost_constant = effective_cost(;
+            itc_basis = s.installed_cost_constant,
+            replacement_cost = s.cost_constant_replacement_year >= f.analysis_years ? 0.0 : s.replace_cost_constant,
+            replacement_year = s.cost_constant_replacement_year,
+            discount_rate = f.owner_discount_rate_fraction,
+            tax_rate = f.owner_tax_rate_fraction,
+            itc = s.total_itc_fraction,
+            macrs_schedule = s.macrs_option_years == 7 ? f.macrs_seven_year : f.macrs_five_year,
+            macrs_bonus_fraction = s.macrs_bonus_fraction,
+            macrs_itc_reduction = s.macrs_itc_reduction
+
+        )
+
         if haskey(d, :degradation)
             degr = Degradation(;dictkeys_tosymbols(d[:degradation])...)
         else
@@ -283,6 +306,7 @@ struct ElectricStorage <: AbstractElectricStorage
         # copy the replace_costs in case we need to change them
         replace_cost_per_kw = s.replace_cost_per_kw 
         replace_cost_per_kwh = s.replace_cost_per_kwh
+        replace_cost_constant = s.replace_cost_constant
         if s.model_degradation
             if haskey(d, :replace_cost_per_kw) && d[:replace_cost_per_kw] != 0.0 || 
                 haskey(d, :replace_cost_per_kwh) && d[:replace_cost_per_kwh] != 0.0
@@ -290,6 +314,7 @@ struct ElectricStorage <: AbstractElectricStorage
             end
             replace_cost_per_kw = 0.0
             replace_cost_per_kwh = 0.0
+            replace_cost_constant = 0.0
         end
     
         return new(
@@ -305,10 +330,13 @@ struct ElectricStorage <: AbstractElectricStorage
             s.can_grid_charge,
             s.installed_cost_per_kw,
             s.installed_cost_per_kwh,
+            s.installed_cost_constant,
             replace_cost_per_kw,
             replace_cost_per_kwh,
+            replace_cost_constant,
             s.inverter_replacement_year,
             s.battery_replacement_year,
+            s.cost_constant_replacement_year,
             s.macrs_option_years,
             s.macrs_bonus_fraction,
             s.macrs_itc_reduction,
@@ -320,6 +348,7 @@ struct ElectricStorage <: AbstractElectricStorage
             s.grid_charge_efficiency,
             net_present_cost_per_kw,
             net_present_cost_per_kwh,
+            net_present_cost_cost_constant,
             s.model_degradation,
             degr,
             s.minimum_avg_soc_fraction
--- a/src/core/reopt.jl
+++ b/src/core/reopt.jl
@@ -358,9 +358,18 @@ function build_reopt!(m::JuMP.AbstractModel, p::REoptInputs)
         end
     end
 	
+	# indicator function to say: if the energy capacity is greater than zero, then the dvBatteryIncluded binary is 1
+	@constraint(m, [b in p.s.storage.types.elec],
+		m[:dvBatteryIncluded][b] => {m[:dvStorageEnergy][b] >= 0}) # Question: is the >= actually just a greater-than sign?
+
+	# indicator function to say: if the energy capacity is zero, then the dvBatteryIncluded binary is 0
+	@constraint(m, [b in p.s.storage.types.elec],
+		!m[:dvBatteryIncluded][b] => {m[:dvStorageEnergy][b] == 0})
+
 	@expression(m, TotalStorageCapCosts, p.third_party_factor * (
 		sum( p.s.storage.attr[b].net_present_cost_per_kw * m[:dvStoragePower][b] for b in p.s.storage.types.elec) + 
-		sum( p.s.storage.attr[b].net_present_cost_per_kwh * m[:dvStorageEnergy][b] for b in p.s.storage.types.all )
+		sum( p.s.storage.attr[b].net_present_cost_per_kwh * m[:dvStorageEnergy][b] for b in p.s.storage.types.all ) +
+		sum( p.s.storage.attr[b].net_present_cost_cost_constant * m[:dvBatteryIncluded][b] for b in p.s.storage.types.elec)
 	))
 	
 	@expression(m, TotalPerUnitSizeOMCosts, p.third_party_factor * p.pwf_om *
@@ -565,6 +574,7 @@ function add_variables!(m::JuMP.AbstractModel, p::REoptInputs)
 		dvPeakDemandMonth[p.months, 1:p.s.electric_tariff.n_monthly_demand_tiers] >= 0  # Peak electrical power demand during month m [kW]
 		MinChargeAdder >= 0
         binGHP[p.ghp_options], Bin  # Can be <= 1 if require_ghp_purchase=0, and is ==1 if require_ghp_purchase=1
+		dvBatteryIncluded[p.s.storage.types.all], Bin # 0 if no battery is included, 1 if battery is included
 	end
 
 	if !isempty(p.techs.gen)  # Problem becomes a MILP
--- a/src/results/electric_storage.jl
+++ b/src/results/electric_storage.jl
@@ -32,7 +32,8 @@ function add_electric_storage_results(m::JuMP.AbstractModel, p::REoptInputs, d::
         r["storage_to_load_series_kw"] = round.(value.(discharge), digits=3)
 
         r["initial_capital_cost"] = r["size_kwh"] * p.s.storage.attr[b].installed_cost_per_kwh +
-            r["size_kw"] * p.s.storage.attr[b].installed_cost_per_kw
+            r["size_kw"] * p.s.storage.attr[b].installed_cost_per_kw +
+            p.s.storage.attr[b].installed_cost_constant
 
         if p.s.storage.attr[b].model_degradation
             r["state_of_health"] = value.(m[:SOH]).data / value.(m[:dvStorageEnergy])["ElectricStorage"];
--- a/src/results/financial.jl
+++ b/src/results/financial.jl
@@ -146,7 +146,8 @@ function initial_capex(m::JuMP.AbstractModel, p::REoptInputs; _n="")
     for b in p.s.storage.types.elec
         if p.s.storage.attr[b].max_kw > 0
             initial_capex += p.s.storage.attr[b].installed_cost_per_kw * value.(m[Symbol("dvStoragePower"*_n)])[b] + 
-                p.s.storage.attr[b].installed_cost_per_kwh * value.(m[Symbol("dvStorageEnergy"*_n)])[b]
+                p.s.storage.attr[b].installed_cost_per_kwh * value.(m[Symbol("dvStorageEnergy"*_n)])[b] +
+                p.s.storage.attr[b].installed_cost_constant
         end
     end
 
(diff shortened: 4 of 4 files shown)


src/core/energy_storage/electric_storage.jl
```diff
@@ -283,13 +306,15 @@ struct ElectricStorage <: AbstractElectricStorage
         # copy the replace_costs in case we need to change them
         replace_cost_per_kw = s.replace_cost_per_kw 
         replace_cost_per_kwh = s.replace_cost_per_kwh
+        replace_cost_constant = s.replace_cost_constant
         if s.model_degradation
             if haskey(d, :replace_cost_per_kw) && d[:replace_cost_per_kw] != 0.0 || 
                 haskey(d, :replace_cost_per_kwh) && d[:replace_cost_per_kwh] != 0.0
```
adfarth, line 312: should probably add a check for the replace_cost_constant here too, right? e.g.: 
```
||  haskey(d, :replace_cost_constant) && d[:replace_cost_constant] != 0.0
```

src/core/reopt.jl
```diff
@@ -565,6 +574,7 @@ function add_variables!(m::JuMP.AbstractModel, p::REoptInputs)
 		dvPeakDemandMonth[p.months, 1:p.s.electric_tariff.n_monthly_demand_tiers] >= 0  # Peak electrical power demand during month m [kW]
 		MinChargeAdder >= 0
         binGHP[p.ghp_options], Bin  # Can be <= 1 if require_ghp_purchase=0, and is ==1 if require_ghp_purchase=1
+		dvBatteryIncluded[p.s.storage.types.all], Bin # 0 if no battery is included, 1 if battery is included
```
adfarth, line 577: should this be `p.s.storage.types.elec` ?

src/results/electric_storage.jl
```diff
@@ -32,7 +32,8 @@ function add_electric_storage_results(m::JuMP.AbstractModel, p::REoptInputs, d::
         r["storage_to_load_series_kw"] = round.(value.(discharge), digits=3)
 
         r["initial_capital_cost"] = r["size_kwh"] * p.s.storage.attr[b].installed_cost_per_kwh +
-            r["size_kw"] * p.s.storage.attr[b].installed_cost_per_kw
+            r["size_kw"] * p.s.storage.attr[b].installed_cost_per_kw +
+            p.s.storage.attr[b].installed_cost_constant
```
adfarth, line 36: will this report a capital cost if the recommended storage size is zero?

src/results/financial.jl
```diff
@@ -146,7 +146,8 @@ function initial_capex(m::JuMP.AbstractModel, p::REoptInputs; _n="")
     for b in p.s.storage.types.elec
         if p.s.storage.attr[b].max_kw > 0
             initial_capex += p.s.storage.attr[b].installed_cost_per_kw * value.(m[Symbol("dvStoragePower"*_n)])[b] + 
-                p.s.storage.attr[b].installed_cost_per_kwh * value.(m[Symbol("dvStorageEnergy"*_n)])[b]
+                p.s.storage.attr[b].installed_cost_per_kwh * value.(m[Symbol("dvStorageEnergy"*_n)])[b] +
+                p.s.storage.attr[b].installed_cost_constant
```
adfarth, line 150: will this report a capital cost if the recommended storage size is zero?

src/core/reopt.jl
adfarth, line 577: Should this be indexed on `p.s.storage.types.elec` instead of all?

src/core/reopt.jl
```diff
@@ -358,9 +358,18 @@ function build_reopt!(m::JuMP.AbstractModel, p::REoptInputs)
         end
     end
 	
+	# indicator function to say: if the energy capacity is greater than zero, then the dvBatteryIncluded binary is 1
```
adfarth, line 361: I think we should avoid adding indicator constraints that will be applied in every REopt run (or even every run with BESS) since some solvers struggle with these. @zolanaj do you have any suggestions for the best way to reconfigure this constraint to not be an indicator constraint?

adfarth:
@toddleif  in general this is looking good! In addition to my in-line comments above, just a few high-level notes: 
- Can you add tests to `test/test_with_xpress.jl` and `runtests.jl` that check that things are working as expected with your new addition? This could check, for example, some expected financial outputs, that you have manually calculated in Excel.
- I think we need to reconfigure the constraint to not be an indicator constraint, if possible. I suspect this is why so many of the hightests are failing. We'll want all tests to be passing before we merge.

src/core/energy_storage/electric_storage.jl
toddleif: Added, thanks!

src/results/electric_storage.jl
toddleif: I could be misunderstanding the code, but I believe that the "if r["size_kwh"] != 0" expression on line 27 ensures that the r[“initial_capital_cost”] output only gets generated if the battery size is > 0 kWh. So if the battery kWh = 0, then r[“initial_capital_cost”] would not be created, so the installed cost constant would be not included.

src/results/financial.jl
toddleif: I think it would. I added a multiplier to the installed_cost_constant on line 150. The multiplier should ensure that the installed_cost_constant is only included if the "m[Symbol("dvStorageEnergy"*_n)]" variable is greater than 0.

src/core/reopt.jl
toddleif: Fixed to be p.s.storage.types.elec

src/core/reopt.jl
toddleif: As a temporary solution before the implementation of the Big-M constraint, I added an if statement which ensures that the indicator constraints will only be applied if the install_cost_constant or replace_cost_constant is not zero.

toddleif: I removed the indicator constraints, and reformulated the equations so they don't need the Big-M constraint either

src/core/energy_storage/electric_storage.jl
```diff
@@ -36,6 +36,10 @@ worth factor is used in two different ways, depending on the `maintenance_strate
     - `replace_cost_per_kwh`
     - `inverter_replacement_year`
     - `battery_replacement_year`
+    - `replace_cost_per_kw`
```
adfarth, line 39: @toddleif these seem to be repeats of the inputs already listed. My original comment was asking if you could add the new inputs that you have created to this list (the cost_constant ones)

adfarth:
@toddleif could you please update the `CHANGELOG.md` file with a description of these changes? You can see other entries in there for examples.

src/core/reopt.jl
```diff
@@ -387,9 +387,19 @@ function build_reopt!(m::JuMP.AbstractModel, p::REoptInputs)
         end
     end
 	
+	# Include the electric storage cost constants only if the installed_cost_constant or the replace_cost_constant is not zero
+	for b in p.s.storage.types.elec
```
adfarth, line 391: @toddleif Out of curiosity, why did you add this here instead of `constraints/storage_constraints.jl`?

test/runtests.jl
```diff
@@ -101,6 +101,66 @@ else  # run HiGHS tests
...
+    @testset "Solar and ElectricStorage w/ zero ElectricStorage cost constants" begin
+        m1 = Model(optimizer_with_attributes(HiGHS.Optimizer, "output_flag" => false, "log_to_console" => false))
+        m2 = Model(optimizer_with_attributes(HiGHS.Optimizer, "output_flag" => false, "log_to_console" => false))
+        d = JSON.parsefile("./scenarios/pv_storage.json");
+        
+        d["ElectricStorage"]["installed_cost_constant"] = 0
+        d["ElectricStorage"]["replace_cost_constant"] = 0
+        d["ElectricStorage"]["cost_constant_replacement_year"] = 10
+    
+        d["Settings"] = Dict{Any,Any}("add_soc_incentive" => false)
```
adfarth, line 113: @toddleif out of curiosity why did you include setting the add_soc_incentive to false?

test/runtests.jl
```diff
@@ -101,6 +101,66 @@ else  # run HiGHS tests
...
+        s = Scenario(d)
+        inputs = REoptInputs(s)
+        results = run_reopt([m1,m2], inputs)
+        
+        UpfrontCosts_NoIncentive = (results["ElectricStorage"]["size_kw"]*d["ElectricStorage"]["installed_cost_per_kw"] ) +
+                                   (results["ElectricStorage"]["size_kwh"]*d["ElectricStorage"]["installed_cost_per_kwh"]) + 
+                                   d["ElectricStorage"]["installed_cost_constant"] +
+                                   (results["PV"]["size_kw"]*d["PV"]["installed_cost_per_kw"])
+    
+        @test results["PV"]["size_kw"] ≈ 216.6667 atol=0.01
```
adfarth, line 123: @toddleif just because so many other things could affect both PV sizing and LCC, it could make sense to avoid testing for these specific values here. That way, we don't have to update this test every time other changes are made that don't relate to the BESS cost constant value.

test/runtests.jl
```diff
@@ -101,6 +101,66 @@ else  # run HiGHS tests
         @test r["ElectricStorage"]["size_kwh"] ≈ 83.3 atol=0.1
     end
 
+    @testset "Solar and ElectricStorage w/ zero ElectricStorage cost constants" begin
```
adfarth, line 104: @toddleif  would it be worth including a test in which the cost constant is provided, but the BESS max_kw is 0, and checking to see if the cost constant gets added when BESS size is 0?

src/core/reopt.jl
```diff
@@ -387,9 +387,19 @@ function build_reopt!(m::JuMP.AbstractModel, p::REoptInputs)
...
+		if p.s.storage.attr[b].installed_cost_constant != 0 || p.s.storage.attr[b].replace_cost_constant != 0
+			@constraint(m, [b in p.s.storage.types.elec], m[:dvStorageEnergy][b] <= p.s.storage.attr[b].max_kwh * m[:dvBatteryIncluded][b]) # if the dvBatteryIncluded binary is 1, then the storage energy capacity can be greater than 0, but then the battery cost constant is also included in the costs		
+		else
+			m[:dvBatteryIncluded][b] == 0
+		end
+	end
+	
 	@expression(m, TotalStorageCapCosts, p.third_party_factor * (
 		sum( p.s.storage.attr[b].net_present_cost_per_kw * m[:dvStoragePower][b] for b in p.s.storage.types.elec) +
```
adfarth, line 400: @zolanaj do you happen to know if this is a mistake that the `TotalStorageCapCosts` only sums across electric storage types for the power capacity and across all storage types for the energy capacity?

src/core/reopt.jl
toddleif: Thank you for bringing this up. I moved this code to a function in `constraints/storage_constraints.jl`.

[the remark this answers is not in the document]
toddleif: Unless I'm mistaken, I believe this is a duplicated comment from above, which I believe I addressed

test/runtests.jl
toddleif: That's a good point. I removed that line, which will keep add_soc_incentive as the default, true.

test/runtests.jl
toddleif: That's a great point, thank you. I removed a number of the values to test.

test/runtests.jl
toddleif: That's a great idea, thank you. I added a test for that scenario.

src/constraints/storage_constraints.jl
```diff
@@ -128,6 +128,19 @@ function add_elec_storage_dispatch_constraints(m, p, b; _n="")
     end
 end
 
+function add_elec_storage_cost_constant_constraints(m, p, b; _n="")
+    # Include the electric storage cost constants only if the installed_cost_constant or the replace_cost_constant is not zero
+    for b in p.s.storage.types.elec
```
adfarth, line 133: @toddleif since "b" is a function input, I don't think this function should include a for loop over b 's in storage.types.elec. Please correct me if I'm wrong

src/core/energy_storage/electric_storage.jl
```diff
@@ -169,10 +172,13 @@ end
     can_grid_charge::Bool = off_grid_flag ? false : true
     installed_cost_per_kw::Real = 910.0
     installed_cost_per_kwh::Real = 455.0
+    installed_cost_constant::Real = 0.0
```
adfarth, line 175: @toddleif could you please add a brief description of this input to this markdown area?

src/results/financial.jl
```diff
@@ -146,7 +146,8 @@ function initial_capex(m::JuMP.AbstractModel, p::REoptInputs; _n="")
     for b in p.s.storage.types.elec
         if p.s.storage.attr[b].max_kw > 0
             initial_capex += p.s.storage.attr[b].installed_cost_per_kw * value.(m[Symbol("dvStoragePower"*_n)])[b] + 
-                p.s.storage.attr[b].installed_cost_per_kwh * value.(m[Symbol("dvStorageEnergy"*_n)])[b]
+                p.s.storage.attr[b].installed_cost_per_kwh * value.(m[Symbol("dvStorageEnergy"*_n)])[b] +
+                (p.s.storage.attr[b].installed_cost_constant * ceil(value.(m[Symbol("dvStorageEnergy"*_n)])[b]/(value.(m[Symbol("dvStorageEnergy"*_n)])[b] + 1)))
```
adfarth, line 150: @toddleif why is the `* ceil...` multiplier necessary here?

test/runtests.jl
```diff
@@ -160,18 +160,102 @@ else  # run HiGHS tests
             @test r["ElectricStorage"]["size_kwh"] ≈ 83.3 atol=0.1
         end
 
-        # TODO test MPC with outages
```
adfarth, line 163: @toddleif did you remove this test (or change it to the condensed version below), or is this a product of some merge with develop?

test/runtests.jl
```diff
@@ -160,18 +160,102 @@ else  # run HiGHS tests
...
+        UpfrontCosts_NoIncentive = (results["ElectricStorage"]["size_kw"]*d["ElectricStorage"]["installed_cost_per_kw"] ) +
+                                   (results["ElectricStorage"]["size_kwh"]*d["ElectricStorage"]["installed_cost_per_kwh"]) + 
+                                   d["ElectricStorage"]["installed_cost_constant"] +
+                                   (results["PV"]["size_kw"]*d["PV"]["installed_cost_per_kw"])
+        
+        ReplacementCosts_NoIncentive = (results["ElectricStorage"]["size_kw"]*d["ElectricStorage"]["replace_cost_per_kw"] ) +
+                                        (results["ElectricStorage"]["size_kwh"]*d["ElectricStorage"]["replace_cost_per_kwh"]) + 
+                                        d["ElectricStorage"]["replace_cost_constant"] 
+
+        @test results["Financial"]["initial_capital_costs"] ≈ UpfrontCosts_NoIncentive rtol=1e-5  #Note: the value of the ["Financial"]["initial_capital_costs"] dictionary key does not include incentives
```
adfarth, line 185: @toddleif is the point of this test to ensure that the cost constants of 0 do indeed get used in the initial and replacement cost calculations?

test/runtests.jl
```diff
@@ -160,18 +160,102 @@ else  # run HiGHS tests
...
+        UpfrontCosts_NoIncentive = results["PV"]["size_kw"]*d["PV"]["installed_cost_per_kw"]
+        
+        ReplacementCosts_NoIncentive = 0
+
+        @test results["Financial"]["initial_capital_costs"] ≈ UpfrontCosts_NoIncentive rtol=1e-5
+        @test results["Financial"]["replacements_future_cost_after_tax"] ≈ ReplacementCosts_NoIncentive  rtol=1e-5
+    
+    end 
+
+    @testset "Outage with Generator" begin
```
adfarth, line 240: @toddleif did you add this "Outage with Generator" test as well?

test/test_with_xpress.jl
```diff
@@ -439,6 +439,37 @@ end
     @test round(sum(r["ElectricStorage"]["soc_series_fraction"]), digits=2) / 8760 >= 0.7199
 end
 
+@testset "Solar and ElectricStorage w/ ElectricStorage cost constants" begin
```
adfarth, line 442: @toddleif I think you could remove all of these changes to `test_with_xpress.jl`. We don't have to update this file any more

adfarth:
@toddleif I think this is looking good! I left a few comments throughout, and have noticed just two more places where changes may need to be made: 
- It looks like you'll need to make a few changes to `proforma.jl`. See e.g. line 71 "`# calculate Storage o+m costs, incentives, and depreciation`" (there may be other places in there that require updates too)
- `outage_constraints.jl` line 80 where I think the cost constant should be added to the "bigM" value being subtracted from the capital costs
```
@constraint(m,
            m[:dvMGStorageUpgradeCost] >= p.s.financial.microgrid_upgrade_cost_fraction * m[:TotalStorageCapCosts] - (
                p.s.financial.microgrid_upgrade_cost_fraction * p.third_party_factor * (
                    sum( p.s.storage.attr[b].net_present_cost_per_kw * p.s.storage.attr[b].max_kw for b in p.s.storage.types.elec) + 
                    sum( p.s.storage.attr[b].net_present_cost_per_kwh * p.s.storage.attr[b].max_kwh for b in p.s.storage.types.all )
                ) * (1-m[:binMGStorageUsed])  # Big-M is capital cost of battery with max size kw and kwh
            )
        )
```

adfarth:
Also noting that the tests did fail previously (looks like they may have timed out) so I re-ran them on Tucker's last commit. The commit that I made was just a change to the Changelog, so only the documentation tests passed, which is misleading :)

src/constraints/storage_constraints.jl
toddleif: That makes sense. I just removed the for loop.

src/core/energy_storage/electric_storage.jl
toddleif: That's a great idea. I just added a description.

src/results/financial.jl
toddleif: That's a great idea. I just switched it to binIncludeStorageCostConstant.

test/runtests.jl
toddleif: That appears to be a product of a merge with develop -- I'm not sure why it was deleted. I added it back in to the code.

test/runtests.jl
toddleif: I didn't add that. I'm thinking maybe it was deleted in develop because I no longer see it in develop. I deleted that test to align the file with the version in develop.

test/test_with_xpress.jl
toddleif: Done

test/runtests.jl
toddleif: I think this test was more for just making sure the code works if the cost constant is set to zero. But I don't think that's worth putting in a test. I just deleted it.

toddleif:
> @toddleif I think this is looking good! I left a few comments throughout, and have noticed just two more places where changes may need to be made:
> 
> * It looks like you'll need to make a few changes to `proforma.jl`. See e.g. line 71 "`# calculate Storage o+m costs, incentives, and depreciation`" (there may be other places in there that require updates too)
> * `outage_constraints.jl` line 80 where I think the cost constant should be added to the "bigM" value being subtracted from the capital costs
> 
> ```
> @constraint(m,
>             m[:dvMGStorageUpgradeCost] >= p.s.financial.microgrid_upgrade_cost_fraction * m[:TotalStorageCapCosts] - (
>                 p.s.financial.microgrid_upgrade_cost_fraction * p.third_party_factor * (
>                     sum( p.s.storage.attr[b].net_present_cost_per_kw * p.s.storage.attr[b].max_kw for b in p.s.storage.types.elec) + 
>                     sum( p.s.storage.attr[b].net_present_cost_per_kwh * p.s.storage.attr[b].max_kwh for b in p.s.storage.types.all )
>                 ) * (1-m[:binMGStorageUsed])  # Big-M is capital cost of battery with max size kw and kwh
>             )
>         )
> ```

@adfarth    I addressed these edits too, thanks for catching these as well!

src/core/energy_storage/electric_storage.jl
adfarth: @toddleif I just modified the description to be in-line with the inputs and to just give a short high-level description of what that input is. Could you review?

Bill-Becker:
@adfarth I **think** the tests will all pass now, and this is up to date with the develop branch. I added the %-based O&M that we talked about (see PR description), and updated the cost defaults to ATB 2024 (latest). I added the CHANGELOG stuff and a warning about adding a binary variable. I added a test for the %-based O&M.

PR #351, opened 2024-02-21
Pv default update

bpulluta:
### Changed
- In `core/pv.jl` a change was made to make sure we are using the same assumptions as PVWatts guidelines, the default `tilt` angle for a fixed array should be 20 degrees, irrespective of it being a rooftop or ground-mounted (open-rack) system. By default the `tilt` will be set to 20 degrees for ground-mount and rooftop, and 0 degrees for axis-tracking (`array_type = 3 or 4`)

> "The PVWatts® default value for the tilt angle depends on the array type: For a fixed array, the default value is 20 degrees, and for one-axis tracking the default value is zero. A common rule of thumb for fixed arrays is to set the tilt angle to the latitude of the system's location to maximize the system's total electrical output over the year. Use a lower tilt angle favor peak production in the summer months when the sun is high in the sky, or a higher tilt angle to increase output during winter months. Higher tilt angles tend to cost more for racking and mounting hardware, and may increase the risk of wind damage to the array."


bpulluta:
@Bill-Becker 

Let me know when you have a chance to review these changes regarding the PV tilt default

Bill-Becker (approved):
Looks good!

Bill-Becker:
I forgot to ask, can you run the test_with_xpress.jl tests locally? Those are not tested on GitHub Actions.

bpulluta:
> I forgot to ask, can you run the test_with_xpress.jl tests locally? Those are not tested on GitHub Actions.

@Bill-Becker I did not, I'll add the additional test that I added for the highs test and then run the test with `test_with)xpress.jl`

Bill-Becker:
@bpulluta one reviewer is enough, so you can remove @adfarth from needing to review. I agree, you are good to merge.

Bill-Becker:
> > @bpulluta no need to duplicate a test you added with HiGHS in the test_with_xpress set, if that's what you were saying. The more test coverage we can do with HiGHS only, the better, because they can run on GitHub Actions.
> 
> All xpress test are successfully passing. I think it might be ready to merge into develop Screenshot 2024-03-07 at 10 28 11 AM
> 
> @Bill-Becker, do you know if the reason why `Run tests / highstests (1.8, x64, macOS-11) (push)` is not passing is due to the underscores I am using in the branch name?

The Mac tests are finicky; if the other tests pass, that's good enough. It's a GitHub Actions runner/environment issue with the Mac OS container.

adfarth:
@bpulluta and @Bill-Becker I think we should discuss this one as a team before merging. Dan might have opinions, and I think we currently use 10 deg for rooftop to address potential wind loading issues.

bpulluta:
> @bpulluta and @Bill-Becker I think we should discuss this one as a team before merging. Dan might have opinions, and I think we currently use 10 deg for rooftop to address potential wind loading issues.

We reached out to both Dan and Nate for feedback. See the email i sent for context.

Bill-Becker:
@bpulluta I think we are good to merge this.

bpulluta:
> @bpulluta I think we are good to merge this.

@Bill-Becker, I'm gonna finalize it now. I just merged Amanda's recent changes into this branch. Seems like the windows test is not passing anymore but the ubuntu is. **Am i still okay to proceed?**

PR #357, opened 2024-03-08
v0.41.0 master < develop

adfarth:
See #352  and #235


Bill-Becker:
@adfarth have you run test_with_xpress.jl locally?

adfarth:
> @adfarth have you run test_with_xpress.jl locally?

Thank you for the reminder! I had not, but now I have and the tests have passed locally.

test/runtests.jl
```diff
@@ -1895,6 +1982,7 @@ else  # run HiGHS tests
                     end
                 end 
                 
+                # TODO: refresh tests and test values as needed after Cambium update; revised values are in test_with_xpress.jl
```
Bill-Becker, line 1985: test_with_xpress.jl is using a different/updated Cambium version/data?

test/test_with_xpress.jl
```diff
@@ -1377,6 +1377,82 @@ end
     @test calculated_ghp_capex ≈ reopt_ghp_capex atol=300
 end
 
+@testset "Cambium Emissions" begin
```
Bill-Becker, line 1380: Is this the same test as in "runtests.jl" with the HiGHS solver? If so, we don't need to add it here. We should be prioritizing adding tests to runtests.jl with an open source solver whenever possible.

Bill-Becker:
@adfarth also, just confirming again that you ran test_with_xpress.jl locally with this develop branch.

test/runtests.jl
adfarth: @Bill-Becker These are the emissions tests that Alex commented out a while back because they weren't compatible with the HiGHS solver. The test values are slightly updated because of the AVERT data update that we made in this PR. Alex added this note to indicate that once we are able to run these tests with a non-Xpress solver, we should update the values here. Does that all make sense?

test/test_with_xpress.jl
adfarth: yes, this was exactly the same as the one in runtests. I deleted this testset in `test_with_xpress` in https://github.com/NREL/REopt.jl/pull/357/commits/cb6d9d9da27e0dfc26ad5c1cbceea9ac0f3a88f1

adfarth:
> @adfarth also, just confirming again that you ran test_with_xpress.jl locally with this develop branch.

Yes, confirmed that I ran this on the develop branch.

adfarth:
> Is there breaking changes other than expected differences in emissions accounting and results when doing emissions reduction goals or CO2 cost in the optimization?
> 
> Otherwise, just a couple of questions about runtests.jl versus test_with_xpress.jl.

Summary of changes at a high level: 
- Emissions results will change for all analyses (Cambium now used by default for climate emissions, health emissions from AVERT are updated to more recent version) 
- If an emissions reduction goal is added or CO2 cost considered in optimization, then optimal results will change
- (Breaking) Updated `electric_utility` **emissions_region** to **avert_emissions_region** and **distance_to_emissions_region_meters** to **distance_to_avert_emissions_region_meters** in `src/electric_utility.jl` and `results/electric_utility.jl`
- (Breaking) Changed name of exported function **emissions_profiles** to **avert_emissions_profiles**
- Changed calculation of all `annual` emissions results (e.g. **Site.annual_emissions_tonnes_CO2**) to simple annual averages (lifecycle emissions divided by analysis_years). This is because the default climate emissions from Cambium are already levelized over the analysis horizon and therefore "year_one" emissions cannot be easily obtained.

Bill-Becker (approved):
Thanks for the responses - they all look good. Related to the discussion of Xpress versus HiGHS tests - does anything you added with this PR require Xpress to work properly? We are about to take Xpress down from the publicly available API.

adfarth:
> Thanks for the responses - they all look good. Related to the discussion of Xpress versus HiGHS tests - does anything you added with this PR require Xpress to work properly? We are about to take Xpress down from the publicly available API.

Not that I can think of. I think there were some emissions modeling capabilities that were timing out or failing with HiGHS (hence the commented out tests in runtests), but those are not explicitly related to any change introduced by this PR (this is just changing the data sources, not any constraints)

PR #366, opened 2024-03-26
Solver-specific workarounds for Big-M Constraints

zolanaj:
### Added 
- In `src/settings.jl`, added new const **INDICATOR_COMPATIBLE_SOLVERS**
- In `src/settings.jl`, added new member **solver_name** within the settings object.  This is currently not connected to the solver but does determine whether indicator constraints are modeled or if their big-M workarounds are used.
- Added replacements for indicator constraints with the exception of battery degradation, which is implemented in a separate model, and FlexibleHVAC.  TODO's have been added for these remaining cases.

### Fixed
- Fixed previously broken tests using HiGHS in `test/runtests.jl` due to solver incompatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,15 @@ Classify the change according to the following categories:
     ### Deprecated
     ### Removed
 
+## Develop 2024-03-26
+### Added 
+- in `src/settings.jl`, added new const **INDICATOR_COMPATIBLE_SOLVERS**
+- in `src/settings.jl`, added new member **solver_name** within the settings object.  This is currently not connected to the solver but does determine whether indicator constraints are modeled or if their big-M workarounds are used.
+- added replacements for indicator constraints with the exception of battery degradation, which is implemented in a separate model, and FlexibleHVAC.  TODO's have been added for these remaining cases.
+
+### Fixed
+- Fixed previously broken tests using HiGHS in `test/runtests.jl` due to solver incompatibility.
+
 ## v0.43.0
 ### Fixed
 - `simple_payback_years` calculation when there is export credit
--- a/src/REopt.jl
+++ b/src/REopt.jl
@@ -103,6 +103,7 @@ const EMISSIONS_DECREASE_DEFAULTS = Dict(
     "SO2" => 0.02163,
     "PM25" => 0.02163
 )
+const INDICATOR_COMPATIBLE_SOLVERS = ["Cbc","CPLEX","Xpress","SCIP"]
 
 include("logging.jl")
 
--- a/src/constraints/battery_degradation.jl
+++ b/src/constraints/battery_degradation.jl
@@ -106,6 +106,7 @@ function add_degradation(m, p; b="ElectricStorage")
         @warn "Adding binary and indicator constraints for 
          ElectricStorage.degradation.maintenance_strategy = \"replacement\". 
          Not all solvers support indicators and some are slow with integers."
+        # TODO import the latest battery degradation model in the degradation branch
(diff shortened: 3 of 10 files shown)


zolanaj:
@Bill-Becker tests should pass now, though we won't know officially for a couple hours.  A few notes beyond the PR description: 
- The tests have slowed down a lot which I think we expected.  The entire test suite in GitHub actions using HiGHS took about 3 hours in the last push.  I don't know which tests are causing the delay right now unfortunately but I suspect stochastic outages may be the biggest culprit right now due to the number of constraints.
- One potential offender is the use of `p.max_sizes` for a big-M when we don't have a good value in there.  I can bring in some of the "reasonable max size" work from v2 but I ran out of time yesterday.  I can work on this today but I don't know how early today I can finish it.  Since this is time sensitive, please let me know if you'd rather that be in a separate PR, and I can do that to allow this to be deployed sooner.
- While there's a TODO in place for FlexibleHVAC and Battery Degradation, we have already implemented the degradation along with other updates in the degradation branch: https://github.com/NREL/REopt.jl/blob/74f3f945599bbf6f0d5f56b2a0f5fc4af1a6fdf2/src/constraints/battery_degradation.jl#L111
I know that's a lower priority but FYI it's just a matter of importing that with the other work in that branch and we may be able to remove that TODO.  
- FlexibleHVAC includes constraint with free variables (if we're using Celsius) so that one will take a bit more time to get good big-M values.

Bill-Becker:
@zolanaj Amazing work! It looks promising that at least the tests pass in the Windows environment, and I believe the other environments failed because they were taking too long. Do you know why this PR is not showing those tests as required "checks"? It looks like just the documentation check is being enforced here.

Bill-Becker:
@zolanaj did you happen to run the `tests_with_xpress.jl` tests locally? I'm trying to make sure we didn't break those ones, accidentally. I just kicked off running those locally, but let me know if you've already done that. 

Did you test using Cbc and/or SCIP to see if they were still using the indicator constraints? Only HiGHS is supposed to use these Big M notation constraints, right?

Bill-Becker:
@zolanaj looks like the inputs/scenarios for the test_with_xpress.jl tests need `Settings.solver_name == "Xpress"` or else they'll use Big M. Makes sense.

Bill-Becker:
> @zolanaj Amazing work! It looks promising that at least the tests pass in the Windows environment, and I believe the other environments failed because they were taking too long. Do you know why this PR is not showing those tests as required "checks"? It looks like just the documentation check is being enforced here.

Well, I just made a commit that made all of the test_with_xpress.jl tests pass locally for me, and that seemed to kick-off the steps/checks that I was asking about, so nevermind on that.

Bill-Becker (approved):
Looks great to me, and I just made one commit for fixing test_with_xpress.jl tests for local checking.

zolanaj:
> Looks great to me, and I just made one commit for fixing test_with_xpress.jl tests for local checking.

Thanks a bunch @Bill-Becker!  Monitoring for tests to pass and will merge assuming that's the case.  Thanks for the check and fix on the Xpress tests.

zolanaj:
> @zolanaj Amazing work! It looks promising that at least the tests pass in the Windows environment, and I believe the other environments failed because they were taking too long. Do you know why this PR is not showing those tests as required "checks"? It looks like just the documentation check is being enforced here.

I think that when you edit the README in the GitHub repo through the web interface it only does the documentation check and skips the test suite.  I'm making a mental TODO to see if this is always the case in GitHub actions or if we can change this.

PR #372, opened 2024-04-08
Add ASHP

atpham88:
# Notes
- This PR adds space heating/cooling ASHP and water heating ASHP to REopt


zolanaj:
@Bill-Becker tagging you for a second set of eyes on this PR; once testing improvements have been merged, so will add-process-heat-load, and then the target ranch will be moved to develop.  I can then work on the version update.

test/test_with_xpress.jl
```diff
@@ -1734,6 +1734,41 @@ end
...
+    annual_energy_supplied = 87600 + annual_ashp_consumption
+
+    #Second run: ASHP produces the required heat with free electricity
+    @test results["ASHP"]["size_mmbtu_per_hour"] ≈ 0.8 atol=0.1
+    @test results["ASHP"]["annual_thermal_production_mmbtu"] ≈ annual_thermal_prod rtol=1e-4
+    @test results["ASHP"]["annual_electric_consumption_kwh"] ≈ annual_ashp_consumption rtol=1e-4
+    @test results["ElectricUtility"]["annual_energy_supplied_kwh"] ≈ annual_energy_supplied rtol=1e-4
+
+end
+
```
zolanaj, line 1771: Just confirming - we're going to excise this test file, right?  (I can do that in a separate PR, `removing the test_with_xpress.jl` file and the testset header for imported xpress tests.  I didn't want to do that for the testing speed improvements PR because I wanted the test adjustments to be clear to the reviewer.)

zolanaj (commented):
Thank you for getting this together @atpham88!  My edits were focused on allowing the ASHP system to cool as well, and then some additional things beyond that to get all tests passing.  I'm passing to @Bill-Becker for a second set of eyes now.

src/core/ashp.jl
```diff
@@ -0,0 +1,127 @@
...
+    can_serve_dhw::Union{Bool, Nothing} = nothing # If ASHP can supply heat to the domestic hot water load
+    can_serve_space_heating::Union{Bool, Nothing} = nothing # If ASHP can supply heat to the space heating load
+    can_serve_process_heat::Union{Bool, Nothing} = nothing # If ASHP can supply heat to the process heating load
+    can_serve_cooling::Union{Bool, Nothing} = nothing # If ASHP can supply heat to the cooling load
+)
+```
+"""
+function ASHP(;
+        min_mmbtu_per_hour::Real = 0.0,
+        max_mmbtu_per_hour::Real = BIG_NUMBER,
```
Bill-Becker, line 47: We should use `ton` instead of `mmbtu_per_hour` for heat pump thermal production. That's how they typically state the capacity in the industry. So this should apply to min, max, installed cost, O&M cost, and any incentives which I see we don't have a generic $/ton or percent-based incentive currently.

src/results/ashp.jl
```diff
@@ -0,0 +1,126 @@
+# REopt®, Copyright (c) Alliance for Sustainable Energy, LLC. See also https://github.com/NREL/REopt.jl/blob/master/LICENSE.
+
+"""
+`ASHP` results keys:
+- `size_mmbtu_per_hour`  # Thermal production capacity size of the ASHP [MMBtu/hr]
+- `electric_consumption_series_kw`  # Fuel consumption series [kW]
+- `annual_electric_consumption_kwh`  # Fuel consumed in a year [kWh]
+- `thermal_production_series_mmbtu_per_hour`  # Thermal energy production series [MMBtu/hr]
```
Bill-Becker, line 8: Related but counter to changing the sizing capacity/basis to ton, I'm not opposed to reporting heating load production as mmbtu/hr because that's how we have other heating results. Let's just make sure we're using the same naming convention for thermal/heating/cooling results as other heating and cooling techs, like CHP and GHP. I think this is consistent, and we're differentiating between heating production and cooling production based on the units.

src/results/ashp.jl
```diff
@@ -0,0 +1,126 @@
...
+
+!!! note "'Series' and 'Annual' energy outputs are average annual"
+	REopt performs load balances using average annual production values for technologies that include degradation. 
+	Therefore, all timeseries (`_series`) and `annual_` results should be interpretted as energy outputs averaged over the analysis period. 
+
+"""
+
+function add_ashp_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _n="")
+    r = Dict{String, Any}()
+    r["size_mmbtu_per_hour"] = round(value(m[Symbol("dvSize"*_n)]["ASHP"]) / KWH_PER_MMBTU, digits=3)
```
Bill-Becker, line 28: Is this size based on the peak coincident heating and cooling thermal production?

test/runtests.jl
```diff
@@ -2297,6 +2297,56 @@ else  # run HiGHS tests
 
         end
 
+        @testset "ASHP" begin
+            #Case 1: Boiler produces the required heat instead of ASHP - ASHP is not purchased
```
Bill-Becker, line 2301: Do we have an equivalent way to "force in space heating ASHP", like we do for GHP, right now? I think the new `ExistingBoiler.retire_in_optimal` would "force in space heating **and DHW** ASHP", but we don't have a way to just force in "DHW ASHP".

Bill-Becker: I guess it would be setting the min size, but in order to do that accurately you'd have to know what the peak heating or coincident peak heating + cooling thermal load.

test/test_with_cplex.jl
```diff
@@ -163,6 +163,45 @@ end
     @test r["ElectricTariff"]["year_one_demand_cost_before_tax"] ≈ expected_demand_cost
 end
 
+@testset "ASHP" begin
+    d = JSON.parsefile("./scenarios/ashp.json")
```
Bill-Becker, line 167: Is someone currently testing with CPLEX?!

test/test_with_xpress.jl
Bill-Becker: Yes, we should be doing all tests in HiGHS, and we don't need to add to test_with_xpress.jl.

Bill-Becker (commented):
Thanks Alex and An, looks good! I have one units change request and a few other comments.

test/runtests.jl
zolanaj: Would the force-in function set set all other heating production variables to zero, i.e., would we allow only ASHP to be the source of a given heating load?  I ask because we'd also probably exclude other possible techs like the CHP from providing heat as well.  (And, if we do this, does that mean no storage as well?  CHP would serve load using storage as a middleman otherwise.)

src/results/ashp.jl
zolanaj: yes - see the constraint here: 
```
function add_ashp_heating_cooling_constraints(m, p; _n="")
    @constraint(m, [t in setdiff(intersect(p.techs.cooling, p.techs.heating), p.techs.ghp), ts in p.time_steps],
        sum(m[Symbol("dvHeatingProduction"*_n)][t,q,ts] for q in p.heating_loads) + m[Symbol("dvCoolingProduction"*_n)][t,ts] <= m[Symbol("dvSize"*_n)][t]
    )
end
```

src/core/ashp.jl
atpham88: I converted mmbtu to ton for ASHP capacity here https://github.com/NREL/REopt.jl/commit/032e0d6cd659b47851b2a41c45bfe06c434afd4f

src/results/ashp.jl
atpham88: Addressed here https://github.com/NREL/REopt.jl/commit/032e0d6cd659b47851b2a41c45bfe06c434afd4f

test/test_with_cplex.jl
atpham88: Deleted this. I performed this test for my own only but accidentally pushed it.

src/core/ashp.jl
Bill-Becker: The unit "ton" is actually a thermal **power/rate**, so no need to add "per_hour" for the min/max/installed_cost/om_cost fields.

test/runtests.jl
zolanaj: Ok, thanks for the clarification on this @Bill-Becker.  I'm going to also include "retire_in_optimal" for the ExistingChiller as well since the ASHP space heating and cooling may replace that as well (and could also compete with the AbsorptionChiller and GHP for serving cooling load).  I'll also add a test for retiring these in the ASHP test suite.

Bill-Becker: Good idea!

src/core/ashp.jl
zolanaj: addressed here: https://github.com/NREL/REopt.jl/pull/372/commits/03f7172af860c06d6dbe994f343afb5d299d0572

and here: https://github.com/NREL/REopt.jl/pull/372/commits/012cc400bdbf7e3c75f460810db26c4e02e9e6c3

test/test_with_cplex.jl
zolanaj: yes, removed here - https://github.com/NREL/REopt.jl/pull/372/commits/05bcdc048fed964dab2e094c6144d087901e7abb

test/runtests.jl
zolanaj: Ok, with the most recent push this has been implemented via the new function here: 
https://github.com/NREL/REopt.jl/blob/efc94a29a501b0bf3d283cfba63ab3edc01f5896/src/constraints/thermal_tech_constraints.jl#L81-L86

And the new checks for other cooling and heating techs here (this is edited from the original PR with `retire_in_optimal` for the boiler):
https://github.com/NREL/REopt.jl/blob/efc94a29a501b0bf3d283cfba63ab3edc01f5896/src/core/reopt.jl#L299-L335

I have warnings in place instead of errors if (1) there are more heating techs than the existing boiler, (2) the existing boiler is to be retired, and (3) the new techs can't meet all the heating loads without the existing boiler present.  I just warn and run the problem with the boiler instead of an error, but we can raise an exception if you think that's more appropriate.

zolanaj:
@Bill-Becker I believe we've addressed all comments so this is ready for your review again.  I expect to have a PR on the API side that connects to this branch for testing ready this week.

test/runtests.jl
Bill-Becker: I think we should raise an exception and not allow running the scenario if the new heating techs can't serve all of the specified heating load(s).

src/core/scenario.jl
```diff
@@ -644,11 +647,54 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
...
     electric_heater = nothing
     if haskey(d, "ElectricHeater") && d["ElectricHeater"]["max_mmbtu_per_hour"] > 0.0
         electric_heater = ElectricHeater(;dictkeys_tosymbols(d["ElectricHeater"])...)
     end
 
+    # ASHP
+    ashp = nothing
+    cop_heating = []
+    cop_cooling = []
+    if haskey(d, "ASHP") && d["ASHP"]["max_ton"] > 0.0
```
Bill-Becker, line 660: Depending how fast ASHP sizing solves, it may be useful to check if existing boiler and chiller is "retire_in_optimal" and then we may be able to determine the ASHP size in pre-processing, right?

src/results/ashp.jl
```diff
@@ -0,0 +1,126 @@
+# REopt®, Copyright (c) Alliance for Sustainable Energy, LLC. See also https://github.com/NREL/REopt.jl/blob/master/LICENSE.
+
+"""
+`ASHP` results keys:
+- `size_ton`  # Thermal production capacity size of the ASHP [ton/hr]
+- `electric_consumption_series_kw`  # Fuel consumption series [kW]
+- `annual_electric_consumption_kwh`  # Fuel consumed in a year [kWh]
+- `thermal_production_series_mmbtu_per_hour`  # Thermal energy production series [MMBtu/hr]
+- `annual_thermal_production_mmbtu`  # Thermal energy produced in a year [MMBtu]
```
Bill-Becker, line 9: Let's add in the comment string that this _mmbtu is **heating** while the _tonhour is **cooling** since we don't distinguish between heating and cooling thermal production in the name.

src/results/ashp.jl
```diff
@@ -0,0 +1,126 @@
...
+	r["thermal_to_storage_series_mmbtu_per_hour"] = round.(value.(ASHPToHotTESKW) / KWH_PER_MMBTU, digits=3)
+
+    if !isempty(p.techs.steam_turbine) && p.s.ashp.can_supply_steam_turbine
+        @expression(m, ASHPToSteamTurbine[ts in p.time_steps], sum(m[:dvThermalToSteamTurbine]["ASHP",q,ts] for q in p.heating_loads))
+        @expression(m, ASHPToSteamTurbineByQuality[q in p.heating_loads, ts in p.time_steps], m[:dvThermalToSteamTurbine]["ASHP",q,ts])
+    else
+        ASHPToSteamTurbine = zeros(length(p.time_steps))
+        @expression(m, ASHPToSteamTurbineByQuality[q in p.heating_loads, ts in p.time_steps], 0.0)
+    end
+    r["thermal_to_steamturbine_series_mmbtu_per_hour"] = round.(value.(ASHPToSteamTurbine) / KWH_PER_MMBTU, digits=3)
```
Bill-Becker, line 60: Wondering about needing the option or result of "ASHP to steamturbine" when that would never practically happen. Is there any value in have that always as zeros as opposed to not existing?

Bill-Becker (commented):
@zolanaj @atpham88 We're close, but I have some more comments. Thanks!

src/core/scenario.jl
```diff
@@ -644,11 +647,54 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
...
+                            gcr=pv.gcr, inv_eff=pv.inv_eff*100, timeframe="hourly", radius=pv.radius, time_steps_per_hour=settings.time_steps_per_hour)
+                    end
+                else
+                    # if PV is not evaluated, call PVWatts to get ambient temperature series
+                    pv_prodfactor, ambient_temp_celsius = call_pvwatts_api(site.latitude, site.longitude; time_steps_per_hour=settings.time_steps_per_hour)    
+                end
+            end
+
+            if !haskey(d["ASHP"], "cop_heating")
+                cop_heating = round.(0.083 .* ambient_temp_celsius .+ 2.8255, digits=2)
```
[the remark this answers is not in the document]
atpham88: The COPs must have been updated in the document since I coded this in. Updated here https://github.com/NREL/REopt.jl/pull/372/commits/2a77f89edac6d4560b8c7e3d7a3b66bd8e730d77

src/results/ashp.jl
atpham88: Updated here https://github.com/NREL/REopt.jl/pull/372/commits/ab4b107873012068db45e6906a5d30eb094819cc

Bill-Becker (commented):
High level comment on ASHP_WH - it seems so similar to ASHP, that I'm wondering how we can reduce code by leveraging the core and results structs/scripts for one and just using different default values for distinguishing technologies (even if we don't have unique defaults now, we may in the future). ASHP_WH doesn't do cooling, so that stuff would be zeroed.

Also, if a user just wants to do ASHP for space heating, and they want to "force it into the solution", how would we still account for the DHW loads served by the ExistingBoiler if we only have "retire_in_optimal" (no techs would be able to serve DHW).

data/ashp/ashp_wh_defaults.json
```diff
@@ -0,0 +1,11 @@
+{
+    "installed_cost_per_ton": 4050,
```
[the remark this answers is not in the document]
Bill-Becker: @atpham88 @zolanaj I think this is my last open comment, to update the defaults to be consistent with the requirements doc.

src/constraints/thermal_tech_constraints.jl
```diff
@@ -64,7 +64,7 @@ function add_heating_tech_constraints(m, p; _n="")
     # Constraint (7_heating_prod_size): Production limit based on size for non-electricity-producing heating techs
     if !isempty(setdiff(p.techs.heating, union(p.techs.elec, p.techs.ghp)))
         @constraint(m, [t in setdiff(p.techs.heating, union(p.techs.elec, p.techs.ghp)), ts in p.time_steps],
-            sum(m[Symbol("dvHeatingProduction"*_n)][t,q,ts] for q in p.heating_loads)  <= m[Symbol("dvSize"*_n)][t]
+            sum(m[Symbol("dvHeatingProduction"*_n)][t,q,ts] for q in p.heating_loads)  <= m[Symbol("dvSize"*_n)][t] * p.heating_cf[t][ts]
```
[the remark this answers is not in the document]
zolanaj: These are amended in https://github.com/NREL/REopt.jl/commit/bb7361c23870f2969eb1420d986ef8b077e6ccdf (just noting for posterity)

src/core/ashp_wh.jl
```diff
@@ -0,0 +1,132 @@
+# REopt®, Copyright (c) Alliance for Sustainable Energy, LLC. See also https://github.com/NREL/REopt.jl/blob/master/LICENSE.
+
+struct ASHP_WH <: AbstractThermalTech
```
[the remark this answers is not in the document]
zolanaj: In the refactoring we performed, we're now using the struct `ASHP` for both the `ASHP_SpaceHeater` (for which cooling is optional) and `ASHP_WaterHeater` (which only serves domestic hot water and does not cool).  The boolean attributes include process heat mostly to give future flexibility and to maintain consistency with the other heating techs.

src/results/ashp_wh.jl
```diff
@@ -0,0 +1,98 @@
...
+- `size_ton`  # Thermal production capacity size of the ASHP_WH [ton/hr]
+- `electric_consumption_series_kw`  # Fuel consumption series [kW]
+- `annual_electric_consumption_kwh`  # Fuel consumed in a year [kWh]
+- `thermal_production_series_mmbtu_per_hour`  # Thermal heating energy production series [MMBtu/hr]
+- `annual_thermal_production_mmbtu`  # Thermal heating energy produced in a year [MMBtu]
+- `thermal_to_storage_series_mmbtu_per_hour`  # Thermal power production to TES (HotThermalStorage) series [MMBtu/hr]
+- `thermal_to_steamturbine_series_mmbtu_per_hour`  # Thermal power production to SteamTurbine series [MMBtu/hr]
+- `thermal_to_load_series_mmbtu_per_hour`  # Thermal power production to serve the heating load series [MMBtu/hr]
+- `thermal_to_storage_series_ton` # Thermal production to ColdThermalStorage
+- `thermal_to_load_series_ton` # Thermal production to cooling load
```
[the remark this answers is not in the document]
zolanaj: removed in https://github.com/NREL/REopt.jl/commit/7ce8fbb324c19df8e1fa789945cde9530e4a2173

src/results/ashp_wh.jl
```diff
@@ -0,0 +1,98 @@
...
+    if "DomesticHotWater" in p.heating_loads && p.s.ashp_wh.can_serve_dhw
+        @expression(m, ASHPWHToDHWKW[ts in p.time_steps], 
+            m[:dvHeatingProduction]["ASHP_WH","DomesticHotWater",ts] - ASHPWHToHotTESByQualityKW["DomesticHotWater",ts] - ASHPWHToSteamTurbineByQuality["DomesticHotWater",ts]
+        )
+    else
+        @expression(m, ASHPWHToDHWKW[ts in p.time_steps], 0.0)
+    end
+    r["thermal_to_dhw_load_series_mmbtu_per_hour"] = round.(value.(ASHPWHToDHWKW ./ KWH_PER_MMBTU), digits=5)
+    
+    if "SpaceHeating" in p.heating_loads && p.s.ashp_wh.can_serve_space_heating
```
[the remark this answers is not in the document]
zolanaj: Just for reference, the new `force_in_system` attribute added in https://github.com/NREL/REopt.jl/commit/9e611ec3e98c720617b2c7d01164499693943979 will set all other producing techs to zero for any load served by ASHP.  This allows storage to still be considered as a useful option when there is time-variant pricing.

src/core/ashp.jl
```diff
@@ -0,0 +1,227 @@
...
+If a user provides the `ASHP_SpaceHeater` key then the optimal scenario has the option to purchase 
+this new `ASHP` to meet the heating load in addition to using the `ExistingBoiler`
+to meet the heating load. 
+
+```julia
+function ASHP_SpaceHeater(;
+    min_ton::Real = 0.0, # Minimum thermal power size
+    max_ton::Real = BIG_NUMBER, # Maximum thermal power size
+    installed_cost_per_ton::Union{Real, nothing} = nothing, # Thermal power-based cost
+    om_cost_per_ton::Union{Real, nothing} = nothing, # Thermal power-based fixed O&M cost
```
Bill-Becker, line 36: @zolanaj @atpham88 For configs 1 and 3, where we keep the existing boiler and/or chiller, we need to use the O&M number/default in the requirement doc. Configs 2 and 4 have a default of 0/ton/year because we assume ASHP has an equivalent O&M as the existing HVAC which is now retired.

atpham88: I set the default o&m to 40 and set o&m to 0 when force into system here https://github.com/NREL/REopt.jl/pull/372/commits/7297d80518c17dcab07ecea9b06682d9ae116406

data/ashp/ashp_defaults.json
```diff
@@ -0,0 +1,30 @@
...
+        "om_cost_per_ton": 40,
+        "macrs_option_years": 0,
+        "macrs_bonus_fraction": 0.0,
+        "can_supply_steam_turbine": false,
+        "can_serve_process_heat": false,
+        "can_serve_dhw": false, 
+        "can_serve_space_heating": true,
+        "can_serve_cooling": true,
+        "force_into_system": false,
+        "back_up_temp_threshold": 10.0
```
Bill-Becker, line 14: @atpham88 can you append "_degF" to this field name? That's what we do for other temperature inputs.

atpham88: done here https://github.com/NREL/REopt.jl/pull/372/commits/ce7b6826c8009a22999d30e9e203d91a176631c4

Bill-Becker (approved):
Thanks @zolanaj  and @atpham88 , approved !!!

Alex - I re-initiated your review, just to confirm you didn't plan to make any more updates. I'd like to merge this in.

zolanaj:
> Thanks @zolanaj and @atpham88 , approved !!!
> 
> Alex - I re-initiated your review, just to confirm you didn't plan to make any more updates. I'd like to merge this in.

Thanks @Bill-Becker!  No further changes planned for this PR.  I'll approve the pass-back and merge.

zolanaj (approved):

atpham88:
Awesome! Thanks so much @Bill-Becker and @zolanaj

PR #376, opened 2024-04-22
Update EASIUR years

adfarth:
- Updated `pop_year` and `income_year` used in call to EASIUR data (`get_EASIUR2005`) each to 2024, from 2020. 
- Updated usd conversion used for EASIUR health cost calcs from USD_2010_to_2020 = 1.246 to USD_2010_to_2024 = 1.432


src/core/financial.jl
```diff
@@ -388,7 +389,7 @@ function get_EASIUR2005(stack::String; pop_year::Int64=2005, income_year::Int64=
                 setindex!(ret_map, v .* adj, k)
             end
         catch
-            throw(@error("income year is $(income_year) but must be between 1990 to 2024"))
+            throw(@error("EASIUR income year is $(income_year) but must be between 1990 to 2024"))
```
hdunham, line 392: also throw an error if pop_year is outside its bounds?

hdunham (approved):

src/core/financial.jl
adfarth: Since we're manually setting these (i.e. they're not user inputs) and the pop year goes up to 2050 I think we should be good on that for a while?

hdunham: right!

Issue #377, opened 2024-04-23
Reduce test times


zolanaj:
Avoid failed test suites due to timeouts by simplifying certain tests that reduce solve times while still preserving the test of the feature

zolanaj:
Addressed in https://github.com/NREL/REopt.jl/pull/378 - closing

PR #378, opened 2024-04-24
Reduce time of testing

zolanaj:


zolanaj:
Most edits to inputs are done in `runtests.jl` but can all be migrated to the test .jsons

zolanaj:
referencing https://github.com/NREL/REopt.jl/issues/377

.github/workflows/CI.yml
```diff
@@ -13,7 +13,8 @@ jobs:
       matrix:
         julia-version: ['1.8']
         julia-arch: [x64]
-        os: [ubuntu-latest, windows-latest, macOS-11]
+        # os: [ubuntu-latest, windows-latest, macOS-11]
+        os: [windows-latest, macOS-11]
```
adfarth, line 17: @zolanaj do we definitely want to keep `ubuntu-latest` out of the test sets from here on out? Did you test with ubuntu with these latest changes?

test/runtests.jl
adfarth, line 1: @zolanaj can we remove the following uncommented line?

"# @testset "Minimize Unserved Load" is too slow with Cbc (killed after 8 hours)" (line 220)

test/runtests.jl
```diff
@@ -909,18 +909,47 @@ else  # run HiGHS tests
...
+            d["CHP"]["max_kw"] = 100.0
+            d["Generator"]["existing_kw"] = 0.0
+            d["Generator"]["min_kw"] = 100.0
+            d["Generator"]["max_kw"] = 100.0
+            d["ElectricStorage"]["min_kw"] = 20
+            d["ElectricStorage"]["max_kw"] = 20
+            d["ElectricStorage"]["min_kwh"] = 50
+            d["ElectricStorage"]["max_kwh"] = 50
+            d["Financial"]["microgrid_upgrade_cost_fraction"] = 0.0
+            s = Scenario(d)
```
adfarth, line 930: @zolanaj just curious, was there a compute time advantage to creating the Scenario and REoptInputs first, rather than just running `run_reopt(m,d)` ?

test/runtests.jl
```diff
@@ -909,18 +909,47 @@ else  # run HiGHS tests
...
+            d["CHP"]["min_kw"] = 100.0
+            d["CHP"]["max_kw"] = 100.0
+            d["Generator"]["existing_kw"] = 0.0
+            d["Generator"]["min_kw"] = 100.0
+            d["Generator"]["max_kw"] = 100.0
+            d["ElectricStorage"]["min_kw"] = 20
+            d["ElectricStorage"]["max_kw"] = 20
+            d["ElectricStorage"]["min_kwh"] = 50
+            d["ElectricStorage"]["max_kwh"] = 50
+            d["Financial"]["microgrid_upgrade_cost_fraction"] = 0.0
```
adfarth, line 929: @zolanaj the original test did have a `microgrid_upgrade_cost_fraction` in the post and I'm wondering if that's a good thing to test. Would including this cost slow down this test significantly?

test/runtests.jl
```diff
@@ -909,18 +909,47 @@ else  # run HiGHS tests
...
+            # Increase cost of microgrid upgrade and PV Size, PV not used and some load not met
+            d["Financial"]["microgrid_upgrade_cost_fraction"] = 0.3
+            d["PV"]["min_kw"] = 200.0
+            d["PV"]["max_kw"] = 200.0
+            s = Scenario(d)
+            p = REoptInputs(s)
+            m = Model(optimizer_with_attributes(HiGHS.Optimizer, "output_flag" => false, "log_to_console" => false, "mip_rel_gap" => 0.01, "presolve" => "on"))
+            results = run_reopt(m, p)
+            @test value(m[:binMGTechUsed]["PV"]) ≈ 0
+            @test sum(results["Outages"]["unserved_load_per_outage_kwh"]) > 0
```
adfarth, line 951: @zolanaj any reason to not set this to the value of 24.16 instead of just > 0 ?

test/scenarios/re_emissions_elec_only.json
```diff
@@ -48,7 +48,7 @@
     "production_incentive_max_benefit": 0.0,
     "state_ibi_fraction": 0.0,
     "electric_efficiency_full_load": 0.349006,
-    "electric_efficiency_half_load": 0.326729,
```
adfarth, line 51: @zolanaj just curious why you made this change?

adfarth (approved):
@zolanaj this looks good and it's great to have the tests passing in ~an hour rather than 3!! I am approving but also have a question and suggestion (and a few more minor qs in the comments): 
- Should we definitely remove ubuntu from from the test os's? 
- I do think it would make sense to pull your changes into the post jsons, as you've suggested

test/runtests.jl
zolanaj: I don't think there's any difference in computing - run_reopt(m, d) does the same sequence of commands.  I've always used the Scenario and REoptInputs creation upfront for debugging purposes; it's fewer lines of code in the testing though so I can convert it.

test/runtests.jl
zolanaj: I think this one is stable enough to make 24.16.  I usually stay away from fixed values when a 1% mip tolerance is involved, but I think this one should be ok fo now.  Good catch!

.github/workflows/CI.yml
zolanaj: This is temporary, I believe we will bring it back soon.  @Bill-Becker may have more to say on this.

test/scenarios/re_emissions_elec_only.json
zolanaj: by making the part-load and full-load efficiency the same, we no longer have fuel burn tied to the binary variables indicating that the technology is running in a given hour.  This effectively make the fuel burn constraints linear.

test/runtests.jl
zolanaj: done in 50c8d4b9f56d1a63b9936353e7875e95f96ccc30.  Thanks for catching!

test/runtests.jl
zolanaj: done in 674780496cf626717396860c4257af90ada54ffd for the updated tests only; we can probably do this more widely but I don't see any reason for a computational difference so that's low priority.

test/runtests.jl
zolanaj: done in 296bbd63acc4755283536de6c70c88e191ad54f7

zolanaj:
> @zolanaj this looks good and it's great to have the tests passing in ~an hour rather than 3!! I am approving but also have a question and suggestion (and a few more minor qs in the comments):
> 
> * Should we definitely remove ubuntu from from the test os's?
> * I do think it would make sense to pull your changes into the post jsons, as you've suggested

Thank you for all of this @adfarth !  I have an issue [here](https://github.com/NREL/REopt.jl/issues/380) for the Ubuntu tests and implemented the suggested edits with links to specific commits in the replies.  Once tests are confirmed to pass, I'll merge into develop.

PR #390, opened 2024-05-07
3 tiered TOU demand bugs

hdunham:
### Fixed
- Fixed a reshape call in function `parse_urdb_tou_demand` that incorrectly assumed row major instead of column major ordering
- Fixed a loop range in function `parse_urdb_tou_demand` that incorrectly started at 0 instead of 1
- Added the missing tier index when accessing `p.s.electric_tariff.tou_demand_rates` in function `add_elec_utility_expressions`
### Added
- Test for tiered TOU demand rates in `test/runtests.jl`


src/constraints/electric_utility_constraints.jl
```diff
@@ -384,7 +384,7 @@ function add_elec_utility_expressions(m, p; _n="")
 
     if !isempty(p.s.electric_tariff.tou_demand_rates)
         m[Symbol("DemandTOUCharges"*_n)] = @expression(m, 
-            p.pwf_e * sum( p.s.electric_tariff.tou_demand_rates[r] * m[Symbol("dvPeakDemandTOU"*_n)][r, tier]
```
zolanaj, line 387: This explains why one tier still worked and multiple tiers broke - julia treated the single-value array as a scalar.

zolanaj (approved):
great catches @hdunham, thank you for this!  I just merged the recent changes to develop and updated the changelog to merge with the rest of v. 0.46.0, I'll merge as soon as passing tests are confirmed and register the new version.

PR #391, opened 2024-05-09
Fix ghp-to-storage dispatch

zolanaj:
- Prevents a bug in which the model fails to build when both GHP and thermal storage are present
- While decision variables are present for GHP to charge to storage, these are fixed to zero, i.e., GHP cannot charge storage (which is consistent with the original formulation for GHP; it only served load previously)

### DIFF
--- a/src/constraints/storage_constraints.jl
+++ b/src/constraints/storage_constraints.jl
@@ -181,6 +181,19 @@ function add_hot_thermal_storage_dispatch_constraints(m, p, b; _n="")
         sum(m[Symbol("dvHeatFromStorage"*_n)][b,q,ts] for q in p.heating_loads)
     )
 
+    #Do not allow GHP to charge storage
+    if !isempty(p.techs.ghp)
+        for b in p.s.storage.types.hot
+            for t in p.techs.ghp
+                for q in p.heating_loads
+                    for ts in p.time_steps
+                        fix(m[Symbol("dvHeatToStorage"*_n)][b,t,q,ts], 0.0, force=true)
+                    end
+               end
+            end
+        end
+    end
+
 end
 
 function add_cold_thermal_storage_dispatch_constraints(m, p, b; _n="")
@@ -207,6 +220,17 @@ function add_cold_thermal_storage_dispatch_constraints(m, p, b; _n="")
         m[Symbol("dvStoragePower"*_n)][b] >= m[Symbol("dvDischargeFromStorage"*_n)][b,ts] + 
         sum(m[Symbol("dvProductionToStorage"*_n)][b,t,ts] for t in p.techs.cooling)
     )
+
+    #Do not allow GHP to charge storage
+    if !isempty(p.techs.ghp)
+        for b in p.s.storage.types.cold
+            for t in p.techs.ghp
+                for ts in p.time_steps
+                    fix(m[Symbol("dvProductionToStorage"*_n)][b,t,ts], 0.0, force=true)
+                end
+            end
+        end
+    end
 end
 
 function add_storage_sum_constraints(m, p; _n="")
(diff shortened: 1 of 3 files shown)


src/core/reopt.jl
```diff
@@ -584,7 +584,7 @@ function add_variables!(m::JuMP.AbstractModel, p::REoptInputs)
 		dvGridPurchase[p.time_steps, 1:p.s.electric_tariff.n_energy_tiers] >= 0  # Power from grid dispatched to meet electrical load [kW]
 		dvRatedProduction[p.techs.all, p.time_steps] >= 0  # Rated production of technology t [kW]
 		dvCurtail[p.techs.all, p.time_steps] >= 0  # [kW]
-		dvProductionToStorage[p.s.storage.types.all, p.techs.all, p.time_steps] >= 0  # Power from technology t used to charge storage system b [kW]
+		dvProductionToStorage[p.s.storage.types.all, union(p.techs.ghp,p.techs.all), p.time_steps] >= 0  # Power from technology t used to charge storage system b [kW]
```
adfarth, line 587: Since gph is not in techs.all, are there other places this causes an issue? I'm looking through the code and wondering about: 

- The code nested under reopt.jl, `if !isempty(p.techs.all)` line 269. E.g. `add_ghp_constraints(m, p)` is nested under this. So does this mean GHP has to be evaluated with another tech? 
- And then just confirming that `dvSize`, `dvPurchaseSize`, `dvRatedProduction`, and `dvCurtail` don't need to also be indexed on GHP?

src/constraints/storage_constraints.jl
```diff
@@ -207,6 +220,17 @@ function add_cold_thermal_storage_dispatch_constraints(m, p, b; _n="")
         m[Symbol("dvStoragePower"*_n)][b] >= m[Symbol("dvDischargeFromStorage"*_n)][b,ts] + 
         sum(m[Symbol("dvProductionToStorage"*_n)][b,t,ts] for t in p.techs.cooling)
     )
+
+    #Do not allow GHP to charge storage
+    if !isempty(p.techs.ghp)
+        for b in p.s.storage.types.cold
```
adfarth, line 226: Could this be written as 

`for b in p.s.storage.types.cold, t in p.techs.ghp, ts in p.time_steps` ?

src/constraints/storage_constraints.jl
```diff
@@ -207,6 +220,17 @@ function add_cold_thermal_storage_dispatch_constraints(m, p, b; _n="")
         m[Symbol("dvStoragePower"*_n)][b] >= m[Symbol("dvDischargeFromStorage"*_n)][b,ts] + 
         sum(m[Symbol("dvProductionToStorage"*_n)][b,t,ts] for t in p.techs.cooling)
     )
+
+    #Do not allow GHP to charge storage
```
adfarth, line 224: @zolanaj  or @Bill-Becker is this assumption documented anywhere? If not, is it important to do so at some point?

adfarth:
Adding a note here to double check that `m[:dvHeatToStorage]` is indexed on `[b,t,q,ts]` everywhere (missing `t` in a few places)

src/constraints/storage_constraints.jl
zolanaj: I am raising an issue [here](https://github.com/NREL/REopt.jl/issues/392) as a placeholder to allow for further discussion.

src/constraints/storage_constraints.jl
zolanaj: yes, that's a better way!  I implemented it in 2228b394f823e8c8ce2f76c3ef52477054bac302

zolanaj:
> Adding a note here to double check that `m[:dvHeatToStorage]` is indexed on `[b,t,q,ts]` everywhere (missing `t` in a few places)

I caught a couple others in addition to the ones you showed, and fixed them in 428379489e9a007cf7e4c32c593c231a2135cfce

src/core/reopt.jl
zolanaj: Right now, I believe the answer is yes - a GHP-only scenario isn't likely to produce the result we want.  I raised issue https://github.com/NREL/REopt.jl/issues/393 to address this.

GHP does not produce electricity and the capital cost is attached to binary variables elsewhere, and GHP does not dispatch according to system sizing, so for now the four variables mentioned above do not need to be indexed on GHP.

adfarth (approved):
Thanks for addressing all the comments and fixing this bug!

PR #397, opened 2024-05-17
Chp default bug

hdunham:
### Fixed
- Convert `max_electric_load_kw` to its expected type of _Float64_ (can be an _Int_) before passing to function `get_chp_defaults_prime_mover_size_class`


zolanaj (approved):
Thanks for this @hdunham!  Approved with one question: Do you know if there are other potential issues where we may need to convert other than this? I think running conversions on all defaults inputs may be overkill, but I'm just wondering if there's something to watch for if we have maximums of vectors in future defaults population (or something similar).

PR #398, opened 2024-05-20
Error when non standard value for "units" field in "energyratestructure" of URDB response

hdunham:
### Changed
- When the URDB response `energyratestructure` has a "unit" value that is not "kWh", throw an error instead of averaging rates in each energy tier.


zolanaj (approved):
Looks good, thank you @hdunham!  I added a test to ensure the exotic units are passed up to results. Assuming test pass (the last one was due to a timeout for PVWatts so might just need to be refreshed), this can be merged.

PR #399, opened 2024-05-20
Add third party factor to PV O&M results calculation

lixiangk1:
Added third party factor to the PV O&M results calculation to match what is done for all other technologies.


adfarth (approved):
Looks good, thanks for catching this!

Issue #386, opened 2024-05-03
Increase maximum wholesale benefit


zolanaj:
Citing a previous observation by @adfarth: 

https://reopt.nrel.gov/tool/results/cb9b426f-242a-40d0-965a-09ea843e4105

In this scenario, the WHL is extremely high AND a NEM limit is set. However, REopt still sizes the system under the the NEM limit kW, but compensates the exports at the WHL rate. I had expected REopt to just size the system infinitely large, as it would be more profitable to blow past the NEM limit. 

The underlying cause is a relatively small maximum wholesale benefit in place [here.](https://github.com/NREL/REopt.jl/blob/1199d1c007461c93b64887e2ef7634086f8dd9f3/src/constraints/electric_utility_constraints.jl#L125)  It may be worth increasing this so that the optimal size is much than the current limit; for now the proposed fix is that a user is warned if the wholesale limit is met (see https://github.com/NREL/REopt.jl/pull/383).

zolanaj:
Closing - addressed in https://github.com/NREL/REopt.jl/pull/401.

PR #401, opened 2024-05-22
bug fixes for max net metering benefit, multi-period and multi-tier demand charges

zolanaj:
### Fixed
- Increased the big-M bound on maximum net metering benefit to prevent artificially low export benefits
- Fixed a bug in which tier limits did not load correctly when the number of tiers vary by period in the inputs


zolanaj:
addresses https://github.com/NREL/REopt.jl/issues/386

src/constraints/electric_utility_constraints.jl
```diff
@@ -52,7 +52,7 @@ function add_export_constraints(m, p; _n="")
             @warn "Adding binary variable for net metering choice. Some solvers are slow with binaries."
 
             # Good to bound the benefit - we use max_bene as a lower bound because the benefit is treated as a negative cost
-            max_bene = sum([ld*rate for (ld,rate) in zip(p.s.electric_load.loads_kw, p.s.electric_tariff.export_rates[:NEM])])*10
+            max_bene = sum([ld*rate for (ld,rate) in zip(p.s.electric_load.loads_kw, p.s.electric_tariff.export_rates[:NEM])])*p.pwf_e*p.hours_per_time_step*10
```
adfarth, line 55: Just curious why `*p.hours_per_time_step` helps here?

adfarth:
Checked and confirmed that these fixes address previous issues in results from these posts: 
[high_wholesale.json](https://github.com/NREL/REopt.jl/files/15439756/high_wholesale.json)
[high_energy_rate_max.json](https://github.com/NREL/REopt.jl/files/15439763/high_energy_rate_max.json)

CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ Classify the change according to the following categories:
 ## Develop
 ### Fixed
 - Convert `max_electric_load_kw` to _Float64_ before passing to function `get_chp_defaults_prime_mover_size_class`
+- Increased the big-M bound on maximum net metering benefit to prevent artificially low export benefits
```
adfarth, line 29: Is this missing the fix to the max energy and demand tiers?

src/constraints/electric_utility_constraints.jl
adfarth: Alex, this bound on NEM is less important than the WHL bound, because the NEM benefit also gets constrained by net metering size limit and the constraint that the system cannot produce > 100% of the site's load, right? So, wouldn't the NEM benefit never be > `[ld*rate for (ld,rate) in zip(p.s.electric_load.loads_kw, p.s.electric_tariff.export_rates[:NEM])]` ?

adfarth: Oh! But this limit is also used for the EXC bin, so if someone supplied a very high `export_rate_beyond_net_metering_limit` this may become active. 

Let me know if this thinking is off.

CHANGELOG.md
zolanaj: yes, thanks for catching!  this is amended in https://github.com/NREL/REopt.jl/commit/209ba633c5601b4fdb0d417e9019e6616ebb42b8

src/core/urdb.jl
```diff
@@ -229,7 +229,7 @@ function parse_urdb_energy_costs(d::Dict, year::Int; time_steps_per_hour=1, bigM
         energy_tier_max = get(energy_tier, "max", bigM)
 
         if "rate" in keys(energy_tier) || "adj" in keys(energy_tier)  || "sell" in keys(energy_tier)
-            append!(energy_tier_limits_kwh, energy_tier_max)
+            append!(energy_tier_limits_kwh, min(energy_tier_max, bigM))
```
adfarth, line 232: Do you think we should consider increasing the bigM default in `parse_urdb_energy_costs(d::Dict, year::Int; time_steps_per_hour=1, bigM = 1.0e8)`? Something like a big steel plant might use more than 1.0e8 kWh per year

src/constraints/electric_utility_constraints.jl
zolanaj: Thats correct on the wholesale and excess, you still need a large enough limit for the excess in particular.  Also, the net metering is bounded elsewhere but that bound was too low due to the present worth factor being >10.  Including `p.hours_per_time_step` and, more importantly, the annuity factor `p.pwf_e` just makes the big-M reconcile with the other components of the benefit calculation: https://github.com/NREL/REopt.jl/blob/209ba633c5601b4fdb0d417e9019e6616ebb42b8/src/constraints/electric_utility_constraints.jl#L89-L91

src/core/urdb.jl
```diff
@@ -313,8 +313,8 @@ Parse monthly ("flat") and TOU demand rates
 """
 function parse_demand_rates(d::Dict, year::Int; bigM=1.0e8, time_steps_per_hour::Int)
```
adfarth, line 314: We could remove `bigM` from these function inputs, right?

src/constraints/electric_utility_constraints.jl
adfarth: Thanks for clarifying!

adfarth:
@zolanaj Done reviewing with just two remaining comments!

src/core/urdb.jl
zolanaj: Good call!  Noting here that after a quick sidebar with you I've increased it to 1e10 for now.  Longer term, we should try to get the loads as input so that we can get more accurate big-M's (we can use the max monthly energy demand for example) - I'll add an issue to that effect.

zolanaj: > Do you think we should consider increasing the bigM default in `parse_urdb_energy_costs(d::Dict, year::Int; time_steps_per_hour=1, bigM = 1.0e8)` and `parse_urdb_demand_tiers`? Something like a big steel plant might use more than 1.0e8 kWh per year

addressed here: https://github.com/NREL/REopt.jl/commit/0c5207fe1c65d671056078bd6e2c8ce890df41c3

src/core/urdb.jl
zolanaj: I'm using them now here: https://github.com/NREL/REopt.jl/commit/3d623f007ed7390c8303195f67f362499564c214

I think we will want this to stay because the bigM's are still going to be necessary when we have loads (they'll just be customized bigM's for a given instance rather than a fixed large number).

zolanaj:
Tagging issue https://github.com/NREL/REopt.jl/issues/297 ads this may be addressed now?

Also added issue https://github.com/NREL/REopt.jl/issues/297 for future bigM improvement.

src/core/urdb.jl
adfarth: Looks good! Just documenting that I should have said "per month" and that this change only applies to the energy tiers bigM, not the demand tiers.

adfarth:
> Tagging issue #297 ads this may be addressed now?
> 
> Also added issue #297 for future bigM improvement.

Regarding #297 , documenting that currently the below (a) results in `tier_limits: [100.0, 1.0e8]` and (b) results in `tier_limits: [1.0e8, 1.0e8]`

(a) 
```
"demandratestructure": [
                [
                    {
                        "max": 100000000000000000000000000000000000000,
                        "rate": 15
                    }
                ],
                [
                    {
                        "max": 100,
                        "rate": 25
                    },
                    {
                        "max": 100000000000000000000000000000000000000,
                        "rate": 35
                    }
                ]
            ],
```

(b)
```
[
                    {
                        "max": 100,
                        "rate": 25
                    },
                    {
                        "max": 100000000000000000000000000000000000000,
                        "rate": 35
                    }
                ],
                [
                    {
                        "max": 100000000000000000000000000000000000000,
                        "rate": 15
                    }
                ]
```

zolanaj:
@adfarth Ready for you to check again!  There are more tests in the new set and the limits are now indexed on month (or ratchet).

src/core/urdb.jl
```diff
@@ -12,15 +12,15 @@ Contains some of the data for ElectricTariff
 """
 struct URDBrate
     energy_rates::Array{Float64,2}  # time X tier
-    energy_tier_limits::Array{Real,1}
+    energy_tier_limits::Array{Float64,2}
```
adfarth, line 15: @zolanaj could you add a comment explaining the dimensions as there is for `energy_rates` here and under `struct ElectricTariff`?

src/core/urdb.jl
```diff
@@ -196,51 +195,34 @@ end
...
     energy_cost_vector = Float64[]
     sell_vector = Float64[]
+    energy_tier_limits_kwh = Array{Float64}(undef, 12, n_energy_tiers)
 
     for tier in range(1, stop=n_energy_tiers)
 
         for month in range(1, stop=12)
+            period = Int(d["energyweekdayschedule"][month][1] + 1)
+            for tier in range(1, stop=n_energy_tiers)
+                # tiered energy schedules are assumed to be consistent for each month (i.e., the first hour can represent all 24 hours of the schedule).
```
adfarth, line 221: "the first hour can represent all 24 hours of the schedule" indicates to me that we couldn't model a TOU rate that varies by hour of day. Is that what this comment is intending to say?

src/core/urdb.jl
```diff
@@ -359,45 +342,27 @@ end
 
 
 """
-    parse_urdb_demand_tiers(A::Array; bigM=1.0e8)
+get_num_tiers(d::Dict)
 
-set up and validate demand tiers
-    returns demand_tiers::Array{Float64, n_tiers}
+    get maximum number of demand tiers in any period from scrubbed demand rate structure
```
adfarth, line 347: This function is used for energy rates as well but the description says just demand

CHANGELOG.md
```diff
@@ -23,6 +23,14 @@ Classify the change according to the following categories:
     ### Deprecated
     ### Removed
 
+## Develop - 2024-06-05
+### Fixed
+- Increased the big-M bound on maximum net metering benefit to prevent artificially low export benefits
+- Fixed a bug in which tier limits did not load correctly when the number of tiers vary by period in the inputs
+- Set a limit for demand and energy tier maxes to avoid errors returned by HiGHS due to numerical limits
+- Index utility rate demand and energy tier limits on month in addition to tier
```
adfarth, line 31: @zolanaj  I pulled the changes for this PR out under a new heading and added the last bullet. Could you please check that the added bullet is accurate, and maybe add a note about how this changes our assumptions compared to the previous behavior?

test/runtests.jl
adfarth, line 1: @zolanaj I would have expected that your change might have caused NPVs to change and a need to correct more expected values in runtests. Are you also surprised this isn't the case?

adfarth:
@zolanaj Adding one more thing that we can definitely punt to another PR unless you think it makes sense to address here: 

It looks like the `["ElectricUtility"]["interconnection_limit_kw"]` is currently **not** getting applied to wholesale techs (only to net metering techs). I confirmed this by (looking at the constraints and) running a scenario with a wholesale rate and an interconnection limit. REopt sizes over the interconnection limit but still receives WHL compensation for exports.

src/core/urdb.jl
zolanaj: > So this means that if you have a rate with multiple periods within a month, and the tier maxes (kWh) vary between those periods, REopt will overwrite the tier max values for all periods that are not the hour 1 period, using hour 1 period maxes.
> 
> For example, period 1 here would be assumed to have a 1st tier max of 88 in May-Sept, and a 1st tier max of 100 in Nov (given the resulting energy_tier_limits matrix shown below); is that a correct interpretation? `energy_tier_limits_kwh: [88.0 1.0e10; 88.0 1.0e10; 88.0 1.0e10; 88.0 1.0e10; 88.0 1.0e10; 88.0 1.0e10; 88.0 1.0e10; 88.0 1.0e10; 88.0 1.0e10; 88.0 1.0e10; 100.0 1.0e10; 88.0 1.0e10]` image
> 
> Is it accurate to say this is still not fully capturing the intended rate, but is one step better than what we did before, which was assume that all periods[tiers] had the same max limits?

That's correct if we are talking about monthly rates and limits, but the time-of-use "ratchets" allow for custom definition of periods and so this setup likely assigns a separate "ratchet" that consists of thirty hourly periods from 12-1am and another for all the other hours in November, each of which has its own separate tiers and limits.  I'd expect to see 18 time-of-use periods in this setup, each of which has two tiers.

So, to that end: is the above matrix what's actually produced when you insert this custom output or do you get the 2x18 matrix?  One thing I did not touch as part of this PR is the logic for when custom time-of-use vs. monthly tiers are generated.  We can revisit that, but it's likely best reserved for an issue at this point.

zolanaj:
> It looks like the `["ElectricUtility"]["interconnection_limit_kw"]` is currently **not** getting applied to wholesale techs (only to net metering techs). I confirmed this by (looking at the constraints and) running a scenario with a wholesale rate and an interconnection limit. REopt sizes over the interconnection limit but still receives WHL compensation for exports.

Great catch!  I'm adding an issue here: https://github.com/NREL/REopt.jl/issues/416

test/runtests.jl
zolanaj: > @zolanaj I would have expected that your change might have caused NPVs to change and a need to correct more expected values in runtests. Are you also surprised this isn't the case?

I'm not surprised for the following two reasons: 
- The net-metering bug that we caught did indeed change test results in that suite; and, 
- The other tests that I've seen thus far in runtests.jl do not have net metering or multi-tiered demand with various maxes, at least in the instances I've taken a look at in the past and a couple spot-checked tests I've done.

src/core/urdb.jl
```diff
@@ -300,23 +285,21 @@ Parse monthly ("flat") and TOU demand rates
 """
```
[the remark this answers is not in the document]
zolanaj: > Should the note "can modify URDB dict when there is inconsistent numbers of tiers in rate structures" be removed or modified?

I agree as it's part of the scrubbing function, not this one. It's removed here:  https://github.com/NREL/REopt.jl/commit/226cd67b36b972559a16af48982849c0cbfaccd6

src/core/urdb.jl
zolanaj: Yes it should; this snippet of code is actually redundant because it all occurs in the scrubbing function to make the number of tiers equal across periods.  I refactored the code so we don't do the same thing twice here:  https://github.com/NREL/REopt.jl/commit/5fbc355766cf2dad01394c4a16c7006852a68d19

src/core/urdb.jl
zolanaj: > @zolanaj could you add a comment explaining the dimensions (as there is for `energy_rates`) here and under `struct ElectricTariff`?

addressed here: https://github.com/NREL/REopt.jl/commit/f1301b2537993b2230d17af0ea71ca0a02b73166

CHANGELOG.md
zolanaj: Thanks for checking the documentation, especially as the PR evolved.  I made a small edit and compared to previous behavior here: https://github.com/NREL/REopt.jl/pull/401/commits/ccfcb512169f7927f48294416176dddd81229e3a

zolanaj:
> @zolanaj Approving with just a few comments requesting updates to help text and a few questions to help me understand the changes. Thanks for making these improvements and fixes!

Thanks for the thorough review @adfarth! after making some edits in response to this last round I will merge once tests are confirmed to pass.

src/core/urdb.jl
zolanaj: piggybacking on this comment: what I said above applies to demand charges but **not** energy rates, so this is a missing feature as compared to everything that the URDB rate builder allows for.

PR #406, opened 2024-05-23
LDES

hdunham:


src/constraints/storage_constraints.jl
```diff
@@ -43,7 +43,8 @@ function add_general_storage_dispatch_constraints(m, p, b; _n="")
     #Constraint (4j): Dispatch from storage is no greater than power capacity
 	@constraint(m, [ts in p.time_steps],
         m[Symbol("dvStoragePower"*_n)][b] >= m[Symbol("dvDischargeFromStorage"*_n)][b, ts]
-    )
+        + m[Symbol("dvStorageToGrid")][ts] # TODO: add "b" index to this decision variable
```
hdunham, line 46: Either dvStorageToGrid should be included within dvDischargeFromStorage and subtracted out in the load balance equations, or if it's not included like you've done here then dvDischargeFromStorage needs to be renamed to dvStorageToLoad.

src/constraints/storage_constraints.jl
```diff
@@ -77,8 +80,13 @@ function add_elec_storage_dispatch_constraints(m, p, b; _n="")
 	@constraint(m, [ts in p.time_steps_with_grid],
         m[Symbol("dvStoragePower"*_n)][b] >= m[Symbol("dvDischargeFromStorage"*_n)][b, ts] + 
             sum(m[Symbol("dvProductionToStorage"*_n)][b, t, ts] for t in p.techs.elec) + m[Symbol("dvGridToStorage"*_n)][b, ts]
+            + m[Symbol("dvStorageToGrid")][ts]
     )
 
+    #Dispatch from electrical storage is no greater than power capacity
```
hdunham, line 86: why was this added? doesn't seem to do anything beyond what (4l)-alt already accomplishes since it's over time steps without grid

src/constraints/storage_constraints.jl
```diff
@@ -99,6 +107,32 @@ function add_elec_storage_dispatch_constraints(m, p, b; _n="")
             sum(m[Symbol("dvStorageEnergy"*_n)][b])
         )
     end
+
+    if p.s.storage.attr[b].is_ldes
```
hdunham, line 111: I'd suggest that we make the nomenclature more generalized because users could want to consider non LDES storage techs with fixed duration. We could have a bool input called duration_is_fixed, but actually I don't think we really need this to be an input at all. We can just have the duration input default to nothing and a non-exposed struct field duration_is_fixed that is set to true if the user provides a value for duration. I can make this change if it sounds alright with y'all.

src/constraints/storage_constraints.jl
```diff
@@ -99,6 +107,32 @@ function add_elec_storage_dispatch_constraints(m, p, b; _n="")
...
+
+    if p.s.storage.attr[b].require_start_and_end_charge_to_be_equivalent
+        # @constraint(m, m[:dvStoredEnergy]["ElectricStorage",maximum(p.time_steps)] == p.s.storage.attr[b].soc_init_fraction * m[Symbol("dvStorageEnergy"*_n)][b] )
+        print("\nSetting init SOC = final SOC\n")
+        @constraint(m,
+            m[Symbol("dvStoredEnergy"*_n)][b, 0] == m[:dvStoredEnergy]["ElectricStorage", maximum(p.time_steps)]
+        )
+    end
+
+    # Prevent charging and discharging of the battery at the same time
```
hdunham, line 123: would to be good to make this an input option (like ElectricUtility.allow_simultaneous_export_import) so can avoid the 8760 bin if want/need to.

src/core/reopt.jl
```diff
@@ -213,8 +213,12 @@ function build_reopt!(m::JuMP.AbstractModel, p::REoptInputs)
 				@constraint(m, [ts in p.time_steps], m[:dvGridToStorage][b, ts] == 0)
 				@constraint(m, [t in p.techs.elec, ts in p.time_steps_with_grid],
 						m[:dvProductionToStorage][b, t, ts] == 0)
+				@constraint(m, [ts in p.time_steps], m[Symbol("dvStorageToGrid")][ts] == 0)  # if there isn't a battery, then the battery can't export power to the grid
```
hdunham, line 216: do we need this? dvStorageToGrid should have b index so if there's no storage than the dv would have dimension of 0. Even without b index, isn't it already constrained to the storage size which would be 0 if no storage?

src/core/reopt.jl
```diff
@@ -363,9 +367,11 @@ function build_reopt!(m::JuMP.AbstractModel, p::REoptInputs)
 		sum( p.s.storage.attr[b].net_present_cost_per_kwh * m[:dvStorageEnergy][b] for b in p.s.storage.types.all )
 	))
 	
-	@expression(m, TotalPerUnitSizeOMCosts, p.third_party_factor * p.pwf_om *
-		sum( p.om_cost_per_kw[t] * m[:dvSize][t] for t in p.techs.all )
-	)
+	@expression(m, TotalPerUnitSizeOMCosts, p.third_party_factor * p.pwf_om * (
+		sum(p.om_cost_per_kw[t] * m[:dvSize][t] for t in p.techs.all) + 
+		sum(p.s.storage.attr[b].om_cost_per_kw * m[:dvStoragePower][b] for b in p.s.storage.types.elec) +
```
hdunham, line 372: why is O&M only being added for electric storage when the inputs for it were added to all storage models?

src/core/reopt.jl
```diff
@@ -565,8 +571,11 @@ function add_variables!(m::JuMP.AbstractModel, p::REoptInputs)
 		dvPeakDemandMonth[p.months, 1:p.s.electric_tariff.n_monthly_demand_tiers] >= 0  # Peak electrical power demand during month m [kW]
 		MinChargeAdder >= 0
         binGHP[p.ghp_options], Bin  # Can be <= 1 if require_ghp_purchase=0, and is ==1 if require_ghp_purchase=1
+		dvStorageToGrid[p.time_steps] >= 0 # TODO, add: "p.StorageSalesTiers" as well? export of energy from storage to the grid
```
hdunham, line 574: shall I work on adding b index?

src/results/pv.jl
```diff
@@ -66,6 +66,8 @@ function add_pv_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _n="")
 		PVPerUnitSizeOMCosts = p.om_cost_per_kw[t] * p.pwf_om * m[Symbol("dvSize"*_n)][t]
 		r["lifecycle_om_cost_after_tax"] = round(value(PVPerUnitSizeOMCosts) * (1 - p.s.financial.owner_tax_rate_fraction), digits=0)
         r["lcoe_per_kwh"] = calculate_lcoe(p, r, get_pv_by_name(t, p.s.pvs))
+        r["year_one_om_cost_before_tax"] = round(value(PVPerUnitSizeOMCosts) / p.pwf_om, digits=0)
```
hdunham, line 69: why does PVPerUnitSizeOMCosts not include third party factor but it does for storage and wind?

src/core/reopt.jl
```diff
@@ -213,8 +213,12 @@ function build_reopt!(m::JuMP.AbstractModel, p::REoptInputs)
 				@constraint(m, [ts in p.time_steps], m[:dvGridToStorage][b, ts] == 0)
 				@constraint(m, [t in p.techs.elec, ts in p.time_steps_with_grid],
 						m[:dvProductionToStorage][b, t, ts] == 0)
+				@constraint(m, [ts in p.time_steps], m[Symbol("dvStorageToGrid")][ts] == 0)  # if there isn't a battery, then the battery can't export power to the grid
 			end
 		else
+			# Additional variables for exporting storage energy to the grid
+			m[:binBattCharging] = @variable(m, p.time_steps], base_name=dv, Bin) # Binary for battery charging (vs discharging)
```
hdunham, line 220: is there a reason you didn't put this where the other battery vars are defined?

test/runtests.jl
```diff
@@ -520,6 +520,38 @@ else  # run HiGHS tests
...
+        @test results2["ElectricStorage"]["year_one_om_cost_before_tax"] ≈ 800 atol=1
+        @test results2["ElectricStorage"]["lifecycle_om_cost_after_tax"] ≈ 12543 atol=1 
+        @test results2["Financial"]["lcc"] ≈ results1["Financial"]["lcc"] + 12543 - 6272 atol=1
+        
+        data["ElectricStorage"]["self_discharge_fraction_per_timestep"] = 0.0025/24
+        model = Model(optimizer_with_attributes(HiGHS.Optimizer, "output_flag" => false, "log_to_console" => false))
+        inputs = REoptInputs(data)
+        results3 = run_reopt(model, inputs)
+        @test sum(results2["ElectricStorage"]["storage_to_load_series_kw"]) - 
+            sum(results3["ElectricStorage"]["storage_to_load_series_kw"]) ≈ 15.382 atol=0.01
```
hdunham, line 552: how was this 15 number calculated? put calc in test

hdunham:
I figured out the reason some tests are resulting in DUAL_INFEASIBLE or INFEASIBLE_OR_UNBOUNDED. When ElectricStorage maxes are 0, dvStorageToGrid is created but the battery constraints are not added so dvStorageToGrid is not constrained. Because dvStorageToGrid is a value stream in the objective, this makes the problem unbounded. On reopt.jl line 216, you fix dvStorageToGrid to 0. However, this line will not be reached because ElectricStorage has a max_kw and max_kwh of 0 so it is not even added to p.s.storage.types.all (see storage.jl line 47). Actually all the code from reopt.jl line 207 to 218 seems to be unnecessary for that same reason. But the other variables being fixed in this unreachable code block (e.g. dvDischargeFromStorage) do not cause issues like dvStorageToGrid does because they have b as an index, so if the storage constraints on them aren't added, they will also have a dimension of 0. Adding a b index to dvStorageToGrid should fix the unbounded issue.

src/constraints/storage_constraints.jl
hdunham: done

src/core/reopt.jl
hdunham: done

src/core/reopt.jl
hdunham: this also should only be defined if there for elements in storage.types.elec not storage.types.all

hdunham: Well, if it has "batt" in its name and is only used in the elec storage constraints, it should. Or, the it should be generalized to all storage.

hdunham:
> I figured out the reason some tests are resulting in DUAL_INFEASIBLE or INFEASIBLE_OR_UNBOUNDED. When ElectricStorage maxes are 0, dvStorageToGrid is created but the battery constraints are not added so dvStorageToGrid is not constrained. Because dvStorageToGrid is a value stream in the objective, this makes the problem unbounded. On reopt.jl line 216, you fix dvStorageToGrid to 0. However, this line will not be reached because ElectricStorage has a max_kw and max_kwh of 0 so it is not even added to p.s.storage.types.all (see storage.jl line 47). Actually all the code from reopt.jl line 207 to 218 seems to be unnecessary for that same reason. But the other variables being fixed in this unreachable code block (e.g. dvDischargeFromStorage) do not cause issues like dvStorageToGrid does because they have b as an index, so if the storage constraints on them aren't added, they will also have a dimension of 0. Adding a b index to dvStorageToGrid should fix the unbounded issue.

FYI I addressed this on 6/4

src/constraints/storage_constraints.jl
```diff
@@ -26,9 +26,9 @@ end
 
 function add_general_storage_dispatch_constraints(m, p, b; _n="")
     # Constraint (4a): initial state of charge
-	@constraint(m,
-        m[Symbol("dvStoredEnergy"*_n)][b, 0] == p.s.storage.attr[b].soc_init_fraction * m[Symbol("dvStorageEnergy"*_n)][b]
-    )
+	# @constraint(m,
```
hdunham, line 29: We can't just get rid of this constraint entirely because the constraint setting init and final SOC equal is only applied to elec storage (and only when require_start_and_end_charge_to_be_equal is true, though that was my change)

src/constraints/storage_constraints.jl
```diff
@@ -99,6 +106,33 @@ function add_elec_storage_dispatch_constraints(m, p, b; _n="")
             sum(m[Symbol("dvStorageEnergy"*_n)][b])
         )
     end
+
+    if p.s.storage.attr[b] isa ElectricStorage && !isnothing(p.s.storage.attr[b].fixed_duration)
```
hdunham, line 110: I don't think we need these isa ElectricStorage checks because we're already inside the function add_elec_storage_dispatch_constraints(m, p, b; _n="") which is only passed ElectricStorage as b

src/constraints/storage_constraints.jl
hdunham: fixed

src/constraints/storage_constraints.jl
hdunham: done

PR #415, opened 2024-06-07
Hot Fix: Add thermal efficiency as input to chp defaults function

Bill-Becker:
To allow heuristic sizing with overrides to zero thermal efficiency defaults, e.g. for microturbine

### DIFF
--- a/src/core/chp.jl
+++ b/src/core/chp.jl
@@ -221,7 +221,8 @@ function CHP(d::Dict;
                                                                 boiler_efficiency=eff,
                                                                 avg_electric_load_kw=avg_electric_load_kw,
                                                                 max_electric_load_kw=max_electric_load_kw,
-                                                                is_electric_only=chp.is_electric_only)
+                                                                is_electric_only=chp.is_electric_only,
+                                                                thermal_efficiency=chp.thermal_efficiency_full_load)
     defaults = chp_defaults_response["default_inputs"]
     for (k, v) in custom_chp_inputs
         if k in [:installed_cost_per_kw, :tech_sizes_for_cost_curve]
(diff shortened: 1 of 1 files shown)


Bill-Becker:
@zolanaj This branch has been tested in a branch of the API with the webtool, which was the main reason for making this update. You can assume this has been successfully tested, to that end.

src/core/chp.jl
```diff
@@ -515,11 +518,18 @@ function get_chp_defaults_prime_mover_size_class(;hot_water_or_steam::Union{Stri
...
+                                prime_mover, size_class, hot_water_or_steam, boiler_effic, thermal_efficiency=NaN)
+    if isnan(thermal_efficiency)
+        therm_effic = prime_mover_defaults_all[prime_mover]["thermal_efficiency_full_load"][hot_water_or_steam][size_class+1]
+    else
+        therm_effic = thermal_efficiency
+    end
     if therm_effic == 0.0
+        if prime_mover == "micro_turbine" && isnothing(size_class)
+            size_class = "any"
+        end
```
zolanaj, line 530: given there's an error thrown after this, do we need this code? or should a warning be given instead?

zolanaj (approved):
Approved with one minor comment, thanks @Bill-Becker!

src/core/chp.jl
Bill-Becker: The size_class = "any" is just meant to populate the $size_class string in the error message because for micro_turbine the user may not have specified size_class, whereas for reciprocating_engine, the only way it will throw an error here is if they input a size_class of 1-3.

zolanaj: that makes sense, thank you!

PR #417, opened 2024-06-14
Add JuMP error message when non-optimal (typ infeasible)

Bill-Becker:
Currently only add first entry of the message which is typically the most informative and human readable.


Bill-Becker:
@rathod-b can you check this out?

rathod-b:
> @rathod-b can you check this out?

Hi Bill, I did a quick test and wanted to better understand the intent of this change. Are we trying to return a `dict` when models are infeasible, irrespective of what information is contained in the dictionary? I did a simple test and got a dictionary in response instead of the model object, which is an improvement. But the error key contained a very long error (see below) which doesnt make it clear that the problem was infeasible.

image

PR #424, opened 2024-07-25
Improve GitHub Actions Behavior

Bill-Becker:
- After updating MacOS runner to 13 because the previously used 11 was recently deprecated by GitHub Actions, and seeing that it does **occasionally** run, it's still way to intermittent/finicky. I'm proposing that we remove MacOS from the runner list and just run with Windows OS. 
  - Until we start seeing different test behavior across different OS's (I don't think we have), I think it's safe to stick to the one type of runner that's working for us (Windows).
- Suppress JuMP warning messages from 15-minute and multiple PVs test scenarios
  - To avoid flooding the test logs with those warnings

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         julia-version: ['1.8']
         julia-arch: [x64]
-        os: [ubuntu-latest, windows-latest, macOS-13]
+        os: [windows-latest, macOS-13]
 
     steps:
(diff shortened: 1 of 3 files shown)


adfarth (approved):
@Bill-Becker approving--thanks for this! Before merging, light suggestion to update the changelog so we remember why we did this. 

Maybe a future TODO would be the substring conditional that Bhavesh mentioned (filter out all "JuMP warnings")?

PR #428, opened 2024-07-29
Enable tracking of REopt.jl runs with PV and Wind from api.data.gov

Bill-Becker:
- Add header User-Agent=REopt.jl to PVWatts and Wind Toolkit API requests
  - This enables tracking of REopt.jl usage from within the api.data.gov platform, with admin monitoring access to the PVWatts and Wind Toolkit API's on api.data.gov
  - (Suggestion by Nick Muerdter for quick ability to track, including by user from the user's API key, even if for only when the users are using the PVWatts or Wind Toolkit APIs for production factor)

### DIFF
--- a/src/core/production_factor.jl
+++ b/src/core/production_factor.jl
@@ -72,7 +72,7 @@ function get_production_factor(wind::Wind, latitude::Real, longitude::Real, time
             resource = []
             try
                 @info "Querying Wind Toolkit for resource data ..."
-                r = HTTP.get(url; retries=5)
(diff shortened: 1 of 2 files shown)


adfarth:
@Bill-Becker is it intentional to merge this directly into master?

adfarth (approved):
@Bill-Becker I'm assuming that since the tests pass this change won't cause any issues, but let me know if there's something else you think is worth testing. Also flagging that this is set to merge directly into `master`, in case that was unintentional.

Bill-Becker:
> @Bill-Becker is it intentional to merge this directly into master?

Nope! Changed to develop.

PR #433, opened 2024-08-13
Make all tests run and show summary table

Bill-Becker:
This update to runtests.jl:

- Makes sure that all tests are run and displays a summary table at the end of the GitHub Actions step
  - By wrapping/nesting all the tests that we run on GitHub Actions within a high-level @testset, this gives a big summary table of all the nested @testsets after running all of them
  - Before, if there was certain errors in tests, the remaining tests would not run because it was not all nested under one
 - Remove the "Imported Xpress tests" @testset now that we plan to only use HiGHS or other OS solver going forward, and all of those imported tests have been vetted and pass
- By including verbose=true, we see the results and a time of each testset even if they pass
- Removes an "inputs" @testset because there are tons of other inputs tests within other @testsets

There is still a bit of inconsistency for when we wrap certain @testsets within a higher-level @testset to categorize and group tests, so there's room to improve the organization of this for clarity. This PR does NOT attempt to take that on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Classify the change according to the following categories:
 
 ## Develop 08-09-2024
 ### Changed
+- Improve the full test suite reporting with a verbose summary table, and update the structure to reflect long-term open-source solver usage
 - Removed MacOS from the runner list and just run with Windows OS, since MacOS commonly freezes and gets cancelled. We have not seen Windows OS pass while other OS's fail. .
 - Suppress JuMP warning messages from 15-minute and multiple PVs test scenarios to avoid flooding the test logs with those warnings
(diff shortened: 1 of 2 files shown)


adfarth (commented):
Thanks for bringing back the summary table! Did you test this with having a test fail? (e.g. temporarily changing a test value)

Bill-Becker:
> Thanks for bringing back the summary table! Did you test this with having a test fail? (e.g. temporarily changing a test value)

Yes because I had added these commits to another branch where I was developing something else, and I just cherry picked the commits into this branch off of develop. This branch had a couple of tests fail, with these same edits to runtests.jl:
https://github.com/NREL/REopt.jl/actions/runs/10363166826/job/28686315729

adfarth (approved):

PR #434, opened 2024-08-13
Fix proforma for thermal techs

Bill-Becker:
- Fixes proforma calcs including "simple" payback and IRR for thermal techs/scenarios. 
  - The operating costs of fuel and O&M were missing for all thermal techs such as ExistingBoiler, CHP, and others; this adds those sections of code to properly calculate the operating costs.
- Adds a test to validate the simple payback calculation with CHP (and ExistingBoiler) and checks the REopt result value against a spreadsheet proforma calculation (see Bill's spreadsheet)
- Adds a couple of missing techs for the initial capital cost calculation in financial.jl
- Fixes a few bugs here and there, including the setup_boiler_inputs in reopt_inputs.jl
- Improves DRY coding by replacing multiple instances of the same chunks of code for MACRS deprecation and CHP capital cost into functions that are now in financial.jl
- Simplifies the CHP sizing test to avoid a ~30 minute solve time, by avoiding the fuel burn y-intercept binaries which come with differences between full-load and part-load efficiency
- Added two new fields to the Metrics struct to append to for any fuel-burning technologies (just Generator, ExistingBoiler, Boiler, and CHP currently): fuel_cost_series and fuel_cost_series_bau which are now only added to the offtaker cashflows and not the owner/developer cashflows for third party.

### DIFF
--- a/src/core/reopt_inputs.jl
+++ b/src/core/reopt_inputs.jl
@@ -396,8 +396,8 @@ function setup_tech_inputs(s::AbstractScenario)
     end
 
     if "Boiler" in techs.all
-        setup_boiler_inputs(s, max_sizes, min_sizes, existing_sizes, cap_cost_slope, 
-            boiler_efficiency, production_factor, fuel_cost_per_kwh)
+        setup_boiler_inputs(s, max_sizes, min_sizes, existing_sizes, cap_cost_slope, boiler_efficiency,
+            om_cost_per_kw, production_factor, fuel_cost_per_kwh)
     end
 
     if "CHP" in techs.all
@@ -693,14 +693,16 @@ end
 
 """
     function setup_boiler_inputs(s::AbstractScenario, max_sizes, min_sizes, existing_sizes, cap_cost_slope, boiler_efficiency,
-        production_factor, fuel_cost_per_kwh)
+        om_cost_per_kw, production_factor, fuel_cost_per_kwh)
 
 Update tech-indexed data arrays necessary to build the JuMP model with the values for (new) boiler.
 This version of this function, used in BAUInputs(), doesn't update renewable energy and emissions arrays.
 """
-function setup_boiler_inputs(s::AbstractScenario, max_sizes, min_sizes, cap_cost_slope, om_cost_per_kw, boiler_efficiency, production_factor, fuel_cost_per_kwh)
+function setup_boiler_inputs(s::AbstractScenario, max_sizes, min_sizes, existing_sizes, cap_cost_slope, boiler_efficiency,
+                            om_cost_per_kw, production_factor, fuel_cost_per_kwh)
     max_sizes["Boiler"] = s.boiler.max_kw
     min_sizes["Boiler"] = s.boiler.min_kw
+    existing_sizes["Boiler"] = 0.0
     boiler_efficiency["Boiler"] = s.boiler.efficiency
     
     # The Boiler only has a MACRS benefit, no ITC etc.
--- a/src/results/existing_boiler.jl
+++ b/src/results/existing_boiler.jl
@@ -18,7 +18,7 @@
 """
 function add_existing_boiler_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _n="")
     r = Dict{String, Any}()
-
+    r["size_mmbtu_per_hour"] = round(value(m[Symbol("dvSize"*_n)]["ExistingBoiler"]) / KWH_PER_MMBTU, digits=3)
 	r["fuel_consumption_series_mmbtu_per_hour"] = 
         round.(value.(m[:dvFuelUsage]["ExistingBoiler", ts] for ts in p.time_steps) ./ KWH_PER_MMBTU, digits=5)
     r["annual_fuel_consumption_mmbtu"] = round(sum(r["fuel_consumption_series_mmbtu_per_hour"]), digits=5)
--- a/src/results/existing_boiler.jl
+++ b/src/results/existing_boiler.jl
@@ -1,6 +1,7 @@
 # REopt®, Copyright (c) Alliance for Sustainable Energy, LLC. See also https://github.com/NREL/REopt.jl/blob/master/LICENSE.
 """
 `ExistingBoiler` results keys:
+- `size_mmbtu_per_hour`
 - `fuel_consumption_series_mmbtu_per_hour` 
 - `annual_fuel_consumption_mmbtu`
 - `thermal_production_series_mmbtu_per_hour`
--- a/src/results/financial.jl
+++ b/src/results/financial.jl
@@ -334,7 +334,7 @@ end
 Get the depreciation schedule for MACRS. First check if tech.macrs_option_years in [5 ,7], then call function to return depreciation schedule
 Used in results/financial.jl and results/proformal.jl multiple times
 """
-function get_depreciation_schedule(p::REoptInputs, tech::AbstractTech, federal_itc_basis::Float64=0.0)
+function get_depreciation_schedule(p::REoptInputs, tech::Union{AbstractTech,AbstractStorage}, federal_itc_basis::Float64=0.0)
     schedule = []
     if tech.macrs_option_years == 5
         schedule = p.s.financial.macrs_five_year
--- a/src/results/financial.jl
+++ b/src/results/financial.jl
@@ -161,34 +161,17 @@ function initial_capex(m::JuMP.AbstractModel, p::REoptInputs; _n="")
     end
 
     if "CHP" in p.techs.all
-        # CHP.installed_cost_per_kw is now a list with potentially > 1 elements
-        cost_list = p.s.chp.installed_cost_per_kw
-        size_list = p.s.chp.tech_sizes_for_cost_curve
-        chp_size = value.(m[Symbol("dvPurchaseSize"*_n)])["CHP"]
-        if typeof(cost_list) == Vector{Float64}
-            if chp_size <= size_list[1]
-                initial_capex += chp_size * cost_list[1]  # Currently not handling non-zero cost ($) for 0 kW size input
-            elseif chp_size > size_list[end]
-                initial_capex += chp_size * cost_list[end]
-            else
-                for s in 2:length(size_list)
-                    if (chp_size > size_list[s-1]) && (chp_size <= size_list[s])
-                        slope = (cost_list[s] * size_list[s] - cost_list[s-1] * size_list[s-1]) /
-                                (size_list[s] - size_list[s-1])
-                        initial_capex += cost_list[s-1] * size_list[s-1] + (chp_size - size_list[s-1]) * slope
-                    end
-                end
-            end
-        else
-            initial_capex += cost_list * chp_size
-        #Add supplementary firing capital cost
-        # chp_supp_firing_size = self.nested_outputs["Scenario"]["Site"][tech].get("size_supplementary_firing_kw")
-        # chp_supp_firing_cost = self.inputs[tech].get("supplementary_firing_capital_cost_per_kw") or 0
-        # initial_capex += chp_supp_firing_size * chp_supp_firing_cost
-        end
+        chp_size_kw = value.(m[Symbol("dvPurchaseSize"*_n)])["CHP"]
+        initial_capex += get_chp_initial_capex(p, chp_size_kw)
     end
 
-    # TODO thermal tech costs
+    if "Boiler" in p.techs.all
+        initial_capex += p.s.boiler.installed_cost_per_kw * value.(m[Symbol("dvPurchaseSize"*_n)])["Boiler"]
+    end
+
+    if "AbsorptionChiller" in p.techs.all
+        initial_capex += p.s.absorption_chiller.installed_cost_per_kw * value.(m[Symbol("dvPurchaseSize"*_n)])["AbsorptionChiller"]
+    end
 
     if !isempty(p.s.ghp_option_list)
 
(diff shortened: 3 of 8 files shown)


src/results/proforma.jl
```diff
@@ -111,6 +97,7 @@ function proforma_results(p::REoptInputs, d::Dict)
         # In the two party case the developer does not include the fuel cost in their costs
         # It is assumed that the offtaker will pay for this at a rate that is not marked up
         # to cover developer profits
+        # TODO escalate fuel cost with p.s.financial.generator_fuel_cost_escalation_rate_fraction
```
adfarth, line 100: @Bill-Becker you can remove this TODO now, right?

src/core/reopt_inputs.jl
```diff
@@ -396,8 +396,8 @@ function setup_tech_inputs(s::AbstractScenario)
     end
 
     if "Boiler" in techs.all
-        setup_boiler_inputs(s, max_sizes, min_sizes, existing_sizes, cap_cost_slope, 
-            boiler_efficiency, production_factor, fuel_cost_per_kwh)
+        setup_boiler_inputs(s, max_sizes, min_sizes, existing_sizes, cap_cost_slope, boiler_efficiency,
```
adfarth, line 399: I'm curious how this wasn't causing any big issues before?

src/results/existing_boiler.jl
```diff
@@ -18,7 +18,7 @@
 """
 function add_existing_boiler_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _n="")
     r = Dict{String, Any}()
-
+    r["size_mmbtu_per_hour"] = round(value(m[Symbol("dvSize"*_n)]["ExistingBoiler"]) / KWH_PER_MMBTU, digits=3)
```
adfarth, line 21: any reason not to add this to the markdown text with the results keys?

src/results/financial.jl
```diff
@@ -286,10 +273,10 @@ function calculate_lcoe(p::REoptInputs, tech_results::Dict, tech::AbstractTech)
 
     capital_costs = new_kw * tech.installed_cost_per_kw # pre-incentive capital costs
 
-    annual_om = new_kw * tech.om_cost_per_kw # NPV of O&M charges escalated over financial life
+    annual_om = new_kw * tech.om_cost_per_kw 
 
     om_series = [annual_om * (1+p.s.financial.om_cost_escalation_rate_fraction)^yr for yr in 1:years]
-    npv_om = sum([om * (1.0/(1.0+discount_rate_fraction))^yr for (yr, om) in enumerate(om_series)])
+    npv_om = sum([om * (1.0/(1.0+discount_rate_fraction))^yr for (yr, om) in enumerate(om_series)]) # NPV of O&M charges escalated over financial life
```
adfarth, line 279: I just moved this comment here

src/results/proforma.jl
```diff
@@ -120,19 +107,95 @@ function proforma_results(p::REoptInputs, d::Dict)
...
+        m.om_series_bau += escalate_fuel(annual_fuel_bau, p.s.financial.generator_fuel_cost_escalation_rate_fraction)
+        m.om_series_bau += escalate_om(annual_om_bau)
+    end
+
+    # calculate CHP o+m costs, incentives, and depreciation
+    if "CHP" in keys(d) && d["CHP"]["size_kw"] > 0
+        update_metrics(m, p, p.s.chp, "CHP", d, third_party)
+    end
+
+    # calculate (new) Boiler o+m costs (just fuel, no non-fuel operating costs currently)
```
adfarth, line 134: Should be "Existing" not "(new)", right?

src/results/proforma.jl
```diff
@@ -120,19 +107,95 @@ function proforma_results(p::REoptInputs, d::Dict)
...
+        # BAU ExistingBoiler
+        fuel_cost_bau = d["ExistingBoiler"]["year_one_fuel_cost_before_tax_bau"]
+        m.om_series_bau += escalate_fuel(-1 * fuel_cost_bau, p.s.financial.existing_boiler_fuel_cost_escalation_rate_fraction)
+        var_om_bau = 0.0
+        fixed_om_bau = 0.0
+        annual_om_bau = -1 * (var_om_bau + fixed_om_bau)
         m.om_series_bau += escalate_om(annual_om_bau)
+    end    
+
+    # calculate (new) Boiler o+m costs and depreciation (no incentives currently)
```
adfarth, line 153: Should instead say (no incentives currently, other than MACRS)?

src/results/proforma.jl
```diff
@@ -284,10 +347,20 @@ function update_metrics(m::Metrics, p::REoptInputs, tech::AbstractTech, tech_nam
...
-    capital_cost = new_kw * tech.installed_cost_per_kw
+    if tech_name == "CHP"
+        capital_cost = get_chp_initial_capex(p, results["CHP"]["size_kw"])
+    else
+        capital_cost = new_kw * tech.installed_cost_per_kw
+    end
 
-    # owner is responsible for both new and existing PV maintenance in optimal case
-    if third_party
+    # owner is responsible for both new and existing PV (or Generator) maintenance in optimal case
```
adfarth, line 356: Given the if statement below, shouldn't this instead be "owner is responsible for only new_kw PV (or Generator) maintenance in optimal case." ?

src/results/proforma.jl
```diff
@@ -297,6 +370,12 @@ function update_metrics(m::Metrics, p::REoptInputs, tech::AbstractTech, tech_nam
     m.om_series += escalate_om(annual_om)
     m.om_series_bau += escalate_om(-1 * existing_kw * tech.om_cost_per_kw)
 
+    if tech_name == "CHP"
+        escalate_fuel(val, esc_rate) = [val * (1 + esc_rate)^yr for yr in 1:years]
```
adfarth, line 374: so there wasn't a way to make these functions globally available so they aren't defined twice?

src/results/proforma.jl
adfarth, line 1: @Bill-Becker  Documenting my observations from `proforma_results()` of a few inconsistencies that I think are worth addressing. We had originally discussed punting on the generator existing_kw issue, but since it is only one of several inconsistencies here, I think it could be worth addressing all in one go:
- For PV and Wind, we call the `update_metrics()` function. If the analysis is third party, then this only includes `new_kw` in the `annual_om` cost (not `total_kw`)
- For Generator, `update_metrics` is *not* called. In the custom code in this if statement, the `annual_om` cost does not distinguish between `new_kw` and `total_kw` for third party (it just uses total for both direct and third party). *However* the `annual_om` cost *does* exclude fuel costs for third party. 
- For CHP, the new code added within `update_metrics` does not do the same accounting of fuel costs as Generator for third party
- For Existing Boiler, Boiler, and Steam Turbine, the new code added does not do the same accounting of fuel costs as Generator for third party

In summary: it seems like all techs should follow the rule of, if `third_party == true`, then only count new_kw O&M costs in `om_series`, and do not include fuel costs in the `om_series` 

What do you think?

adfarth: You can also see (in line 282), that the `generator_fuel_cost_series` gets added into the `"offtaker_annual_free_cashflows"` and the existing generator fuel costs get separately added into the free_cashflows_bau

src/results/proforma.jl
Bill-Becker: Yes

src/results/proforma.jl
Bill-Becker: Yep

src/results/proforma.jl
Bill-Becker: Done

src/results/existing_boiler.jl
Bill-Becker: No reason, added.

src/results/financial.jl
Bill-Becker: Thanks.

src/results/proforma.jl
Bill-Becker: Yes, and since I am going to update Generator to do that too, I'm just going to say "new technologies" (even though Generator will not use this function)

src/core/reopt_inputs.jl
Bill-Becker: Good question! We were putting in the "existing_sizes" array for the argument where the setup_boiler_function() was expecting "om_cost_per_kw". Since the function call had the expected number of arguments, it didn't error there. Then it was actually assigning existing_sizes["Boiler"] = s.boiler.om_cost_per_kw, when it ran the line om_cost_per_kw["Boiler"] = s.boiler.om_cost_per_kw. I just tested this with the develop branch and verified inputs.existing_sizes["Boiler"] = [a non zero value] when I input a non-zero value for Boiler.om_cost_per_mmbtu_per_hour (gets converted into _per_kw). I guess having an "existing_size" for Boiler didn't cause any other issues - it still passes the test with or without a non-zero value.

src/results/proforma.jl
Bill-Becker: Right, I guess these are "local scope" functions when they are defined within the function. I don't think they will ever change though, so low risk of these becoming obsolete when the other function is updated.

Bill-Becker:
@adfarth ready for your re-review. The Solar Dataset test that is failing is unrelated to this PR, I'm pretty sure, but it seems like there's two issues with that test: 1) at least for the version of Julia I'm using, the test format of "String approximately equal to String" isn't valid anymore, and 2) the third test for Younde, Cameroon is responding with dataset=nsrdb instead of the expected "intl". I was seeing on the NSRDB website that they recently (July) made some major updates, and I'm wondering if Cameroon is now considered "nsrdb"?

src/results/proforma.jl
```diff
@@ -120,19 +108,95 @@ function proforma_results(p::REoptInputs, d::Dict)
             year_one_fuel_cost_bau = d["Generator"]["year_one_fuel_cost_before_tax_bau"]
         end
         if !third_party
-            annual_om = -1 * (fixed_and_var_om + d["Generator"]["year_one_fuel_cost_before_tax"])
```
adfarth, line 123: @Bill-Becker  I don't think we need a third_party if statement here. Just as you've done for the other techs (ExistingBoiler, Boiler, CHP), I think we can just get the `om_series`, and `fuel_cost_series` (opt and bau) here. Then, they will appropriately be added to the owner vs. offtaker cashflows in the third party if statements down below. 

As it is now, the Generator fuel costs get zeroed out if it's a 3rd party analysis and therefore don't get included in e.g. `r["offtaker_annual_free_cashflows"]`.

src/results/proforma.jl
```diff
@@ -195,20 +259,8 @@ function proforma_results(p::REoptInputs, d::Dict)
...
-                existing_genertor_fuel_cost_series = zeros(years)
-            end
-        else
-            existing_genertor_fuel_cost_series = zeros(years)
-            generator_fuel_cost_series = zeros(years)
-        end
-        net_energy_costs = -electricity_bill_series_bau - export_credit_series_bau + electricity_bill_series + 
-                           export_credit_series + annual_income_from_host_series - existing_genertor_fuel_cost_series + 
-                           generator_fuel_cost_series
+        net_energy_costs = -electricity_bill_series_bau - export_credit_series_bau - m.fuel_cost_series_bau + electricity_bill_series +
```
adfarth, line 262: It looks like `net_energy_costs` only gets used in `deductable_net_energy_costs`, which does not get used anywhere else. Could we remove this code altogether?

Bill-Becker: Yep.

src/results/proforma.jl
Bill-Becker: Agreed, done.

adfarth:
@Bill-Becker I sent a word doc summarizing open issues for 3rd party analyses

Bill-Becker:
> @Bill-Becker I sent a word doc summarizing open issues for 3rd party analyses

@adfarth see the latest commit 8f167336c117179e4bf5b119b3aa0c08d7563021 for what we discussed updating for now. I think the way I removed the "BAU O&M" which should have the both fixed **and** variable O&M of the Generator.existing_kw, subtracted from the optimal case O&M which includes both existing_kw + new_kw, should effectively separate existing O&M from new. The variable O&M part is certainly a question because the optimal case may use the "existing" portion of the Generator more or less than the BAU case, but our default O&M cost is all in the fixed /kw/year term, so I think we should mostly be good (and I still don't know how we model BAU Generator dispatch).

And then I added the om_series_bau to both offtaker cashflow terms in that commit too, as we discussed for completeness (even though they cancel).

I also made an issue about removing existing O&M from the 3rd party factor in the optimization, for "when we have time":
https://github.com/NREL/REopt.jl/issues/442

src/results/proforma.jl
```diff
@@ -117,22 +106,92 @@ function proforma_results(p::REoptInputs, d::Dict)
         if p.s.generator.existing_kw > 0
             fixed_and_var_om_bau = d["Generator"]["year_one_fixed_om_cost_before_tax_bau"] + 
                                    d["Generator"]["year_one_variable_om_cost_before_tax_bau"]
+            fixed_and_var_om -= fixed_and_var_om_bau
```
adfarth, line 109: @Bill-Becker I think we only want to subtract out the _bau costs if third_party is true. How it is now, for a *direct purchase* scenario the "optimal" `om_series` for the site would only account for new O&M costs, while we would want them to be all O&M costs. So I think we just need a third party if statement here, like there is in `update_metrics` for PV and Wind

adfarth: (Also I'm realizing a previous comment suggested removing the 3rd party if statement, but I think that was only applicable before this separation of new vs existing o&m costs.

Bill-Becker: Oh right, done!

adfarth (approved):
Looks good!

PR #332, opened 2023-12-01
Limit battery storage duration

atpham88:
## Summary
This PR adds user-input option to limit the hour duration for battery storage.

## Implementation notes
Changes in this PR to add this option:

**1. electric_storage.jl**: added a `max_duration_hours` input, default as a big number - 100,000 (hours), and `min_duration_hours`, default is 0 (hour). This defaults can be changed in `["ElectricStorage"]["max_duration_hours"]` and `["ElectricStorage"]["min_duration_hours"]`.
**2. storage_constraints.jl**: Add a constraint to limit battery energy capacity <= battery power rating * max_duration_hours and battery energy capacity >= battery power rating * min_duration_hours.
**3. runtests.jl**: Add "Storage Duration" test.

PR #447, opened 2024-10-08
LDES project updates

hdunham:
### Added
- Added inputs **om_cost_per_kw** and **om_cost_per_kwh** to `ElectricStorage` for modeling capacity-based O&M 
- Added outputs **lifecycle_om_cost_after_tax** and **year_one_om_cost_before_tax** to `ElectricStorage` 
- Added new input **self_discharge_fraction_per_timestep** to `ElectricStorage` for modeling battery self-discharge
- Added input option **can_export_to_grid** (defaults to _false_) to `ElectricStorage` and decision variable **dvStorageToGrid**
- Added input **fixed_duration** to `ElectricStorage`, which fixes the ratio between **size_kw** and **size_kwh** in the optimized solution if provided
- Added input option **optimize_soc_init_fraction** (defaults to _false_) to `ElectricStorage`, which makes the optimization choose the inital SOC (equal to final SOC) instead of using **soc_init_fraction**. The initial SOC is also constrained to equal the final SOC, which eliminates the "free energy" issue. We currently do not fix SOC when **soc_init_fraction** is used because this has caused infeasibility. 
- Added output **initial_capital_cost** to all techs
- Added ability to model multiple `ElectricStorage` at once. Inputs are provided as a _list_ of _dict_ under the `ElectricStorage` key and outputs are similarly formatted. This mirrors the existing multiple `PV` implementation. This functionality is not yet implemented for MPC scenarios.
### Fixed
- Added missing outputs **lifecycle_export_benefit_before_tax** and **year_one_export_benefit_after_tax** to `ElectricTariff`
- Add missing output **year_one_om_cost_before_tax** to `PV`


CHANGELOG.md
```diff
@@ -23,6 +25,35 @@ Classify the change according to the following categories:
     ### Deprecated
     ### Removed
 
+## Develop LDES
```
adfarth, line 28: @hdunham I'm seeing some discrepancies and overlap between this branch and: 
- https://github.com/NREL/REopt.jl/pull/354/files -- for the `ElectricStorage` O&M changes by @lixiangk1. Is that PR or this PR more up-to-date for those changes? 
- https://github.com/NREL/REopt.jl/pull/332 -- for the `ElectricStorage` min, max duration inputs by @atpham88, which is an alternative way to implement the **fixed_duration** changes in this branch. 

I think it makes sense to merge each of those branches above into this PR. What do you think?

Issue #448, opened 2024-10-11
ASHP min_allowable_kw not enforced when max_kw is at default large number


zolanaj:
A test script revealed the following for an ASHPSpaceHeater:
`min_allowable_kw = 909.761870048896`
`size_kw = 349.9673825862397`

This can be adjusted by providing better big-M's for the derived max values in `p.seg_max_size` so that any user-specified value for `max_kw` isn't overridden.

Bill-Becker:
@zolanaj is this resolved?

PR #449, opened 2024-10-11
Fix ASHP min allowable size enforcement

zolanaj:
Addresses https://github.com/NREL/REopt.jl/issues/448

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ Classify the change according to the following categories:
 - Change type of **value_of_lost_load** in **FinancialInputs** struct to fix convert error when user provides an _Int_
 - Change international location in "Solar Dataset" test set from Cameroon to Oulu because the locations in the NSRDB have been expanded significantly so there is now an NSRDB point at Cameroon
 - Handle edge case where the values of **outage_start_time_steps** and **outage_durations** makes an outage extend beyond the end of the year. The outage will now wrap around to the beginning of the year.
+- Enforce minimum allowable sizes for ASHP technologies by introducing improved big-M values for segmented size constraints.
 
 ## v0.48.0
 ### Added
--- a/src/core/ashp.jl
+++ b/src/core/ashp.jl
@@ -444,14 +444,14 @@ function get_ashp_performance(cop_reference,
                 cf_reference,
                 reference_temps,
                 ambient_temp_degF,
-                back_up_temp_threshold_degF = 10.0
+                back_up_temp_threshold_degF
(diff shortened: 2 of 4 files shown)


Bill-Becker:
Hey Alex, this is more of a question from the original PR than this fix, but it looks like there's a couple of places where we have default values hard-coded in various places which seem redundant to defaults which are either stored in the ashp_defaults.json file or in other locations in /core/ashp.jl:

This peak_load_thermal_factor looks like it's meant to be the same value as the min_allowable_peak_capacity_fraction:
https://github.com/NREL/REopt.jl/blob/452a7175f430b84f8a6bb436dd64144c8652a1a6/src/core/ashp.jl#L498

And this default for the value of back_up_temp_threshold_degF for this function: https://github.com/NREL/REopt.jl/blob/452a7175f430b84f8a6bb436dd64144c8652a1a6/src/core/ashp.jl#L447

Not a big deal, but ideally we would only have to make changes to the default values in one place in the future. If you want to punt on updating for now, maybe we could at least make a comment for the first time these parameters are assigned the default value that if the default changes, you also need to update in these other locations.

Bill-Becker (approved):
Approved with a suggestion about linking defaults for a couple of parameters.

zolanaj:
> Hey Alex, this is more of a question from the original PR than this fix, but it looks like there's a couple of places where we have default values hard-coded in various places which seem redundant to defaults which are either stored in the ashp_defaults.json file or in other locations in /core/ashp.jl:
> 
> This peak_load_thermal_factor looks like it's meant to be the same value as the min_allowable_peak_capacity_fraction:
> 
> https://github.com/NREL/REopt.jl/blob/452a7175f430b84f8a6bb436dd64144c8652a1a6/src/core/ashp.jl#L498
> 
> And this default for the value of back_up_temp_threshold_degF for this function:
> 
> https://github.com/NREL/REopt.jl/blob/452a7175f430b84f8a6bb436dd64144c8652a1a6/src/core/ashp.jl#L447
> 
> Not a big deal, but ideally we would only have to make changes to the default values in one place in the future. If you want to punt on updating for now, maybe we could at least make a comment for the first time these parameters are assigned the default value that if the default changes, you also need to update in these other locations.

Thanks @Bill-Becker! I agree, we should not have these defaults in place within the function definition - they should be in one place.  I've pushed an update that should remove then and a fix to some tests that were previously incomplete.  Neither function exports to the API so it shouldn't be a breaking change.  I'll merge once I've confirmed tests pass.

PR #450, opened 2024-10-14
fix timestep issue with generator fuel cost in subhourly data

bpulluta:
# Update Fuel Cost Calculation to Account for Time Step Duration

## Changes
- Modified the fuel cost calculation to correctly account for the time step duration.
- Fuel costs are given per time step (e.g., per 30 minutes), not per hour.
- To get the correct hourly cost, we need to divide the cost per time step by the number of hours in a time step: `fuel_cost_per_time_step / hours_per_time_step`
- This conversion ensures we're using the correct hourly rate in our calculations.


bpulluta:
Add a test for the subhourly fuel cost

Bill-Becker:
> Add a test for the subhourly fuel cost

I don't see the test. Can you also see what's going on with the failing Generator tests?

bpulluta:
> > Add a test for the subhourly fuel cost
> 
> I don't see the test. Can you also see what's going on with the failing Generator tests?

I just put that there as a reminder. @Bill-Becker , i did just add tests for subhourly analysis to the offgrid case

Bill-Becker:
I think we can close this and this PR replaces it (I also want to avoid those updates to the Project.toml and Manifest.toml - we should not be adding HiGHS.jl to the REopt.jl dependencies.

https://github.com/NREL/REopt.jl/pull/453

bpulluta:
replaced PR with https://github.com/NREL/REopt.jl/pull/453

PR #453, opened 2024-10-24
Fix fuel cost calculation for non-hourly load data

Bill-Becker:
fuel_cost_per_kwh was erroneously being divided by time_steps_per_hour which is 2 for 30-minute interval and 4 for 15-minute interval loads. The fuel used/consumed calculation already accounts for time_steps_per_hour to get the correct energy used in each timestep, so we need the actual (input) fuel_cost_per_kwh to be preserved to multiply that with fuel used.

This PR avoids dividing fuel cost value by time_steps_per_hour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ Classify the change according to the following categories:
     ### Deprecated
     ### Removed
 
+## Develop fix-nonhourly-fuel-cost
+### Fixed
+- Modified the fuel cost calculation to correctly account for the time step duration when using non-hourly data.
+
 ## Develop degradation-cleanup
 ### Added
 - Battery residual value if choosing replacement strategy for degradation
--- a/src/core/utils.jl
+++ b/src/core/utils.jl
@@ -290,7 +290,7 @@ end
 Convert a per hour value (eg. dollars/kWh) to time series that matches the settings.time_steps_per_hour
 """
 function per_hour_value_to_time_series(x::T, time_steps_per_hour::Int, name::String) where T <: Real
-    repeat([x / time_steps_per_hour], 8760 * time_steps_per_hour)
+    repeat([x], 8760 * time_steps_per_hour)
 end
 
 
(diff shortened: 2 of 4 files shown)


bpulluta:
@Bill-Becker, you want me to add the tests we talked about to check that the hourly and subhourly results are the same? Or is that not neccessary?

Bill-Becker:
> @Bill-Becker, you want me to add the tests we talked about to check that the hourly and subhourly results are the same? Or is that not neccessary?

I think that's a good idea in case we don't have that test coverage already. Just do a simple (short solve time) scenario with Generator that runs and check that that the annual fuel cost is the same for both 15 or 30 minute and hourly.

bpulluta:
> > @Bill-Becker, you want me to add the tests we talked about to check that the hourly and subhourly results are the same? Or is that not neccessary?
> 
> I think that's a good idea in case we don't have that test coverage already. Just do a simple (short solve time) scenario with Generator that runs and check that that the annual fuel cost is the same for both 15 or 30 minute and hourly.

@Bill-Becker Okay, added to the tests. Let me know if you'd like me to put it into a separate test section called Subhourly or something. Seems like you fixed the discrepancy with your solution

bpulluta (approved):

bpulluta:
@Bill-Becker did this get integrated separately? Wondering if this was fixed elsewhere

Bill-Becker:
> @Bill-Becker did this get integrated separately? Wondering if this was fixed elsewhere

Hmmm. I vaguely remember you opened up a second PR and that one took precedence over the first one, but I don't know how it  would have gotten fixed if neither were merged?!

bpulluta:
@Bill-Becker, this pr replaced the last one. We might need to open this and review. I approved this and is ready for merging after your review

bpulluta:
@Bill-Becker this pr replaced the previous one that was deleted. We might need to open this and review. I approved this and is ready for merging after your review.

Bill-Becker:
@bpulluta Can you take a stab at making the Offgrid test that you added a little "easier". It's passing for me locally, but failing on GitHub Actions. It takes a long time to solve, and I'm wondering if that's causing some inconsistency in the results on GitHub Actions with the limited computational resources it has to solve.

PR #454, opened 2024-10-26
Normalize and scale load profile, refactor (mostly heating) load code

Bill-Becker:
- Add new input for each load called `normalize_and_scale_load_profile_input` which takes the input custom load profile (e.g. `ElectricLoad.loads_kw`), normalizes it to timestep_energy/yearly_energy for each timestep, and then applies and scales that normalized profile to the input annual monthly energy values.
  - Before, we could only scale annual or monthly energy inputs to the CRB profiles. 
  - This leverages the CRB calculations for applying and scaling annual and monthly energy to the load profile.
  - This will be especially helpful to use along with the upcoming Load Profile Builder feature in the web tool
- Put all of the default `annual_[energy]` data for CRBs by city/zone into .json files in the `data/load_profiles` folder, and then load that data within the load profile constructor functions (`BuiltIn[Load]`) where they lived before (hard to navigate those load file with all that data).
- Consolidate all of the duplicate code for processing the inputs and building structs for the three different types of heating loads into a single `HeatingLoad` and `BuiltInHeatingLoad` functions (3x2=6 functions down to 2) with a new argument of `(heating_)load_type`.
- Update the `simulated_load` function for improved validation flow and adding the `normalize_and_scale_load_profile_input` option


Bill-Becker:
@zolanaj This is needed for end-of-December milestone in the web tool, if you're able to do at least a fatal flaw review soon. Thanks!

test/runtests.jl
```diff
@@ -2910,5 +2912,95 @@ else  # run HiGHS tests
...
+            @test abs(s.space_heating_load.loads_kw[6] / s.space_heating_load.loads_kw[4] - input_data["SpaceHeatingLoad"]["fuel_loads_mmbtu_per_hour"][6] / input_data["SpaceHeatingLoad"]["fuel_loads_mmbtu_per_hour"][4]) < 0.001
+            
+            # Check consistency with simulated_load function
+            d_sim_load = Dict([
+                ("load_type", "space_heating"),
+                ("normalize_and_scale_load_profile_input", true),
+                ("load_profile", input_data["SpaceHeatingLoad"]["fuel_loads_mmbtu_per_hour"]),
+                ("monthly_mmbtu", input_data["SpaceHeatingLoad"]["monthly_mmbtu"]),
+                ("addressable_load_fraction", address_frac)
+                ])
```
zolanaj, line 3009: > these test monthly builds but not annual, I'll add a set.

see ddcbec3393ffe3a8cc3a4fc0da114843b44f1c57

src/core/heating_cooling_loads.jl
```diff
@@ -1,249 +1,250 @@
...
-            addressable_load_fraction = convert(Vector{Real}, addressable_load_fraction)
-        elseif typeof(addressable_load_fraction) <: Vector{}
-            addressable_load_fraction = convert(Real, addressable_load_fraction[1])            
-        else
-            addressable_load_fraction = convert(Real, addressable_load_fraction)
+    # Determine which type of heating load to build
+    if load_type == "space_heating"
+        load = :SpaceHeatingLoad
+        struct_type = SpaceHeatingLoad
+    elseif load_type == "domestic_hot_water"
```
zolanaj, line 68: with this naming convention, should we rename the `dhw_load` object in the scenario object to `domestic_hot_water_load`?  If you agree, I may add that to the other const refactoring in a separate PR so it doesn't block progress on this.  That way, load types would stay consistent

test/test_with_xpress.jl
```diff
@@ -1549,9 +1549,9 @@ end
     input_data["DomesticHotWaterLoad"]["doe_reference_name"] = building
     elec_load = REopt.ElectricLoad(latitude=latitude, longitude=longitude, doe_reference_name=building)
     input_data["ElectricLoad"]["annual_kwh"] = elec_load_multiplier * sum(elec_load.loads_kw)
-    space_load = REopt.SpaceHeatingLoad(latitude=latitude, longitude=longitude, doe_reference_name=building, existing_boiler_efficiency=input_data["ExistingBoiler"]["efficiency"])
+    space_load = REopt.HeatingLoad(load_type="space_heating", latitude=latitude, longitude=longitude, doe_reference_name=building, existing_boiler_efficiency=input_data["ExistingBoiler"]["efficiency"])
     input_data["SpaceHeatingLoad"]["annual_mmbtu"] = heat_load_multiplier * space_load.annual_mmbtu / input_data["ExistingBoiler"]["efficiency"]
-    dhw_load = REopt.DomesticHotWaterLoad(latitude=latitude, longitude=longitude, doe_reference_name=building, existing_boiler_efficiency=input_data["ExistingBoiler"]["efficiency"])
```
zolanaj, line 1554: It's good that we're keeping up with this, but is it worth a conversation revisiting whether we keep these?  I think that with battery degaradation updated, all indicator constraints have been refactored for HiGHS compatibility.

zolanaj (approved):
Thanks @Bill-Becker!  This looks great, and I tested a fair amount and didn't see any concerns.  I have a few suggestions, some of which are in the form of additional commits that you can revert if you disagree. The rest of the comments are more for discussion and potential further refactoring.  This is good to merge either way (assuming the merged to update with develop passes).

docs/src/reopt/inputs.md
```diff
@@ -139,17 +139,17 @@ REopt.ColdThermalStorageDefaults
...
 ```@docs
-REopt.SpaceHeatingLoad
+REopt.HeatingLoad
 ```
 
 ## ProcessHeatLoad
 ```@docs
-REopt.ProcessHeatLoad
+REopt.HeatingLoad
 ```
```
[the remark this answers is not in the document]
Bill-Becker: Yeah, that makes sense.

src/core/simulated_load.jl
```diff
@@ -14,23 +14,48 @@ One particular aspect of this function specifically for the webtool/UI is the he
...
+    else
+        doe_reference_name_input = get(d, "doe_reference_name", nothing)
+    end
     percent_share_input = get(d, "percent_share", Real[])
-    
-    # Determine which set of valid names to use based on the origin of the input
     valid_names = default_buildings
     if load_type == "process_heat"
         valid_names = default_process_types
     end
```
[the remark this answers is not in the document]
Bill-Becker: Updated to uppercase names for these consts in 79078bcb2b2effefc7345e6f21d208247d6bc455.

test/runtests.jl
Bill-Becker: Thanks!

src/core/heating_cooling_loads.jl
Bill-Becker: I agree with that change. But yeah I just looked at the "search and replace" and it's a lot especially if we want to include updated the names of the results from e.g. "thermal_to_dhw_load..." to "thermal_to_domestic_hot_water_load...". The results would require some renames on the API side as well. For now, let's punt and put on the TODO list.

test/test_with_xpress.jl
Bill-Becker: Right, I agree we could remove "test_with_xpress.jl". Let's punt on that for now though.

Bill-Becker:
> Thanks @Bill-Becker! This looks great, and I tested a fair amount and didn't see any concerns. I have a few suggestions, some of which are in the form of additional commits that you can revert if you disagree. The rest of the comments are more for discussion and potential further refactoring. This is good to merge either way (assuming the merged to update with develop passes).

Thanks for the review @zolanaj! I agreed with all of your commits and made a couple of updates from your comments. Waiting on tests to pass after those updates, then I will merge.

PR #455, opened 2024-11-20
Set option to force minimum fraction of load served by ASHP if purchased

zolanaj:
### Added
- Added new attribute `` to **ASHPSpaceHeater** and **ASHPWaterHeater** technologies.  When this is populated, this imposes a constraint on the minimum fraction of load served by the ASHP system in every period, if the system is purchased.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,10 @@ Classify the change according to the following categories:
     ### Deprecated
     ### Removed
 
-
+## Develop min-load-to-ashp
+### Added
+- Added new attribute `` to **ASHPSpaceHeater** and **ASHPWaterHeater** technologies.  When this is populated, this imposes a constraint on the minimum fraction of load served by the ASHP system in every period, if the system is purchased.
+ 
 ## Develop
 ### Added
 - Battery residual value if choosing replacement strategy for degradation
--- a/src/constraints/thermal_tech_constraints.jl
+++ b/src/constraints/thermal_tech_constraints.jl
@@ -107,29 +107,47 @@ end
     
 
 function add_ashp_force_in_constraints(m, p; _n="")
-    if "ASHPSpaceHeater" in p.techs.ashp && p.s.ashp.force_into_system
-        for t in setdiff(p.techs.can_serve_space_heating, ["ASHPSpaceHeater"])
-            for ts in p.time_steps
-                fix(m[Symbol("dvHeatingProduction"*_n)][t,"SpaceHeating",ts], 0.0, force=true)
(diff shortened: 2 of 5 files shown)


Bill-Becker:
@zolanaj so if we have a scenario that has a peak load of say 100 ton, and we say min service fraction is 50%, then REopt will size ASHP to 50 ton (at least). But if the "typical load" is more like 50 ton, this still allows the ASHP to only serve 50% of the 50 ton, or 25 tons for that typical load?

zolanaj:
> @zolanaj so if we have a scenario that has a peak load of say 100 ton, and we say min service fraction is 50%, then REopt will size ASHP to 50 ton (at least). But if the "typical load" is more like 50 ton, this still allows the ASHP to only serve 50% of the 50 ton, or 25 tons for that typical load?

Slightly rephrasing this: f the peak load is 100 ton and every other hour the load is 25 ton, then if purchased, the system would produce a minimum of 50 tons in the peak period and 25 tons in every other period.

I'm considering this "option A" and working on a separate option that will serve as much of the load as possible if the system is purchased ("option B" but not in order of preference - just the order of what I'm attempting.)  I think that option B is more in line with the ask but this was the easier implementation when I drew it up.

Bill-Becker:
> > @zolanaj so if we have a scenario that has a peak load of say 100 ton, and we say min service fraction is 50%, then REopt will size ASHP to 50 ton (at least). But if the "typical load" is more like 50 ton, this still allows the ASHP to only serve 50% of the 50 ton, or 25 tons for that typical load?
> 
> @Bill-Becker Slightly rephrasing this: 5f the peak load is 100 ton and every other hour the load is 50 ton, then if purchased, the system would produce a minimum of 50 tons in the peak period and 25 tons in every other period.
> 
> I'm considering this "option A" and working on a separate option that will serve as much of the load as possible if the system is purchased ("option B" but not in order of preference - just the order of what I'm attempting.) I think that option B is more in line with the ask but this was the easier implementation when I drew it up.

Ok, yeah that's what I thought - this would work pretty well when the peak-to-typical/avg load ratio is relatively low (<1.5), but I'm concerned this will still lead to a lot of the same kind of "herky jerky" dispatch for higher ratios that we're seeing now because REopt will just do the minimum of what's required if it's not economic with the temp-dependent COP and TOU/demand pricing.

Bill-Becker:
@zolanaj do we think the new functionality that you developed would replace this, or would this be another option? I personally think we just need what you recently developed with forced dispatch.

Bill-Becker:
> @zolanaj do we think the new functionality that you developed would replace this, or would this be another option? I personally think we just need what you recently developed with forced dispatch.

@zolanaj just pinging you on this comment.

zolanaj:
Closing as we went with the "max-output" option.

PR #459, opened 2024-12-05
Allow user to overwrite automatic hybrid GHP sizing

lixiangk1:
Change hybrid GHP workflow from (1) running a 2-year simulation to guess if an auxiliary heating or cooling unit is required and outputting results to (2) running non-hybrid first in all hybrid runs and comparing to hybrid outputs to guarantee an actual hybrid solution where possible

[fy25_ghp_updates.pdf](https://github.com/user-attachments/files/22391096/fy25_ghp_updates.pdf)


lixiangk1:
@atpham88 Somewhat but not completely related, when I was testing this, I came across a case where if I selected "Heater", REopt produced a reasonable hybrid result, but if I selected "Cooler" ("Automatic" also selects a cooler in this case), REopt exited out of GhpGhx.jl after hitting the max number of iterations with some ridiculous result of 16,000 boreholes cause it didn't converge. `GHP.ghpghx_chosen_outputs` returned an empty dictionary. What would be the ideal behavior in this case? Do we want to throw an error message whenever we hit max iterations because that most likely indicates a convergence issue and just return a non-hybrid result?

atpham88:
> @atpham88 Somewhat but not completely related, when I was testing this, I came across a case where if I selected "Heater", REopt produced a reasonable hybrid result, but if I selected "Cooler" ("Automatic" also selects a cooler in this case), REopt exited out of GhpGhx.jl after hitting the max number of iterations with some ridiculous result of 16,000 boreholes cause it didn't converge. `GHP.ghpghx_chosen_outputs` returned an empty dictionary. Which might be what's messing up the webtool outputs? What would be the ideal behavior in this case? Do we want to throw an error message whenever we hit max iterations because that most likely indicates a convergence issue and just return a non-hybrid result?

I tested with "Cooler" and had the same issue (borehole number is 67,575). But when I tested "Heater" and got the same number of boreholes as the no hybrid case. Does it mean, "Heater" is the wrong guess in this case? Let's chat more in our meeting today.

src/core/scenario.jl
```diff
@@ -577,33 +580,60 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
 
             if hybrid_ghx_sizing_method == "Automatic"
```
Bill-Becker, line 581: We can punt on this for now, but let's make it a priority for the GHP task next FY to move all this GHP code to a separate file with a function to load and use in scenario.jl.

src/results/heating_cooling_load.jl
```diff
@@ -64,7 +64,7 @@ function add_heating_load_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict
     process_heat_load_series_kw  =   p.s.process_heat_load.loads_kw
 
     existing_boiler_efficiency =   isnothing(p.s.existing_boiler) ? EXISTING_BOILER_EFFICIENCY : p.s.existing_boiler.efficiency
-    
+    # print("\nheating_cooling_load.jl", isnothing(p.s.existing_boiler))
```
Bill-Becker, line 67: Remove

Bill-Becker (approved):
I left a TODO comment that we can punt on for now. Can you add a GitHub Issue for that?

PR #426, opened 2024-07-26
Add Grid Renewable Energy Fraction

adfarth:
### Added 
- Add the following inputs to account for the clean or renewable energy fraction of grid-purchased electricity: 
  - ElectricUtility **cambium_cef_metric** to utilize clean energy data from NREL's Cambium database
  - ElectricUtility **renewable_energy_fraction_series** to supply a custom grid clean or renewable energy scalar or series
  - Site **include_grid_renewable_fraction_in_RE_constraints** - to allow user to choose whether to include grid RE in min max constraints
- Add input option optimize_soc_init_fraction (defaults to false) to ElectricStorage, which makes the optimization choose the inital SOC (equal to final SOC) instead of using soc_init_fraction. The initial SOC is also constrained to equal the final SOC, which eliminates the "free energy" issue. We currently do not fix SOC when soc_init_fraction is used because this has caused infeasibility.
- Add the following outputs: 
  - ElectricUtility **annual_renewable_electricity_supplied_kwh**
  - Site **onsite_and_grid_renewable_electricity_fraction_of_elec_load**
  - Site **onsite_and_grid_renewable_energy_fraction_of_total_load**
  -  ElectricLoad **annual_electric_load_with_thermal_conversions_kwh** which calculates end-use electrical load after including electric consumption by heaters and chillers

### Changed
- Change name of the following inputs: 
  - ElectricUtility input **cambium_metric_col** changed to **cambium_co2_metric**
- Change name of the following outputs:
  - ElectricUtility **cambium_emissions_region** changed to **cambium_region**
  - Site **annual_renewable_electricity_kwh** changed to **annual_onsite_renewable_electricity_kwh**
  - Site **renewable_electricity_fraction** changed to **onsite_renewable_electricity_fraction_of_elec_load** 
  - Site **total_renewable_energy_fraction** changed to **onsite_renewable_energy_fraction_of_total_load**
- Changed name of function (also available as endpoint through REopt API) from **cambium_emissions_profile** to **cambium_profile**
- Update AVERT emissions data to v4.3, which uses Regional Data Files for year 2023 for CONUS. For Alaska and Hawaii (regions AKGD, HIMS, HIOA), updated eGRID data to eGRID2022 datafile, adjusted to CO2e values. Emissions profiles are saved in `data/emissions/AVERT_Data`. 
[...]

PR #460, opened 2024-12-06
Update Emissions to Latest Data

adfarth:
- Updated AVERT emissions data to v4.3, which uses Regional Data Files for year 2023 for CONUS. For Alaska and Hawaii (regions AKGD, HIMS, HIOA), updated eGRID data to eGRID2022 datafile, adjusted to CO2e values. Emissions profiles are saved in `data/emissions/AVERT_Data`. 
- Updated Cambium API call for CO2e emissions within CONUS to Cambium 2023 dataset.  This includes updates to the default values and valid options for the following **ElectricUtility** inputs:  **cambium_scenario**, **cambium_location_type**, and **cambium_start_year**
- Updated EMISSIONS_DECREASE_DEFAULTS from 0.02163 to 0.0459 based on Cambium 2023 data


src/REopt.jl
```diff
@@ -100,10 +100,10 @@ const FUEL_DEFAULTS = Dict(
     )
 )
 const EMISSIONS_DECREASE_DEFAULTS = Dict(
```
hdunham, line 102: If this is Cambium data, why do we use a const for this instead of getting it from the Cambium API?

src/core/electric_utility.jl
```diff
@@ -160,10 +160,9 @@ struct ElectricUtility
         ### Grid Climate Emissions Inputs ### 
         # Climate Option 1 (Default): Use levelized emissions data from NREL's Cambium database by specifying the following fields:
         cambium_scenario::String = "Mid-case", # Cambium Scenario for evolution of electricity sector (see Cambium documentation for descriptions). 
-            ## Options: ["Mid-case",  "Mid-case with tax credit expiration",  "Low renewable energy cost", "Low renewable energy cost with tax credit expiration",   "High renewable energy cost", "High electrification",  "Low natrual gas prices", "High natrual gas prices", "Mid-case with 95% decarbonization by 2050",  "Mid-case with 100% decarbonization by 2035"]
```
hdunham, line 163: Why get rid of option comment?

src/core/electric_utility.jl
```diff
@@ -559,22 +558,22 @@ function cambium_emissions_profile(; scenario::String,
...
             "latitude" => string(round(latitude, digits=3)),
             "longitude" => string(round(longitude, digits=3)), 
-            "start_year" => string(start_year), # biennial from 2022-2050 (data year covers nominal year and years proceeding; e.g., 2040 values cover time range starting in 2036)
+            "start_year" => string(start_year), # data year covers nominal year and 4 years proceeding; e.g., 2040 values cover time range starting in 2036
             "lifetime" => string(lifetime), # Integer 1 or greater (Default 25 yrs)
             "discount_rate" => "0.0", # Zero = simple average (a pwf with discount rate gets applied to projected CO2 costs, but not quantity.)
             "time_type" => "hourly", # hourly or annual
             "metric_col" => metric_col, # lrmer_co2e
             "smoothing_method" => "rolling", # rolling or none (only applicable to hourly queries). "rolling" best with TMY data; "none" best if 2012 weather data used.
-            "gwp" => "100yrAR6", # Global warming potential values. Default: "100yrAR6". Options: "100yrAR5", "20yrAR5", "100yrAR6", "20yrAR6" or a custom tuple [1,10.0,100] with GWP values for [CO2, CH4, N2O]
```
hdunham, line 577: Is this supposed to be reverted before merging?

test/runtests.jl
```diff
@@ -2295,40 +2301,36 @@ else  # run HiGHS tests
...
-                    @test results["ElectricStorage"]["size_kwh"] ≈ 170.94 atol=1
+                    @test results["ElectricStorage"]["size_kwh"] ≈ 156.4 rtol=0.01
                     @test !haskey(results, "Generator")
                     # Renewable energy
-                    @test results["Site"]["renewable_electricity_fraction"] ≈ 0.78586 atol=1e-3
+                    @test results["Site"]["renewable_electricity_fraction"] ≈ 0.745 rtol=0.01
                     @test results["Site"]["renewable_electricity_fraction_bau"] ≈ 0.132118 atol=1e-3 #0.1354 atol=1e-3
                     @test results["Site"]["annual_renewable_electricity_kwh_bau"] ≈ 13308.5 atol=10 # 13542.62 atol=10
                     @test results["Site"]["total_renewable_energy_fraction_bau"] ≈ 0.132118 atol=1e-3 # 0.1354 atol=1e-3
                     # CO2 emissions - totals ≈  from grid, from fuelburn, ER, $/tCO2 breakeven
```
hdunham, line 2325: not in your changes, but do you know what this comment means? it doesn't make any sense to me. what does ER stand for? Why is breakeven val in the list of what equals emissions totals? why approx equal?

hdunham (approved):
looks good I just had a couple questions with one maybe todo

src/core/electric_utility.jl
adfarth: I got rid of it here I think just to save space, since we have the full list of options up above in the markdown text. But, I think we don't have consistency throughout the code in whether we duplicate explanatory text down here too. Do you have a strong opinion on this?

src/core/electric_utility.jl
adfarth: Oh great catch. I don't think this necessarily needs to be reverted because [1, 29.8, 273.0] should == using "100yrAR6". However, it would be good to find out if "100yrAR6" is still a viable option (currently, it errors). I'm still waiting to hear back from the Cambium team on this so let's keep this as an open TODO.

src/REopt.jl
adfarth: That's a good point. I think the main argument for keeping it this way is that we can avoid adding an extra API call to every REopt run, since this value should only change whenever Cambium is updated. We can also avoid having to add the API call or an endpoint to the REopt API/web tool.

hdunham: makes sense!

test/runtests.jl
adfarth: hmmm this isn't my comment but I think it might mean: here we are testing some things related to CO2 Emissions Totals, including approx results for these things. ER = emissions reduction, I believe.

adfarth:
Closing this PR as this branch has been merged into https://github.com/NREL/REopt.jl/pull/426.

PR #462, opened 2024-12-11
Maximize ASHP output in dispatch

zolanaj:
### Added
- New parameter `force_discpatch` in the **ASHPSpaceHeater** and **ASHPWaterHeater** technologies (default = `false`).  When set to `true`, the ASHP's thermal output will be the minimum of the site load(s) served and the system size (adjusted for timestep-specific capacity factor) in each period.


Bill-Becker:
@zolanaj I've done some speed testing comparison between non force_dispatch and force_dispatch, and it looks fine to me. I tested a little more force_dispatch for cooling as well, and that looks good (and clarified a note on the test you added for cooling). The one issue I know of is that when the min_ton is 0, the force_dispatch=true run is sizing the ASHP to 3.4 ton for some reason, instead of 0. You mentioned this may be a big M issue. Otherwise, I think this is good to go!

Bill-Becker:
@zolanaj I'm going to register a new version of REopt.jl by early next week with the fixes for leap year and aligning load years across load types (e.g. if electric load year is 2023, shift the heating/cooling CRB load profiles to align weekday/weekend with 2023). It would be great if we could include this in that release as well, so we can have that as the default dispatch for hybrid ASHP.

Bill-Becker (commented):
I think this looks good other than the weird 3.4 ton size when min_ton is zero with force_dispatch.

zolanaj:
> @zolanaj I'm going to register a new version of REopt.jl by early next week with the fixes for leap year and aligning load years across load types (e.g. if electric load year is 2023, shift the heating/cooling CRB load profiles to align weekday/weekend with 2023). It would be great if we could include this in that release as well, so we can have that as the default dispatch for hybrid ASHP.

Thanks @Bill-Becker!  I'm going to investigate the 3.4 ton issue further and follow up by Friday.

zolanaj:
> > @zolanaj I'm going to register a new version of REopt.jl by early next week with the fixes for leap year and aligning load years across load types (e.g. if electric load year is 2023, shift the heating/cooling CRB load profiles to align weekday/weekend with 2023). It would be great if we could include this in that release as well, so we can have that as the default dispatch for hybrid ASHP.
> 
> Thanks @Bill-Becker! I'm going to investigate the 3.4 ton issue further and follow up by Friday.

Further investigation showed: 
- `force_dispatch = true` doesn't start this in isolation (see additional test), though the case you mention 
- I was able to replicate a small ASHP size resulting in something that looked like one of the cases An put together from her previous analysis of ASHP.  However, a slightly different, but still small, system size also results when `force_dispatch = false` and `min_ton=0` (though the size was different).
- Generally we've seen this before, and we've implemented `min_allowable_ton` and the related minimum allowable peak fraction to address this.  If those are higher than 3.4 tons and and this still shows up, that's a bug and please let me know - but in the instance I tried, setting min_allowable_ton to a high value yielded a zero system size.

Assuming this matches your view of the situation, I can merge this evening as long as tests (including new ones) pass.

zolanaj:
@Bill-Becker I'm waiting on the other tests passing from the push, but the updates above make the big-M handling more robust for the ASHP systems and should prevent small system sizes from coming up that violate `min_allowable_ton` rules, whether a very large value is present for `max_ton` or if it's left blank.

Bill-Becker (approved):
I tested and didn't see any issues now. Thanks!

PR #463, opened 2024-12-11
master < develop

adfarth:


Manifest.toml
```diff
@@ -1,20 +1,29 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.8.3"
+julia_version = "1.9.0"
```
adfarth, line 3: @Bill-Becker  do you think there's any issue with updating the Julia version in the manifest? This change is from https://github.com/NREL/REopt.jl/pull/332

zolanaj: We need to update the `[compat]` section of Project.toml if we merge this or there will likely be an error instantiating the package.

zolanaj (commented):
Tests are passing so this may be minor but I want to test this using a fresh install of Julia 1.9 before I approve.  I still think we need to update Project.toml but there might be other changes beyond the julia compat.

zolanaj:
> Tests are passing so this may be minor but I want to test this using a fresh install of Julia 1.9 before I approve. I still think we need to update Project.toml but there might be other changes beyond the julia compat.

This should reconcile Project.toml and Manifest.toml: https://github.com/NREL/REopt.jl/pull/464

Manifest.toml
Bill-Becker: I don't think we ever want to manually edit the Manifest.toml, as it advises against.

adfarth: @Bill-Becker I don't believe this was a manual update, just the result of local machine-generated changes

zolanaj (approved):
This looks good to me with the new merge - thanks @adfarth!

PR #465, opened 2024-12-17
Add Optional Costs for Existing Boiler and Chiller

Bill-Becker:
Add installed cost input options for `ExistingBoiler` and `ExistingChiller` which are also in the BAU scenario, to represent upgrades/replacements needed for those, which can then be avoided with investment in DERs. 

Two cost options are implemented (swap out `_per_mmbtu_per_hour` with `_per_ton` for the analogous ExistingChiller options):
1. `installed_cost_per_mmbtu_per_hour` which estimates the cost based on the `ExistingBoiler` **size needed** for the scenario. We also include the input `max_thermal_factor_on_peak_load` in the size/cost needed (default is 1.25), so that effectively adds e.g. 25% to the size and cost based on the used/needed amount of `ExistingBoiler`. For example, if the BAU case needs 2 MMBtu/hr boiler, but the optimal case only needs 1 MMBtu/hr, the optimal case will have half the cost as the BAU case for the `ExistingBoiler`.
3. `installed_cost_dollars` which is **not dependent on the size needed**, and the optimal scenario will incur this entire cost **if any size/amount of the `ExistingBoiler` is needed still**. So the optimal case needs to **not use any** of the `ExistingBoiler` to avoid this cost. Note, for `GHP` which cannot serve `DomesticHotWaterLoad` by default, the preference may be to use Option 1 so that the incurred `ExistingBoiler` cost is only for serving DHW load. Or, the user can input `installed_cost_dollars` that just represents the installed cost of the boiler(s) for hot water heating.

Note, no incentives or discounting is applied to these installed cost inputs, so they are meant to be in terms of "net present" cost.


Bill-Becker:
> @Bill-Becker can you also update the REopt.jl wiki with these additional inputs (if you haven't already updated it somewhere else that I miss)?

So I think the Go-To place for seeing inputs and outputs would be the REopt.jl documentation page, and since I added the new inputs to the ExistingBoiler and ExistingChiller doc strings (above the structs), they will be viewable once this gets merged in. This is the version of the docs page that I spun up locally to confirm they will be in there:
image

atpham88:
> > @Bill-Becker can you also update the REopt.jl wiki with these additional inputs (if you haven't already updated it somewhere else that I miss)?
> 
> So I think the Go-To place for seeing inputs and outputs would be the REopt.jl documentation page, and since I added the new inputs to the ExistingBoiler and ExistingChiller doc strings (above the structs), they will be viewable once this gets merged in. This is the version of the docs page that I spun up locally to confirm they will be in there: image

I see these updates in the existingboiler.jl now. I originally thought we have to update a separate doc but this is better. Thanks.

PR #469, opened 2025-01-06
Fix leap year (e.g. 2024!) handling of URDB demand and energy charges

Bill-Becker:
Currently, for a custom electric load profile (`ElectricLoad.loads_kw`) input for a leap `year` (e.g. 2024!), we instruct users to truncate the **last day of the year**, but within the code we actually take out the **leap day** for the URDB energy and demand charge schedule. This was causing the following issues for leap years:
1. The load (which includes the 2/29 day) would be shifted forward by one day after 2/28 relative to the energy charges, so weekday/weekend charges would be off periodically.
2. The facility/flat demand charge (`m[DemandFlatCharges]`) timesteps are shifted by a day for each month after February relative to the load - this only influences things if the billed demand is set on the last day of the month.
3. The TOU demand charges (`m[DemandTOUCharges]`) timesteps are shifted by a day after February which can affect the weekday/weekend aspect to the time periods in which the demand charges are set.

This PR fixes these issues by keeping the leap day in the rate tariff timesteps to align with the load, and instead removes the last day of the year to preserve the 8760 hours in the representative year.

### DIFF
--- a/src/core/urdb.jl
+++ b/src/core/urdb.jl
@@ -408,11 +408,11 @@ return Array{Int, 1} for time_steps in ratchet (aka period)
 function get_tou_demand_steps(d::Dict; year::Int, month::Int, period::Int, time_steps_per_hour=1)
     if month > 1
         plus_days = 0
+        if month == 12 && isleapyear(year)
+            plus_days -= 1
+        end
         for m in range(1, stop=month-1)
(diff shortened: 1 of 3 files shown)


Bill-Becker:
Re-naming branch and creating a replacement PR

PR #470, opened 2025-01-07
Fix leap year (e.g. 2024!) handling of URDB demand and energy charges

Bill-Becker:
Currently, for a custom electric load profile (`ElectricLoad.loads_kw`) input for a leap `year` (e.g. 2024!), we instruct users to truncate the **last day of the year**, but within the code we actually take out the **leap day** for the URDB energy and demand charge schedule. This was causing the following issues for leap years:
1. The load (which includes the 2/29 day) would be shifted forward by one day after 2/28 relative to the energy charges, so weekday/weekend charges would be off periodically.
2. The facility/flat demand charge (`m[DemandFlatCharges]`) timesteps are shifted by a day for each month after February relative to the load - this only influences things if the billed demand is set on the last day of the month.
3. The TOU demand charges (`m[DemandTOUCharges]`) timesteps are shifted by a day after February which can affect the weekday/weekend aspect to the time periods in which the demand charges are set.

This PR fixes these issues by keeping the leap day in the rate tariff timesteps to align with the load, and instead removes the last day of the year to preserve the 8760 hours in the representative year.


Bill-Becker:
@adfarth Tests are all passing now, including the new tests I made for leap year fixes. If you want to observe the current/wrong leap year behavior, you could run the tests that I added in the develop branch and see that the rate and load become misaligned/shifted after Feb 28. Let me know if you have any questions! I'll also add an update to the changelog.

test/runtests.jl
```diff
@@ -3045,5 +3045,63 @@ else  # run HiGHS tests
...
+            # TOU Energy charges
+            weekend_rate = input_data["ElectricTariff"]["urdb_response"]["energyratestructure"][2][1]["rate"]  # Not used in this test
+            weekday_rate = input_data["ElectricTariff"]["urdb_response"]["energyratestructure"][3][1]["rate"]
+            energy_charge_expected = weekday_rate * peak_load
+            @test results["ElectricTariff"]["year_one_energy_cost_before_tax"] ≈ energy_charge_expected atol=1E-6
+        
+            # TOU Demand charges
+            flat_rate = input_data["ElectricTariff"]["urdb_response"]["flatdemandstructure"][3][1]["rate"]
+            tou_rate = input_data["ElectricTariff"]["urdb_response"]["demandratestructure"][3][1]["rate"]
+            demand_charge_expected = (flat_rate + tou_rate) * peak_load
```
adfarth, line 3081: @Bill-Becker  it looks like the peak demand is March 4 at 7am, which would be outside of the 2pm-7pm TOU demand rate window. Why then does this test pass (I must be missing something)?

test/runtests.jl
```diff
@@ -3045,5 +3045,63 @@ else  # run HiGHS tests
...
+            @test results["ElectricTariff"]["year_one_demand_cost_before_tax"] ≈ demand_charge_expected atol=1E-6
+        
+            # Test for flat/facility (non-TOU) demand charge for 2023 (non-leap year) and 2024 (leap year)
```
adfarth, line 3084: @Bill-Becker I'm curious why only the flat non-TOU charge is tested for both 2023 and 2024, rather than including the above tests in this for loop?

test/runtests.jl
```diff
@@ -3045,5 +3045,63 @@ else  # run HiGHS tests
...
+            for year in [2023, 2024]
+                input_data["ElectricLoad"]["year"] = year
+                input_data["ElectricLoad"]["loads_kw"] = zeros(8760)
+                # Weekday off-peak February 28th, to set February Facility demand charge
```
adfarth, line 3088: @Bill-Becker  this doesn't impact the test, but whether this and the next timestep are "off-peak" doesn't really matter for this test since the facility charge is non-TOU, right? If so, could consider just changing these comments to avoid confusion

adfarth (approved):
I thoroughly confused myself while reviewing this and thus posted and deleted several comments. But after coming out the other end of that this all looks great! Just a reminder to update the changelog

test/runtests.jl
Bill-Becker: Good point - I'm going to loop across both years for all tests.

PR #472, opened 2025-01-09
Align heating and cooling load profiles with electric load year

Bill-Becker:
**Current issue/bug:** if a custom `ElectricLoad.loads_kw` is input along with its corresponding calendar `year`, and then a heating and/or cooling load is "simulated" by either a commercial reference building (`doe_reference_name`, or `blended_...`) or schedule-based flat load (e.g. `FlatLoad_16_5`) which uses the 2017 year (start day == Sunday), then the load profiles will **not be aligned** (shifted in the weekdays/weekends). See attached images for current web tool behavior with uploaded electric load for 2024 (starts on Monday) with heating load (starts on Sunday).

This PR fixes this by passing the `ElectricLoad.year` to the constructors for `HeatingLoad` and `CoolingLoad`, and it **shifts the simulated CRB or schedule-based flat load profiles to match the starting day of the year input**.

Also, now that we have a custom load profile builder for heating loads in the REopt web tool, in which you build an 8760 heating load profile and assign a particular `year` to it, custom/uploaded heating loads will be used more often. This PR ensures that the year input for heating load profile matches the year input for the electric load profile input. It throws an error if it doesn't match.

Side note: we kind of got away with this for the last year because 2023 (the previous full year) starts on the same day as 2017, but now that 2024 is the most recent full historical year, I suspect this will start to become an issue more.

"Uploaded" 2024 (starts on Monday) Electric Load:
Uploaded 2024 Electric load with FlatLoad_16_5 profile

"Simulate Building" 16/5 Schedule Flat Load (2017 is the only option, starts on Sunday):
Heating FlatLoad_16_5 profile with uploaded Electric load for 2024


Bill-Becker:
@adfarth I'm not sure why it only says "1 successful check" (Documentation) under "All checks have passed" - the latest runtests tests all passed, as you can see in Actions.

src/core/heating_cooling_loads.jl
```diff
@@ -49,7 +54,7 @@ function HeatingLoad(;
     blended_industrial_reference_names::Array{String, 1} = String[],
     blended_industrial_reference_percents::Array{<:Real,1} = Real[],    
     city::String = "",
-    year::Int = doe_reference_name ≠ "" || blended_doe_reference_names ≠ String[] ? 2017 : 2022, # CRB profiles must use 2017. If providing load profile, specify year of data.
+    year::Int = doe_reference_name ≠ "" || blended_doe_reference_names ≠ String[] ? 2017 : 2022, # CRB profiles are 2017 by default. If providing load profile, specify year of data.
```
adfarth, line 57: @Bill-Becker thoughts on changing the electric_load.jl description for `year` to align with this description?

src/core/heating_cooling_loads.jl
```diff
@@ -9,11 +9,16 @@
...
-    addressable_load_fraction::Any = 1.0,  # Fraction of input fuel load which is addressable by heating technologies. Can be a scalar or vector with length aligned with use of monthly_mmbtu or fuel_loads_mmbtu_per_hour.
+    city::String = "",
+    year::Int = doe_reference_name ≠ "" || blended_doe_reference_names ≠ String[] ? 2017 : 2022, # CRB profiles are 2017 by default. If providing load profile, specify year of data.    
     annual_mmbtu::Union{Real, Nothing} = nothing,
     monthly_mmbtu::Array{<:Real,1} = Real[],
+    addressable_load_fraction::Any = 1.0,  # Fraction of input fuel load which is addressable by heating technologies. Can be a scalar or vector with length aligned with use of monthly_mmbtu or fuel_loads_mmbtu_per_hour.
     fuel_loads_mmbtu_per_hour::Array{<:Real,1} = Real[], # Vector of space heating fuel loads [mmbtu/hr]. Length must equal 8760 * `Settings.time_steps_per_hour`
     normalize_and_scale_load_profile_input::Bool = false,  # Takes fuel_loads_mmbtu_per_hour and normalizes and scales it to annual or monthly energy
+    time_steps_per_hour::Int = 1, # corresponding to `fuel_loads_mmbtu_per_hour`
```
adfarth, line 19: It looks like at least latitude and longitude are always fed into the `HeatingLoad()` function, whereas I've always considered this help text section to be just for user inputs. Do we want users to be able to specify a different lat and long here than provided in Site (and would it get used if specified?)? 

Similarly, just to confirm: do we want users to separately specify a time_steps_per hour here? It looks like in most instances this does get passed into the function from Settings.

adfarth:
Per the dev call, adding a note to check that the different load years do indeed get sent when using the web tool.

adfarth:
> @adfarth I'm not sure why it only says "1 successful check" (Documentation) under "All checks have passed" - the latest runtests tests all passed, as you can see in Actions.

I think this happens when the latest commit is only to the changelog

src/core/electric_load.jl
```diff
@@ -158,19 +158,13 @@ mutable struct ElectricLoad  # mutable to adjust (critical_)loads_kw based off o
             end
```
adfarth, line 158: Adding a note here to require the year if a custom load profile is updated (here and for the heating loads)

src/core/heating_cooling_loads.jl
adfarth: I made this small change

src/core/heating_cooling_loads.jl
Bill-Becker: Agreed, removing from the doc string

src/core/electric_load.jl
Bill-Becker: Added this requirement. Validation to check if year is not nothing can happen above the check for loads_kw, doe_reference_name, etc, since the default is assigned for CRB without a year input.

test/runtests.jl
```diff
@@ -1909,6 +1913,7 @@ else  # run HiGHS tests
             post_name = "wind_intl_offgrid.json" 
             post = JSON.parsefile("./scenarios/$post_name")
             post["ElectricLoad"]["loads_kw"] = [10.0 for i in range(1,8760)]
+            post["ElectricLoad"]["year"] = 2022
             scen = Scenario(post)
             post["Wind"]["production_factor_series"] =  reduce(vcat, readdlm("./data/example_wind_prod_factor_kw.csv", '\n', header=true)[1])
```
adfarth, line 1919: Down below there's a comment "# This get_monthly_energy function is only equivalent for non-leap years with loads_kw normalization and scaling because it removes the leap day from the processing of monthly hours/energy". Does this just need to be updated to say removes last day of year?

test/runtests.jl
```diff
@@ -3112,7 +3117,84 @@ else  # run HiGHS tests
                 end
```
adfarth, line 3117: I didn't catch this before but suggest removing println("year = ", year)

src/core/doe_commercial_reference_building_loads.jl
```diff
@@ -134,25 +134,32 @@ function built_in_load(
...
         else 
             normalized_profile = vec(readdlm(profile_path, '\n', Float64, '\n'))
+            shift_possible = true
         end
     else
         input_normalized = true
     end
 
+    # The normalized_profile for CRBs (not FlatLoads) is based on year 2017 which starts on a Sunday. 
+    # If the year is not 2017 and we're using a CRB, we need to shift the profile to make the first day of the profile equal to the first day of the input year.
```
adfarth, line 150: @Bill-Becker It's possible I'm wrong on this, but I think this needs a slightly different formulation. 

Currently: 
EX: Year = 2021. Start day = Fri (5)
        first_day_of_year = 5
        shift_days = 5- 7 = -2
        normalized_profile = circshift(normalized_profile, 24 * -2)
        ^^ Shifts the first two days (Sun Mon) to the end (moves vector left). So now we are starting on Tues (rather than Friday)

I think this should instead be: 

```
shift_days = (7 - first_day_of_year) % 7 # mod 7 makes it so if first_day_of_year = 7 this is 0
```

src/core/electric_load.jl
adfarth: @Bill-Becker I think it might be worth adding this to the CHANGELOG too

test/runtests.jl
Bill-Becker: Removed

src/core/electric_load.jl
Bill-Becker: Added to changelog.

CHANGELOG.md
```diff
@@ -25,6 +25,12 @@ Classify the change according to the following categories:
     ### Deprecated
     ### Removed
 
+## load-year-align
+### Fixed
+- Align heating and cooling load profiles based on electric load year input, if using custom electric load profile with simulated (CRB or schedule-based flatloads) heating/cooling loads
+### Changed
+- Make `year` input required with `ElectricLoad.loads_kw` input
```
adfarth, line 32: Also for the custom heating and cooling load inputs, correct? Could you add here?

src/core/heating_cooling_loads.jl
```diff
@@ -9,11 +9,13 @@
     industrial_reference_name::String = "",  # For ProcessHeatLoad
     blended_industrial_reference_names::Array{String, 1} = String[],  # For ProcessHeatLoad
     blended_industrial_reference_percents::Array{<:Real,1} = Real[],  # For ProcessHeatLoad
-    addressable_load_fraction::Any = 1.0,  # Fraction of input fuel load which is addressable by heating technologies. Can be a scalar or vector with length aligned with use of monthly_mmbtu or fuel_loads_mmbtu_per_hour.
+    city::String = "",
+    year::Union{Int, Nothing} = doe_reference_name ≠ "" || blended_doe_reference_names ≠ String[] ? 2017 : nothing, # CRB profiles are 2017 by default. If providing load profile, specify year of data.
```
adfarth, line 13: Does this year check for defaulting to 2017 also need to account for `industrial_reference_name` and `blended_industrial_reference_names`? I'm not familiar with exactly how those are used.

src/core/scenario.jl
```diff
@@ -221,57 +221,57 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
...
                                         load_type = "domestic_hot_water",
                                         latitude=site.latitude, longitude=site.longitude, 
                                         time_steps_per_hour=settings.time_steps_per_hour,
-                                        existing_boiler_efficiency = existing_boiler_efficiency,
-                                        year = year
+                                        existing_boiler_efficiency = existing_boiler_efficiency
                                         )
         max_heat_demand_kw = maximum(dhw_load.loads_kw)
     else
+        # TODO need to add a dummy year here or else will error
```
adfarth, line 234: @Bill-Becker is this still a pending TODO?

adfarth:
@Bill-Becker do you think it matters that in `ElectricTariff` the year is still defaulted to 2022?

test/runtests.jl
```diff
@@ -3112,7 +3114,71 @@ else  # run HiGHS tests
...
+            """
+            Common use case: ElectricLoad.loads_kw is input with specific year, but heating and/or cooling is 
+                simulated with either a schedule-based FlatLoad, or b) CRB type with annual or monthly energy
+            This test confirms that the simulated FlatLoad type and CRB are shifted to start on Monday for 2024
+            """
         
+            input_data = JSON.parsefile("./scenarios/load_year_align.json")
+            year = 2024
+            # ElectricLoad.loads_kw is 2024, and heating and cooling loads are shifted to align
+            # Use a FlatLoad_16_5 shifted to 2024 (Monday start) with the web tool's custom load builder
```
adfarth, line 3129: this isn't that important, but what does this comment mean? That the flat load has already been shifted after getting the 2017 version from the web tool?

adfarth (approved):
@Bill-Becker just a few small questions and asks, but I think this is good to merge after addressing those. These are great fixes!

CHANGELOG.md
Bill-Becker: Updated the line to include any load profile input.

src/core/scenario.jl
Bill-Becker: Nope, done.

src/core/heating_cooling_loads.jl
Bill-Becker: Good catch, I'll add that too.

test/runtests.jl
Bill-Becker: I just built a load with the year=2024, and the web tool aligned the weekday/weekend profiles to the 2024 year.

Bill-Becker:
> @Bill-Becker do you think it matters that in `ElectricTariff` the year is still defaulted to 2022?

Removed the year default.

PR #474, opened 2025-01-21
Set ASHP forced dispatch to true by default

zolanaj:
### Changed
- Set default value of `force_dispatch` to `true` in ASHP technologies


Bill-Becker:
@zolanaj I added one significant thing: change default `min_allowable_peak_capacity_fraction` to 0.25, from 0.5. Let me know if you have any issue with this. My reasoning is that now that we're forcing dispatch by default, a smaller size system will produce a good amount more thermal. Also, for the LargeOffice building in particular, an ASHP system sized at 25% of peak load is serving 75%+ of the heating load.

Bill-Becker:
@zolanaj I have a branch of the API with the added force_dispatch input, and I'll change the default `min_allowable_peak_capacity_fraction` there too which is in the clean() method of the two ASHP tech Inputs classes.

zolanaj:
> @zolanaj I added one significant thing: change default `min_allowable_peak_capacity_fraction` to 0.25, from 0.5. Let me know if you have any issue with this. My reasoning is that now that we're forcing dispatch by default, a smaller size system will produce a good amount more thermal. Also, for the LargeOffice building in particular, an ASHP system sized at 25% of peak load is serving 75%+ of the heating load.

@Bill-Becker no issues with this at all, and thanks for catching the docstrings miss.  I didn't alter the changelog in part because if this merges with develop before the new version is registered then there shouldn't be a need (i.e., we're editing something that we just added in the same version).

zolanaj:
> @zolanaj I have a branch of the API with the added force_dispatch input, and I'll change the default `min_allowable_peak_capacity_fraction` there too which is in the clean() method of the two ASHP tech Inputs classes.

Thanks for leading the change! Let me know if I can help with the API side or review.

Bill-Becker (approved):
Assuming the change of `min_allowable_peak_capacity_fraction` didn't break tests, I approve as is.

Bill-Becker:
@zolanaj The API branch is reoptjl-v050. Can you check to make sure I am handling `force_dispatch` and the update to `min_allowable_peak_capacity_fraction` for both the models.py -> validation and communicating defaults in the /get_ashp_defaults endpoint?

PR #476, opened 2025-02-05
DC coupled PV and ElectricStorage

hdunham:
TODO:
- bit more documentation
- move test into runtests.jl
- check on consistency of using grid_charge_efficiency vs rectifier_efficiency_fraction

### Added
- Add option to model DC-coupled **PV** and **ElectricStorage**
    - Add **PV** input **dc_coupled_with_storage** and **ElectricStorage** input **dc_coupled**, which both default to _false_
    - **ElectricStorage** inputs **inverter_efficiency_fraction** and **rectifier_efficiency_fraction** define the DC-coupled inverter and **PV** input **inf_eff** is ignored
    - Add sets in **Techs** struct (**dc_coupled_with_storage** and **ac_coupled_with_storage**) and **StorageTypes** struct (**dc_coupled** and **ac_coupled**)
    - Modify constraints to account for AC and DC coupling
### Removed
- Don't expose **ElectricStorage** fields **charge_efficiency**, **discharge_efficiency**, and **grid_charge_efficiency**. It can lead to unexpected behavior when users provide these instead of the inputs they are calculated from (**inverter_efficiency_fraction**, **rectifier_efficiency_fraction**, and **internal_efficiency_fraction**).


CHANGELOG.md
```diff
@@ -25,6 +25,15 @@ Classify the change according to the following categories:
     ### Deprecated
     ### Removed
 
+## new-dc-couple
+### Added
+- Add option to model DC-coupled **PV** and **ElectricStorage**
+    - Add **PV** input **dc_coupled_with_storage** and **ElectricStorage** input **dc_coupled**, which both default to _false_
```
adfarth, line 31: @hdunham I haven't looked through the code yet, so this maybe become more clear to me, but I'm a bit confused initially by these inputs. 

Why not just have one input for dc_coupled pv and storage rather that two? It seems like having two could allow these inputs to conflict (if one is true and the other is false, what happens?). The changes should only ever have an impact if you model both PV and BESS together; if only one or the other is included in the model, the results should just use the typical inverter setup, right?

src/core/energy_storage/electric_storage.jl
```diff
@@ -274,6 +273,10 @@ struct ElectricStorage <: AbstractElectricStorage
     function ElectricStorage(d::Dict, f::Financial)  
         s = ElectricStorageDefaults(;d...)
 
+        if s.dc_coupled && s.off_grid_flag
```
adfarth, line 276: Curious what the main limitation is here? It seems to me like off-grid is the biggest use case for this!

src/constraints/electric_utility_constraints.jl
```diff
@@ -1,13 +1,6 @@
 # REopt®, Copyright (c) Alliance for Sustainable Energy, LLC. See also https://github.com/NREL/REopt.jl/blob/master/LICENSE.
 function add_export_constraints(m, p; _n="")
 
-    ##Constraint (8e): Production export and curtailment no greater than production
```
adfarth, line 4: Is not needed or was it moved elsewhere?

src/constraints/storage_constraints.jl
```diff
@@ -94,34 +96,34 @@ function add_elec_storage_dispatch_constraints(m, p, b; _n="")
...
-	#Constraint (4l)-alt: Dispatch from electrical storage is no greater than power capacity (no grid connection)
-	@constraint(m, [ts in p.time_steps_without_grid],
-        m[Symbol("dvStoragePower"*_n)][b] >= m[Symbol("dvDischargeFromStorage"*_n)][b,ts] + 
-            sum(m[Symbol("dvProductionToStorage"*_n)][b, t, ts] for t in p.techs.elec)
-    )
+    if b in p.s.storage.types.dc_coupled
+        @constraint(m, [ts in p.time_steps],
+            m[Symbol("dvStoragePower"*_n)][b] >= m[Symbol("dvDischargeFromStorage"*_n)][b, ts] / p.s.storage.attr[b].inverter_efficiency_fraction
+                + sum(m[Symbol("dvProductionToStorage"*_n)][b, t, ts] for t in p.techs.ac_coupled_with_storage) * p.s.storage.attr[b].rectifier_efficiency_fraction
+                + sum(m[Symbol("dvProductionToStorage"*_n)][b, t, ts] for t in p.techs.dc_coupled_with_storage)
```
adfarth, line 117: Shouldn't the techs in `techs.dc_coupled_with_storage` not be added here? If the electricity is going directly to storage wouldn't it not need to pass through the inverter?

src/constraints/storage_constraints.jl
```diff
@@ -94,34 +96,34 @@ function add_elec_storage_dispatch_constraints(m, p, b; _n="")
...
     )
-
-	# Constraint (4i)-1: Dispatch to electrical storage is no greater than power capacity
-	@constraint(m, [ts in p.time_steps],
-        m[Symbol("dvStoragePower"*_n)][b] >= 
-            sum(m[Symbol("dvProductionToStorage"*_n)][b, t, ts] for t in p.techs.elec) + m[Symbol("dvGridToStorage"*_n)][b, ts]
-    )
 	
 	#Constraint (4k)-alt: Dispatch to and from electrical storage is no greater than power capacity
-	@constraint(m, [ts in p.time_steps_with_grid],
```
adfarth, line 115: I sometimes get confused about use of time_steps vs time_steps_with_grid and without_grid. It seems like here you have consolidated just into time_steps and removed constraint (4i)-1. Did you conclude that these were duplicative?

src/core/energy_storage/electric_storage.jl
hdunham: only that I ran out of time to modify the off grid function add_operating_reserve_constraints :(

src/constraints/storage_constraints.jl
hdunham: Yeah I condensed three constraints ((4i)-1, (4k)-alt, (4l)-alt) into one. Constraint (4l)-alt could be removed by changing (4k)-alt to use time_steps because the dvGridToStorage term would just be zero for p.time_steps_without_grid anyway. Then constraint (4i)-1 could be removed because dvDischargeFromStorage >= 0 so (4k)-alt will always be more restrictive.

src/constraints/storage_constraints.jl
hdunham: This is a constraint on dvStoragePower, which is the power of the actual battery. The inverter size DV is dvDCCoupledTechStorageInverterSizeDC.

CHANGELOG.md
adfarth: Is your thought that with multiple storages you might want to be able to choose some with which PV is DC coupled while some have their own separate inverters? Let's discuss on our call

src/core/energy_storage/electric_storage.jl
```diff
@@ -168,7 +169,7 @@ end
     soc_min_fraction::Float64 = 0.2
     soc_min_applies_during_outages::Bool = false
     soc_init_fraction::Float64 = off_grid_flag ? 1.0 : 0.5
-    can_grid_charge::Bool = off_grid_flag ? false : true
+    can_grid_charge::Bool = off_grid_flag || dc_coupled ? false : true
```
adfarth, line 172: Have you tested that allowing grid to charge bess does work when dc coupled?

src/core/energy_storage/electric_storage.jl
```diff
@@ -261,6 +258,8 @@ struct ElectricStorage <: AbstractElectricStorage
     total_rebate_per_kwh::Real
     charge_efficiency::Float64
```
adfarth, line 259: could maybe be worth renaming to ac_dc_charge_efficiency and dc_ac_discharge efficiency? Or if not, at least adding a comment here

src/core/energy_storage/electric_storage.jl
```diff
@@ -329,7 +332,14 @@ struct ElectricStorage <: AbstractElectricStorage
             degr = Degradation()
         end
     
+        charge_efficiency = s.rectifier_efficiency_fraction * s.internal_efficiency_fraction^0.5
+        discharge_efficiency = s.inverter_efficiency_fraction * s.internal_efficiency_fraction^0.5
+        dc_charge_efficiency = s.internal_efficiency_fraction^0.5
+        dc_discharge_efficiency = s.internal_efficiency_fraction^0.5
```
adfarth, line 338: Should `dc_discharge_efficiency` ever be used if we don't have dc loads? Wouldn't all discharge need to run through the inverter? It looks like this is only used in the renewable_energy_constraints currently

src/core/pv.jl
```diff
@@ -44,6 +44,7 @@
     can_export_beyond_nem_limit::Bool = off_grid_flag ? false : true,
     can_curtail::Bool = true,
     operating_reserve_required_fraction::Real = off_grid_flag ? 0.25 : 0.0, # if off grid, 25%, else 0%. Applied to each time_step as a % of PV generation.
+    dc_coupled_with_storage::Bool = false
```
adfarth, line 47: could you add a short description here?

test/debugging_tests.jl
```diff
@@ -0,0 +1,172 @@
+# REopt®, Copyright (c) Alliance for Sustainable Energy, LLC. See also https://github.com/NREL/REopt.jl/blob/master/LICENSE.
```
adfarth, line 1: could be worth adding this to the changelog and explaining how it should be used?

test/runtests.jl
```diff
@@ -23,6 +27,7 @@ elseif "CPLEX" in ARGS
     @testset "test_with_cplex" begin
         include("test_with_cplex.jl")
     end
+
```
adfarth, line 30: @hdunham  do you plan to add tests here?

src/core/pv.jl
adfarth: I think we should also add a markdown note on costs when using dc_coupled_with_storage. Users should be aware that our default PV costs include inverts and the O&M costs account for inverter replacements. Similarly, the default BESS per_kw costs don't account for any added costs from it being a hybrid inverter

src/results/electric_storage.jl
```diff
@@ -23,7 +23,10 @@ function add_electric_storage_results(m::JuMP.AbstractModel, p::REoptInputs, d::
     r = Dict{String, Any}()
     r["size_kwh"] = round(value(m[Symbol("dvStorageEnergy"*_n)][b]), digits=2)
     r["size_kw"] = round(value(m[Symbol("dvStoragePower"*_n)][b]), digits=2)
-
+    if b in p.s.storage.types.dc_coupled
+        r["dc_coupled_inverter_size_kw"] = round(value(m[:dvDCCoupledTechStorageInverterSizeDC][b]), digits=2)
```
adfarth, line 27: from what little I've seen, it seems like these inverters are commonly rated by their AC output, but these are kWDC units, right? Would it make sense to convert to AC sizing? If not, I think we should at least specify the units in the name since they differ from the AC units of the standard `size_kW`

CHANGELOG.md
hdunham: Yes exactly

src/core/energy_storage/electric_storage.jl
hdunham: Do you mean overriding this default and setting can_grid_charge=true for a dc coupled scenario? I haven't made a test for this specifically that e.g. checks the result of grid to storage var. I was originally going to not allow grid charging when dc coupled but then switched to this, so should probably create that test now.

src/core/energy_storage/electric_storage.jl
hdunham: Yeah that's a good thought. Maybe just ac_charge_efficiency and ac_charge_efficiency though? Cause it's dc_charge_efficiency not dc_dc_charge_efficiency.

src/core/energy_storage/electric_storage.jl
hdunham: True, don't need it. I'll delete.

src/results/electric_storage.jl
hdunham: Oh thanks! I wasn't sure whether the inverter should be defined by the ac or dc side so I just picked one and I meant to come back to that decision. I will change to dvDCCoupledTechStorageInverterSizeAC.

test/debugging_tests.jl
hdunham: I wasn't planning on leaving this. I was going to migrate this test over to runtests.jl.

src/results/electric_storage.jl
hdunham: done in commit 6214f80a

src/core/energy_storage/electric_storage.jl
adfarth: Oh maybe it's worth specifying in all of the names the from and two units. Like having all of these be dc_dc, ac_dc, dc_ac so that it's clear

src/core/scenario.jl
```diff
@@ -183,6 +184,13 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
         storage_structs["ColdThermalStorage"] = ColdThermalStorage(params, financial, settings.time_steps_per_hour)
     end
     storage = Storage(storage_structs)
+    dc_coupled_pv = any([pv.dc_coupled_with_storage && pv.max_kw > 0 for pv in pvs])
+    if dc_coupled_pv && isempty(storage.types.dc_coupled)
```
adfarth, line 188: I do think just defaulting this storage attribute to true and not exposing it as a user input might be preferable, so that a user doesn't have to change two inputs to do one thing.

src/core/scenario.jl
```diff
@@ -183,6 +184,13 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
...
+        throw(@error("To model PV that is DC coupled with storage (PV input dc_coupled_with_storage = true), you must also model a non-zero DC coupled storage system (set ElectricStorage input dc_coupled = true)"))
+    end
+    if !isempty(storage.types.dc_coupled) && !dc_coupled_pv
+        throw(@error("To model storage that is DC coupled with PV (ElectricStorage input dc_coupled = true), you must also model a non-zero DC coupled PV system (set PV input dc_coupled_with_storage = true)"))
+    end
```
adfarth, line 193: Does there need to be a check that if you're modeling multiple PVs, `dc_coupled_with_storage` is true for all of them?

src/constraints/load_balance.jl
```diff
@@ -43,21 +55,33 @@ function add_elec_load_balance_constraints(m, p; _n="")
...
             ==
-            sum(sum(m[Symbol("dvProductionToStorage"*_n)][b, t, ts] for b in p.s.storage.types.elec) 
-                + m[Symbol("dvCurtail"*_n)][t, ts] for t in p.techs.elec)
-            + p.s.electric_load.critical_loads_kw[ts]
+            p.s.electric_load.critical_loads_kw[ts]
         )
     else # load balancing constraint for off-grid runs 
         @constraint(m, [ts in p.time_steps_without_grid],
-            sum(p.production_factor[t,ts] * p.levelization_factor[t] * m[Symbol("dvRatedProduction"*_n)][t,ts] for t in p.techs.elec)
+            sum(p.production_factor[t, ts] * p.levelization_factor[t] * m[Symbol("dvRatedProduction"*_n)][t,ts]
```
adfarth, line 73: is this just partially set up for off-grid runs for if you have time to finish this in the future (with the OR constraints)?

src/constraints/outage_constraints.jl
```diff
@@ -3,9 +3,16 @@ function add_dv_UnservedLoad_constraints(m,p)
...
-        - sum(  m[:dvMGRatedProduction][t, s, tz, ts] * (p.production_factor[t, time_step_wrap_around(tz+ts-1, time_steps_per_hour=p.s.settings.time_steps_per_hour)] + p.unavailability[t][time_step_wrap_around(tz+ts-1, time_steps_per_hour=p.s.settings.time_steps_per_hour)]) * p.levelization_factor[t]
-              - m[:dvMGProductionToStorage][t, s, tz, ts] - m[:dvMGCurtail][t, s, tz, ts]
-            for t in p.techs.elec
+        - sum( m[:dvMGRatedProduction][t, s, tz, ts] * (p.production_factor[t, time_step_wrap_around(tz+ts-1, time_steps_per_hour=p.s.settings.time_steps_per_hour)] + p.unavailability[t][time_step_wrap_around(tz+ts-1, time_steps_per_hour=p.s.settings.time_steps_per_hour)]) * p.levelization_factor[t]
+              - m[:dvMGProductionToStorage][t, s, tz, ts]
+              - m[:dvMGCurtail][t, s, tz, ts]
+            for t in p.techs.ac_coupled_with_storage
+        )
+        - sum( p.s.storage.attr["ElectricStorage"].inverter_efficiency_fraction * ( # convert DC to AC
+                m[:dvMGRatedProduction][t, s, tz, ts] * p.production_factor[t, tz+ts-1] * p.levelization_factor[t]
```
adfarth, line 12: why don't these need the `time_step_wrap_around` function in the indices?

src/constraints/outage_constraints.jl
```diff
@@ -76,7 +83,8 @@ function add_outage_cost_constraints(m,p)
         @constraint(m,
             m[:dvMGStorageUpgradeCost] >= p.s.financial.microgrid_upgrade_cost_fraction * m[:TotalStorageCapCosts] - (
                 p.s.financial.microgrid_upgrade_cost_fraction * p.third_party_factor * (
-                    sum( p.s.storage.attr[b].net_present_cost_per_kw * p.s.storage.attr[b].max_kw for b in p.s.storage.types.elec) + 
+                    sum( p.s.storage.attr[b].net_present_cost_per_kw * p.s.storage.attr[b].max_kw for b in p.s.storage.types.ac_coupled) + 
+                    sum( p.s.storage.attr[b].net_present_cost_per_kw * (p.s.storage.attr[b].max_kw + sum(p.max_sizes[t] * maximum(p.production_factor[t,:]) * p.levelization_factor[t] for t in p.techs.dc_coupled_with_storage)) for b in p.s.storage.types.dc_coupled) +
```
adfarth, line 87: for a dc-coupled run, wouldn't `storage.attr[b].max_kw` already capture the total output of the inverter? I'm confused about why `+ sum(p.max_sizes[t] * maximum(p.production_factor[t,:]) * p.levelization_factor[t]` is needed here

src/constraints/outage_constraints.jl
```diff
@@ -302,23 +312,20 @@ function add_MG_storage_dispatch_constraints(m,p)
...
+            m[:dvStoragePower]["ElectricStorage"] >= m[:dvMGDischargeFromStorage][s, tz, ts] / p.s.storage.attr["ElectricStorage"].inverter_efficiency_fraction
+                + sum(m[:dvMGProductionToStorage][t, s, tz, ts] for t in p.techs.ac_coupled_with_storage) * p.s.storage.attr["ElectricStorage"].rectifier_efficiency_fraction
+                + sum(m[:dvMGProductionToStorage][t, s, tz, ts] for t in p.techs.dc_coupled_with_storage)
+        )
+    else
+        @constraint(m, [s in p.s.electric_utility.scenarios, tz in p.s.electric_utility.outage_start_time_steps, ts in p.s.electric_utility.outage_time_steps],
+            m[:dvStoragePower]["ElectricStorage"] >= m[:dvMGDischargeFromStorage][s, tz, ts]
+                + sum(m[:dvMGProductionToStorage][t, s, tz, ts] for t in p.techs.ac_coupled_with_storage)
+                + sum(m[:dvMGProductionToStorage][t, s, tz, ts] for t in p.techs.dc_coupled_with_storage) * p.s.storage.attr["ElectricStorage"].inverter_efficiency_fraction #TODO: remove term or leave for future flexibility?
```
adfarth, line 326: If "ElectricStorage" in **not** in `p.s.storage.types.dc_coupled`, why would we need this here?

src/constraints/outage_constraints.jl
```diff
@@ -302,23 +312,20 @@ function add_MG_storage_dispatch_constraints(m,p)
...
-    
-    # Dispatch to and from electrical storage is no greater than power capacity
-    @constraint(m, [s in p.s.electric_utility.scenarios, tz in p.s.electric_utility.outage_start_time_steps, ts in p.s.electric_utility.outage_time_steps],
-        m[:dvStoragePower]["ElectricStorage"] >= m[:dvMGDischargeFromStorage][s, tz, ts]
-            + sum(m[:dvMGProductionToStorage][t, s, tz, ts] for t in p.techs.elec)
-    )
+    # Dispatch to and from MG electrical storage is no greater than power capacity
+    if "ElectricStorage" in p.s.storage.types.dc_coupled
+        @constraint(m, [s in p.s.electric_utility.scenarios, tz in p.s.electric_utility.outage_start_time_steps, ts in p.s.electric_utility.outage_time_steps],
+            m[:dvStoragePower]["ElectricStorage"] >= m[:dvMGDischargeFromStorage][s, tz, ts] / p.s.storage.attr["ElectricStorage"].inverter_efficiency_fraction
```
adfarth, line 318: I think I'm confused about why the if statement is needed here, whereas in many of the other constraints you were able to avoid if statements by adding flows indexed on ac_coupled to flows index on dc_coupled. And, why does the discharge from storage need to be divided by `inverter_efficiency_fraction` for the dc-coupled case? Shouldn't the discharge from storage be treated the same between ac- and dc-coupled cases, because in both cases the discharge passes through an inverter (and isn't that already captured in `dvMGDischargeFromStorage`)? 

I could be misunderstanding some of this, but just checking!

src/constraints/outage_constraints.jl
```diff
@@ -342,8 +349,32 @@ function add_MG_storage_dispatch_constraints(m,p)
             m[:dvMGDischargeFromStorage][s, tz, ts] <= p.s.storage.attr["ElectricStorage"].max_kw * m[:binMGStorageUsed]
         )
     end
-end
 
+    # Future development could make this an option by adding bool inputs an updating load balance/battery dispatch constraints to include these flows
```
adfarth, line 353: Could be helpful to clarify what the bool inputs would be (probably obvious to you)

src/constraints/outage_constraints.jl
```diff
@@ -342,8 +349,32 @@ function add_MG_storage_dispatch_constraints(m,p)
...
+    if "ElectricStorage" in p.s.storage.types.dc_coupled
+        # Don't let AC coupled elec techs charge DC coupled battery. 
+        for ts in 1:p.s.site.min_resil_time_steps
```
adfarth, line 356: So, if you have a run with wind (AC production) and PV that is dc-coupled with storage, the wind cannot charge the battery? Do you have some test cases in this realm in runtests?

src/constraints/outage_constraints.jl
```diff
@@ -342,8 +349,32 @@ function add_MG_storage_dispatch_constraints(m,p)
...
+            for tz in p.s.electric_utility.outage_start_time_steps
+                for s in p.s.electric_utility.scenarios
+                    for t in p.techs.ac_coupled_with_storage
+                        fix(m[:dvMGProductionToStorage][t, s, tz, ts], 0.0, force=true)
+                    end
+                end
+            end
+        end
+    else
```
adfarth, line 365: would this ever be reached, currently? For this else statement to activate you'd need to have ElectricStorage dc_coupled be false, but have the dc_coupled_with_storage tech set be not empty?

src/core/scenario.jl
hdunham: Nope, you can model a mixture!

src/core/scenario.jl
hdunham: Ok sounds good

src/core/scenario.jl
adfarth: oh cool! How does this work if you're only modeling a single storage system? Or you would have to model multiple storage systems to do this?

src/constraints/storage_constraints.jl
```diff
@@ -77,51 +77,53 @@ function add_elec_storage_dispatch_constraints(m, p, b; _n="")
...
-    )
-
-	#Constraint (4l)-alt: Dispatch from electrical storage is no greater than power capacity (no grid connection)
-	@constraint(m, [ts in p.time_steps_without_grid],
-        m[Symbol("dvStoragePower"*_n)][b] >= m[Symbol("dvDischargeFromStorage"*_n)][b,ts] + 
-            sum(m[Symbol("dvProductionToStorage"*_n)][b, t, ts] for t in p.techs.elec)
-    )
+    if b in p.s.storage.types.dc_coupled
+        @constraint(m, [ts in p.time_steps],
+            m[Symbol("dvStoragePower"*_n)][b] >= m[Symbol("dvDischargeFromStorage"*_n)][b, ts] / p.s.storage.attr[b].inverter_efficiency_fraction
```
adfarth, line 115: Could you explain when `inverter_efficiency_fraction` gets used instead of `ac_discharge_efficiency`. The latter is used in the constraint above on dvStorageEnergy and the former is used here

src/constraints/storage_constraints.jl
adfarth: oh yes thank you I should have realized that

src/constraints/storage_constraints.jl
```diff
@@ -137,23 +139,60 @@ function add_elec_storage_dispatch_constraints(m, p, b; _n="")
             sum(m[Symbol("dvStorageEnergy"*_n)][b])
         )
     end
+
+    # Future development could make this an option by adding bool inputs an updating load balance/battery dispatch constraints to include these flows
+    if b in p.s.storage.types.dc_coupled
```
adfarth, line 144: Same comment from me here as in "outage_constraints.jl"

adfarth:
General comment on things to document: 
- AC-producing techs cannot currently charge a BESS that is DC coupled with PV. Maybe make a note on what dev would be needed to change this? 
- Notes in code on whether dv's are DC or AC units
- Description of overall configuration as it compares to current config (PV and BESS have separate inverter systems)
- Notes on how to consider modifying costs (lower PV costs, raise ElectricStorage costs when DC-Coupled)

src/constraints/outage_constraints.jl
hdunham: added

src/constraints/outage_constraints.jl
hdunham: storage.attr[b].max_kw is just the kw of the battery itself, the inverter kw is a separate DV that could be larger

src/constraints/outage_constraints.jl
hdunham: This is flow from dc coupled techs out inverter to ac coupled storage, which is currently is not allowed. With multiple storage and adding a bool input to say whether this flow is allowed, this term would be relevant, which is why I wrote that TODO.

src/constraints/outage_constraints.jl
hdunham: Yes that's the current set up. Could add an input bool to allow this though. Noting this as a test to add.

src/constraints/storage_constraints.jl
hdunham: ac_discharge_efficiency includes the internal efficiency so that is used for state of charge equations

test/runtests.jl
hdunham: yes they are just in the debugging_tests.jl file right now

PR #479, opened 2025-02-21
Fix tech sets for RE fraction parameters

zolanaj:
### Added
- Added a warning that when **SteamTurbine** is included, renewable energy fractions may not be accurate.
- Added new attribute **fuel_renewable_energy_fraction** to the technology **Boiler**

### Fixed
- Updated the **REoptInputs** parameter **tech_renewable_energy_fraction** so that only electricity-producing and fuel-burning heating technologies are included (instead of all technologies).


src/core/bau_inputs.jl
```diff
@@ -32,7 +32,7 @@ function BAUInputs(p::REoptInputs)
     cooling_cf = Dict{String, Array{Float64,1}}() 
     avoided_capex_by_ashp_present_value = Dict(t => 0.0 for t in techs.all)
     production_factor = DenseAxisArray{Float64}(undef, techs.all, p.time_steps)
-    tech_renewable_energy_fraction = Dict(t => 0.0 for t in techs.all)
+    tech_renewable_energy_fraction = Dict(t => 0.0 for t in union(techs.fuel_burning, techs.elec))
     # !!! note: tech_emissions_factors are in lb / kWh of fuel burned (gets multiplied by kWh of fuel burned, not kWh electricity consumption, ergo the use of the HHV instead of fuel slope)
     tech_emissions_factors_CO2 = Dict(t => 0.0 for t in techs.all)
```
adfarth, line 37: @zolanaj  in the `if "Generator" in techs.all` statement, do you think we should add this?: 

`tech_renewable_energy_fraction["Generator] = p.s.generator.fuel_renewable_energy_fraction`

src/core/reopt_inputs.jl
```diff
@@ -353,7 +353,7 @@ function setup_tech_inputs(s::AbstractScenario, time_steps)
     cap_cost_slope = Dict{String, Any}()
     om_cost_per_kw = Dict(t => 0.0 for t in techs.all)
     production_factor = DenseAxisArray{Float64}(undef, techs.all, 1:length(s.electric_load.loads_kw))
-    tech_renewable_energy_fraction = Dict(t => 1.0 for t in techs.all)
+    tech_renewable_energy_fraction = Dict(t => 1.0 for t in union(techs.elec, techs.fuel_burning))
```
adfarth, line 356: @zolanaj we still don't have a "tech_renewable_energy_fraction" for Boiler. I'm not suggesting we implement that in this PR, but do you think it would be safer to default this to 0 (like we do in bau_inputs) rather than 1 for cases like this in which we haven't yet implemented that input?

adfarth: Oh I'm realizing if we do that we need to explicitly define Wind and PV as tech_renewable_energy_fraction = 1.0. So I guess either: 
- Default it to zero and define PV and Wind as 1 within their setup_tech functions
- Or leave the default as 1 but temporarily define the boiler tech_renewable_energy_fraction as 0 in its setup_tech function
?

src/core/bau_inputs.jl
zolanaj: > @zolanaj in the `if "Generator" in techs.all` statement, do you think we should add this?:
> 
> `tech_renewable_energy_fraction["Generator] = p.s.generator.fuel_renewable_energy_fraction`

Agreed, good catch @adfarth! I addressed this in https://github.com/NREL/REopt.jl/commit/cc3abd0113f92e0f14b84a80ef456e58d9038bba

src/core/reopt_inputs.jl
zolanaj: > @zolanaj we still don't have a "tech_renewable_energy_fraction" for Boiler. I'm not suggesting we implement that in this PR, but do you think it would be safer to default this to 0 (like we do in bau_inputs) rather than 1 for cases like this in which we haven't yet implemented that input?

I think it's worth adding, so I've implemented it here: https://github.com/NREL/REopt.jl/commit/0ce4f171743fd61ca842137dc0d3821081336039

I'll follow up with an update to the changelog.

zolanaj: > Oh I'm realizing if we do that we need to explicitly define Wind and PV as tech_renewable_energy_fraction = 1.0. So I guess either:
> 
> * Default it to zero and define PV and Wind as 1 within their setup_tech functions
> * Or leave the default as 1 but temporarily define the boiler tech_renewable_energy_fraction as 0 in its setup_tech function
>   ?

I think that it's probably best to not include any starting assumptions and change things one technology at a time.  This is implemented here: https://github.com/NREL/REopt.jl/commit/03d5f383dc069b694dd51d50cbabb3bde41af069

src/core/reopt_inputs.jl
```diff
@@ -738,17 +742,19 @@ end
...
 This version of this function, used in BAUInputs(), doesn't update renewable energy and emissions arrays.
 """
 function setup_boiler_inputs(s::AbstractScenario, max_sizes, min_sizes, existing_sizes, cap_cost_slope, boiler_efficiency,
-                            om_cost_per_kw, production_factor, fuel_cost_per_kwh, heating_cf)
+        tech_renewable_energy_fraction, om_cost_per_kw, production_factor, fuel_cost_per_kwh, heating_cf)
     max_sizes["Boiler"] = s.boiler.max_kw
     min_sizes["Boiler"] = s.boiler.min_kw
     existing_sizes["Boiler"] = 0.0
     boiler_efficiency["Boiler"] = s.boiler.efficiency
+    production_factor["Boiler", :] = ones(8760*s.settings.time_steps_per_hour)
```
adfarth, line 756: @zolanaj why is this needed if `production_factor["Boiler", :] = get_production_factor(s.boiler)` is down below?

src/core/reopt_inputs.jl
adfarth: This fix is great!

adfarth (approved):
Just one remaining question about Boiler prod factor but otherwise looks great--thanks for these improvements!

src/core/reopt_inputs.jl
zolanaj: > @zolanaj why is this needed if `production_factor["Boiler", :] = get_production_factor(s.boiler)` is down below?

it isn't needed!  not sure how I put that in there but I removed it in https://github.com/NREL/REopt.jl/commit/219096f7ad6492e1c389effa68cb278726fe6c69

zolanaj:
> Just one remaining question about Boiler prod factor but otherwise looks great--thanks for these improvements!

Thanks for the review and improvements @adfarth!  I'll merge once tests are confirmed to pass.

PR #482, opened 2025-02-27
Fix Production Based Incentives

adfarth:
Previously, production-based incentives for techs other than `PV` were not being applied, due to issues mainly in reopt_inputs.jl. This PR mirrors our process for updating other tech-indexed dictionaries like `cap_cost_slope` for the dictionaries: `pbi_pwf`, `pbi_max_benefit`, `pbi_max_kw`, `pbi_benefit_per_kwh` and fixes some indexing issues. 

### Fixed
- Fix implementation of production-based incentives
### Added
- Added the fields `production_incentive_per_kwh`, `production_incentive_max_benefit`, `production_incentive_years`, and `production_incentive_max_kw` to the **SteamTurbine** struct.


src/core/bau_inputs.jl
```diff
@@ -65,12 +65,12 @@ function BAUInputs(p::REoptInputs)
         cap_cost_slope[pvname] = 0.0
         tech_renewable_energy_fraction[pvname] = 1.0
         if pvname in p.techs.pbi
-            push!(pbi_techs, pvname)
+            push!(techs.pbi, pvname)
         end
         pv = get_pv_by_name(pvname, p.s.pvs)
-        fillin_techs_by_exportbin(techs_by_exportbin, pv, pv.name)
+        fillin_techs_by_exportbin(techs_by_exportbin, pv, pvname)
         if !pv.can_curtail
```
adfarth, line 72: this doesn't make any difference as far as I can tell (pv.name == pvname here)

src/core/reopt_inputs.jl
```diff
@@ -485,44 +492,28 @@ function setup_tech_inputs(s::AbstractScenario, time_steps)
...
+All dicts can be empty if no techs have production_incentive_per_kwh > 0.
 """
-function setup_pbi_inputs(s::AbstractScenario, techs::Techs)
-
-    pbi_pwf = Dict{String, Any}()
-    pbi_max_benefit = Dict{String, Any}()
-    pbi_max_kw = Dict{String, Any}()
-    pbi_benefit_per_kwh = Dict{String, Any}()
-
-    for tech in techs.all
```
adfarth, line 505: tech in techs.all were all strings, so the code below was not working as intended.

zolanaj (approved):
Looks good, thanks @adfarth!

I added some tests, caught a nit, and have the fix proposed [here](https://github.com/NREL/REopt.jl/commit/8624dc696357283799da5f30a418f4ad3633a663).  I also edited a few docstrings while at it.  This is good to merge assuming tests pass and you agree with the updates!

The SteamTurbine defaults assume no incentives, so we won't need to update the API to reflect new inputs unless we envision this being available in the web tool.  It shouldn't be a heavy lift to add the four attributes though.

src/core/bau_inputs.jl
```diff
@@ -201,7 +201,7 @@ function BAUInputs(p::REoptInputs)
         seg_min_size,
         seg_max_size,
         seg_yint,
-        p.pbi_pwf, 
+        p.pbi_pwf, # Same pbi dict as optimal case
```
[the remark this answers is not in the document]
adfarth: Sorry for not understanding, @zolanaj are you proposing a change or saying this is okay as-is?

PR #491, opened 2025-03-19
Option to constrain initial CAPEX 

adfarth:
Adds inputs: `min_initial_capital_costs_before_incentives` and `max_initial_capital_costs_before_incentives` which constrain a calculation equivalent to the result `Financial.initial_capital_costs`. 

Note: The `Financial` results `initial_capital_costs` and `initial_capital_costs_after_incentives` do not currently account for the `third_party_factor`. I.e., they indicate the cost to the third party but not the capex from the host's perspective. See this result which shows the initial capex and lifecycle capex as different despite no replacements or incentives: https://reopt.nrel.gov/tool/results/c6164b3a-e59b-4efe-b034-0397a0429960. This is the current behavior and is persisted in this PR.


src/constraints/cost_curve_constraints.jl
```diff
@@ -65,4 +65,91 @@ function add_cost_curve_vars_and_constraints(m, p; _n="")
     @constraint(m, [t in p.techs.segmented],
         m[Symbol("dvPurchaseSize"*_n)][t] == m[Symbol("dvSize"*_n)][t] - p.existing_sizes[t]
     )
+end
+
+function add_capex_constraints(m, p; _n="")
+    @warn "Adding capital costs constraints. These may cause an infeasible problem in some cases, particularly for resilience runs."
+    initial_capex = initial_capex_opt(m, p)
```
adfarth, line 72: initialize as expression then `add_to_expression!` instead of +=

src/constraints/cost_curve_constraints.jl
```diff
@@ -65,4 +65,91 @@ function add_cost_curve_vars_and_constraints(m, p; _n="")
...
+        end
+    end
+
+    if "Wind" in p.techs.all
+        initial_capex += p.s.wind.installed_cost_per_kw * m[Symbol("dvPurchaseSize"*_n)]["Wind"]
+    end
+
+    if "CHP" in p.techs.all
+        chp_size_kw = m[Symbol("dvPurchaseSize"*_n)]["CHP"]
+        initial_capex += get_chp_initial_capex(p, chp_size_kw)
```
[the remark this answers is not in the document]
adfarth: check for value call in this function

1. start with expression for initial capex keep as model object
2. when in results, just call value of expression!

src/constraints/cost_curve_constraints.jl
```diff
@@ -65,4 +65,144 @@ function add_cost_curve_vars_and_constraints(m, p; _n="")
...
+        add_to_expression!(m[:InitialCapexNoIncentives], 
+            p.s.ashp.installed_cost_per_kw * m[Symbol("dvPurchaseSize"*_n)]["ASHPSpaceHeater"]
+        )
+    end
+
+    if "ASHPWaterHeater" in p.techs.all
+        add_to_expression!(m[:InitialCapexNoIncentives], 
+            p.s.ashp_wh.installed_cost_per_kw * m[Symbol("dvPurchaseSize"*_n)]["ASHPWaterHeater"]
+        )
+    end
```
zolanaj, line 211: I checked and we don't leave any technologies out, but we may want to note in the Wiki that if we add a technology to REopt that this function should be updated in addition to the other parts of the code that require it.

zolanaj (approved):
Looks good, thanks a bunch @adfarth!  I checked and ran a separate local test to confirm the function of the constraints, and I like the moving of the financial expression into a new function before the objective.  I have one quick question on outside test changes and a couples opinions on TODO's in the function, but this is ready to merge as-is, once the changelog is updated!

src/results/financial.jl
```diff
@@ -117,10 +120,10 @@ function add_financial_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _
     
     r["lifecycle_capital_costs"] = value(m[Symbol("TotalTechCapCosts"*_n)] + m[Symbol("TotalStorageCapCosts"*_n)] + m[Symbol("GHPCapCosts"*_n)] +
         m[Symbol("OffgridOtherCapexAfterDepr"*_n)] - m[Symbol("AvoidedCapexByGHP"*_n)] - m[Symbol("ResidualGHXCapCost"*_n)] - m[Symbol("AvoidedCapexByASHP"*_n)])
-    
+    #TODO: Should lifecycle_capital_costs include m[:mgTotalTechUpgradeCost] + m[:dvMGStorageUpgradeCost]?
```
zolanaj, line 123: I'd say yes to this - the MG upgrade is also an upfront capital cost.

zolanaj:
@adfarth in addition to the changelog update, I added an error check for min > max capital cost input.  Let me know if you have any concerns - thanks!

src/constraints/cost_curve_constraints.jl
```diff
@@ -65,4 +65,144 @@ function add_cost_curve_vars_and_constraints(m, p; _n="")
...
+                    seg_yint_no_inc[s] * m[Symbol("binSegment"*t)][s] for s in 1:p.n_segs_by_tech[t])
+            )
+        else
+            add_to_expression!(m[:CHPCapexNoIncentives], cost_list * m[Symbol("dvPurchaseSize"*_n)]["CHP"])
+        end
+        # TODO: include supplementary firing costs?
+        #Add supplementary firing capital cost
+        # chp_supp_firing_size = self.nested_outputs["Scenario"]["Site"][tech].get("size_supplementary_firing_kw")
+        # chp_supp_firing_cost = self.inputs[tech].get("supplementary_firing_capital_cost_per_kw") or 0
+        # add_to_expression!(m[:CHPCapexNoIncentives], chp_supp_firing_size * chp_supp_firing_cost)
```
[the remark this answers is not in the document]
adfarth: added in https://github.com/NREL/REopt.jl/pull/491/commits/d841386488424d3ef8b039afc0c62dc7566f6b2c!

src/constraints/cost_curve_constraints.jl
adfarth: good idea, I added this to https://github.com/NREL/REopt.jl/wiki/Steps-to-register-a-new-version-of-the-REopt-package

zolanaj: looks great, thanks for doing that!

src/results/financial.jl
adfarth: Thanks for weighing in! I've added this here and in the initial capex no incentives function https://github.com/NREL/REopt.jl/pull/491/commits/3431afd9816909149eabee6034f16d486793793d

test/runtests.jl
```diff
@@ -2820,7 +2837,7 @@ else  # run HiGHS tests
             
                 s = Scenario(d)
                 p = REoptInputs(s)
-                m = Model(optimizer_with_attributes(HiGHS.Optimizer, "output_flag" => false, "log_to_console" => false))
+                m = Model(optimizer_with_attributes(HiGHS.Optimizer, "output_flag" => false, "log_to_console" => false, "mip_rel_gap" => 0.05))
```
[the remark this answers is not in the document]
Bill-Becker: @adfarth chiming in here late because I didn't realize how much CHP runs are slowing down because of this CapEx constraint, even when not using the CapEx constraint. We use CHP cost curves by default, so this is concerning if there's a major slowdown with them. Let's chat about what you've seen in your testing and how we can mitigate before putting this into the next registered version.

PR #498, opened 2025-05-16
Reduce memory growth on runtests.jl

Bill-Becker:
Speed up runtests.jl job in GitHub Actions by clearing the memory buildup from the optimization models.

Also attempted to use the ubuntu (linux) runner and the macos-latest-large (more memory (32 GB) Mac OS), and confirmed that those are still not working.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         julia-version: ['1.10']
         julia-arch: [x64]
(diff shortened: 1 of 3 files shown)


rathod-b (approved):

rathod-b:
Approved. Quick note: It is a bit unclear on how `finalize` call differs from `empty!` call. Looking at HiGHS code base, it appears both those calls "destroy" the model object. It could be worth exploring if we can get by with just one of these calls.

PR #503, opened 2025-06-24
Fix issue with No-Incentive CapEx calc

Bill-Becker:
Fixes an issue that was found soon after CapEx constraint was deployed to the web tool, where the user inputs a rebate incentive with a maximum value for CHP (which has a default 2-pt cost curve), which causes n_segs to be greater than the non-incentive cost curve.

This would also be an issue with PV with a 2-pt cost curve where there is also an incentive with a maximum value.


adfarth (approved):
Thank you for catching and fixing this!

PR #505, opened 2025-06-30
Avoid ignoring input max_kw with inputs for space availability

Bill-Becker:
This PR fixes an issue that ignores the input max_kw when there is also a roof/land space input(s). The new behavior checks if the size_kw based on roof+land space is less than the input max_kw, and then assigns the space_constrained status. Before, if there was any roof/land input, even if the max_kw was less than the size based on the roof+land, it would effectively ignore the max_kw input.

### DIFF
--- a/src/core/pv.jl
+++ b/src/core/pv.jl
@@ -420,19 +420,22 @@ function get_pv_cost_params(;
         # Default case: Calculate based on land/roof space available or load
         max_kw_for_size_class = max_kw
         max_kw_roof, max_kw_ground = 0.0, 0.0
+        check_for_space_constrained = false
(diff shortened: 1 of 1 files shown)


bpulluta (approved):
This PR correctly resolves the issue where `max_kw` was ignored when space constraints were present.

User's `max_kw` is now respected as intended - the effective limit is `min(user_max_kw, space_max_kw)`.

Testing confirmed:

User `max_kw = 100`, Space allows 500 kW: Uses 100 kW (user limit), space_constrained = false
User `max_kw = 500`, Space allows 100 kW: Uses 100 kW (space limit), space_constrained = true

PR #524, opened 2025-10-10
 Load Component Alignment & JuMP Performance Improvements

bpulluta:
## Overview
This PR enables combining multiple load profiles from different source years (e.g., site load from 2016, EV load from 2024) with proper weekday alignment and leap year handling. Also includes JuMP expression building optimization for sub-hourly time resolutions.

### Multiple Load Components Support
- **`load_components`** in `ElectricLoad` to combine profiles from different source years
- **`src/core/load_alignment.jl`** module with centralized alignment functions:
  - `align_series_to_year()` - weekday alignment across years

### Sub-hourly Support
- Full support for 30-min, 15-min, 5-min time resolutions in load component alignment

### Code Consolidation
- Refactored day-of-week shifting logic in `doe_commercial_reference_building_loads.jl` and `electric_utility.jl` to use centralized `align_series_to_year()` function
- Eliminated code duplication across load processing functions by consolidating leap year normalization logic

### JuMP Performance Optimization
- **Problem:** Sub-hourly time steps (e.g., 15-min = 35,040 steps/year) caused JuMP warnings with large `sum()` comprehensions
- **Solution:** Replaced with efficient expression building using `JuMP.AffExpr()` and `JuMP.add_to_expression!()`
- **Impact:** Cleaner code, no warnings, better performance for high-resolution models

## Example Usage
```json
{
  "ElectricLoad": {
    "load_components": [
      {"loads_kw": [...], "year": 2016, "label": "Building"},
      {"loads_kw": [...], "year": 2024, "label": "EV"}
    ],
    "year": 2024
  }
}
```

## Testing
✅ All 1100+ existing tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,15 +34,7 @@ Classify the change according to the following categories:
 ### Changed
 - Refactored day-of-week shifting logic in `doe_commercial_reference_building_loads.jl` and `electric_utility.jl` to use centralized `align_series_to_year()` function
 - Consolidated leap year normalization logic to eliminate code duplication across load processing functions
-- When using subhourly time steps (e.g., 15-minute intervals = 35,040 time steps per year), JuMP would display warnings:
-```
-┌ Warning: The addition operator has been used on JuMP expressions a large number of times. 
-This warning is safe to ignore but may indicate that model generation is slower than necessary. 
-For performance reasons, you should not add expressions in a loop. Instead of x += y, use 
-add_to_expression!(x,y) to modify x in place.
-└ @ JuMP
-```
-**Solution**: Replace large `sum()` comprehensions over time steps with efficient expression building using `JuMP.AffExpr()` and `JuMP.add_to_expression!()`.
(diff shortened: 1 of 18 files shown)


Bill-Becker:
@bpulluta it would be good if we could split the "add to expression" stuff into a separate PR, so just start from the develop branch and cherry-pick these four commits (I think):
image

And then as we talked about, probably removing the separate leap day policy option, and I'm curious how much simpler it would be to assume the user must put in the multiple load components **with the same calendar year structure** (e.g. 2017 and 2023 are the same). I'm open to seeing if the load year shift stuff could still be added while minimizing the extra complexity in core/electric_load.jl, but as I mentioned I don't think it's too inconvenient for the user to handle and necessary shifting. I'm mainly worried about maintainability/readability and future development flexibility (sorry for all the "ility's").

bpulluta:
> be good if we could split the "add to expression" stuff into a separate PR, so just start from the develop branch and cherry-pick these four commits (I think):

That sounds like a great idea, I'll do that and separate the load stuff from the JuMP expression bug fixes

PR #532, opened 2025-11-04
Fix 2 type bugs

hdunham:
### Fixed
- Bug where storage input with only _Int_ values restricted _Dict_ subtype passed into `set_sector_defaults!` and caused _InexactError_
- Bug where **ElectricStorage** input with only _Real_ values restricted _Dict_ subtype, causing **off_grid_flag** to be added as an _Int_ instead of _Bool_, leading to a conversion error in **ElectricStorageDefaults**


adfarth (changes requested):
Left a few comments!

adfarth (approved):
Approved pending tests passing!

PR #547, opened 2025-12-16
update Big-M calculation on demand and energy tiers with large storage max sizes

zolanaj:
### Fixed
- Addresses a bug in which demand and energy charges are inaccurately calculated when large max storage system sizes are present.


Bill-Becker:
@zolanaj is this still a relevant bug fix that we should merge?

rathod-b:
> @zolanaj is this still a relevant bug fix that we should merge?

@Bill-Becker This is still a relevant bug (Tiered demand charges get applied out of order under the financial scenario with onsite technologies. I just tested that this PR fixes the issue.). We can merge it into master when you/alex think is best.

rathod-b (approved):
Tested that these changes fix the "tiered demand charges not applying to lower tiers under financial case correctly" bug.

rathod-b:
@zolanaj the updated runtest does take a long time even with commercial solver. This is just an observation and something we should track.

Bill-Becker:
> @zolanaj the updated runtest does take a long time even with commercial solver. This is just an observation and something we should track.

Is this going to slow down all tiered rate analysis with battery similar to this level? That will cause issues in the API with memory consumption as it continues to search, and/or lead to a lot of timeouts. We may need to throw a warning about this to users when they evaluate tiered rates with battery. 

For the test time, I think we should either try to "simplify" this test (taking 26 minutes, about 40% of the previous total runtests.jl test time) if we can still test the desired behavior. If not, I'd be fine to "skip" the test but leave it in there for periodic testing when we want to "unskip".

PR #577, opened 2026-02-11
CHP load-following and absorption-chiller-only dispatch strategies

zolanaj:
- [x] CHANGELOG.md is updated
- [] Any new REopt outputs and required inputs have been added to the corresponding Django model in the REopt API

### Added
- New optional attributes for **CHP** object **CHP.serve_absorption_chiller_only**, **CHP.months_serving_absorption_chiller_only**, and **CHP.follow_electrical_load**, which impose constraints on CHP operations if selected.  The default is set to `false` for both attributes.
- New result **thermal_to_absorption_chiller_series_mmbtu_per_hour** added to heating technologies and new result **storage_to_absorption_chiller_series_mmbtu_per_hour** for hot thermal storage technologies.  This result is included as a part of the thermal site loads served, i.e., the adding this result does not change the existing results.
- Added results fields to **HighTempThermalStorage** to match those of **HotThermalStorage**.

### Changed
- Updated heating dispatch results by separating heat flows to absorption chiller from heating load served (formerly, these were aggregated).

### Fixed
- Fixed a bug in which the CHP system requires a **DomesticHotWater** load.
- Fixed a bug in which the storage to steam turbine flow was included in the thermal heating load served.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ Classify the change according to the following categories:
 ## chp-load-following
 ### Added
 - New optional attributes for **CHP** object **CHP.serve_absorption_chiller_only**, **CHP.months_serving_absorption_chiller_only**, and **CHP.follow_electrical_load**, which impose constraints on CHP operations if selected.  The default is set to `false` for both attributes.
-- New result **thermal_to_absorption_chiller_series_mmbtu_per_hour** added to heating technologies.  This result is included as a part of the thermal site loads served, i.e., the adding this result does not change the existing results.
+- New result **thermal_to_absorption_chiller_series_mmbtu_per_hour** added to heating technologies and new result **storage_to_absorption_chiller_series_mmbtu_per_hour** for hot thermal storage technologies.  This result is included as a part of the thermal site loads served, i.e., the adding this result does not change the existing results.
 - Added results fields to **HighTempThermalStorage** to match those of **HotThermalStorage**.
 
 ### Changed


zolanaj:
@Bill-Becker this is ready for your review!  I'm still crafting a test case for the false CHP DomesticHotWater load requirement in https://github.com/NatLabRockies/REopt.jl/issues/576 but playtesting showed it worked... not sure if the specific instance requiring your workaround is available but I can try that if you have it.  The failing tests in the most recent push are currently the PVWatts tests on actions and not related to this PR.

Only other note is that the storage doesn't have a variable sending to the absorption chiller at the moment - I think that may make this incomplete as a result.  I'm working on that right now.  (API dev will be done on this shortly, too.)

src/core/reopt_inputs.jl
```diff
@@ -231,6 +231,15 @@ function REoptInputs(s::AbstractScenario)
             absorption_chillers_using_heating_load["ProcessHeat"] = Vector{String}()
         end
     end
+    
+    if !isnothing(s.absorption_chiller) && (
```
Bill-Becker: Does this fix the issue where absorption chiller can't work unless DomesticHotWaterLoad is input?

src/core/chp.jl
```diff
@@ -113,6 +116,9 @@ Base.@kwdef mutable struct CHP <: AbstractCHP
     can_serve_space_heating::Bool = true
     can_serve_process_heat::Bool = true
     is_electric_only::Bool = false
+    serve_absorption_chiller_only::Bool = false
```
Bill-Becker: An issue was just identified that is very related to this work. CHP size heuristic/estimate is based on the average heating load to inform the default size class -> cost/performance params and the max size considered (2x the size based on the avg heating load). With an absorption chiller focus, possibly not even including a heating load, the CHP size heuristic will be very small to zero, so the max size will be zero. I think we want to include the cooling load conversion to a heuristic heating load that the absorption chiller will consume when absorption chiller is analyzed and especially with serve_absorption_chiller_only is true.

src/core/chp.jl
```diff
@@ -40,6 +40,9 @@ conflict_res_min_allowable_fraction_of_max = 0.25
     can_serve_space_heating::Bool = true # If CHP can supply heat to the space heating load
     can_serve_process_heat::Bool = true # If CHP can supply heat to the process heating load
     is_electric_only::Bool = false # If CHP is a prime generator that does not supply heat
+    serve_absorption_chiller_only::Bool = false # If CHP only can serve absorption chiller with produced heat
```
adfarth: truly not trying to be nit-picky, but is "If CHP can only serve" more accurate than "If CHP only.." (heat could also come from other technologies to serve the absorption chiller, right?)

src/core/chp.jl
```diff
@@ -40,6 +40,9 @@ conflict_res_min_allowable_fraction_of_max = 0.25
...
+    months_serving_absorption_chiller_only::AbstractVector{Int64} = Int64[] # months in which only absorption chiller is served.
```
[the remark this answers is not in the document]
adfarth: Could also specify here if this only gets applied if serve_absorption_chiller_only = True (and could specify in the previous description if you must also specify months if serve_absorption_chiller_only = True)

src/core/chp.jl
zolanaj: Good clarification, I've tried to address this [here](https://github.com/NatLabRockies/REopt.jl/commit/cd3084fe067deb7a50921f5d930cd9efa1235c13).

Bill-Becker:
@zolanaj I re-reviewed by testing, and I made some updates based on that testing, described below (thanks co-pilot for summarizing!). I added more tests. Please review the updates. I'll check on CI tests and help debug any failing ones, and I'll approve after you review my updates.

## min_turn_down_fraction Conflicts with Load-Following

**Issue:** When `follow_electrical_load = true`, the constraint `CHP_output = min(CHP_size, load)` conflicts with `min_turn_down_fraction` when `load < min_turn_down_fraction * CHP_size`, causing infeasibility.

**Fix Implemented:** Added preprocessing validation in CHP constructor that throws error when `min_load < min_turn_down_fraction * min_kw` (guaranteed infeasibility when min_kw is specified). Provides specific guidance: set `min_turn_down_fraction = 0.0` or reduce min_kw below the calculated threshold.

**Location:** `src/core/chp.jl` lines 290-306

**Future Enhancements:**
1. Auto-override `min_turn_down_fraction` to 0.0 when `follow_electrical_load = true` (with warning)
2. Modify load-following constraints to respect min_turn_down (would require additional binary variables)

---

## Absorption Chiller Monthly Constraint Bug

**Issue:** In `add_chp_to_absorption_chiller_only_constraints()`, heating load variables were fixed to zero for ALL timesteps instead of only restricted months, preventing CHP from serving heating loads year-round.

**Fix Implemented:** Modified constraint to only fix heating variables during `months_serving_absorption_chiller_only` timesteps.

**Location:** `src/constraints/chp_constraints.jl` lines 183-189

---

## CHP Results Double-Counting Absorption Chiller Heat

**Issue:** `thermal_to_dhw_load_series` (and space heating/process heat) didn't subtract heat going to absorption chiller, causing double-counting in results when `heating_load_input = "DomesticHotWater"`.

**Evidence:** Test violations showed `to_dhw > 0` during restricted months when it should be zero.

**Fix Implemented:** Added `- CHPtoAbsorptionChillerByQualityKW[q,ts]` to all load-specific thermal output calculations (DHW, SpaceHeating, ProcessHeat).

**Location:** `src/results/chp.jl` lines 117-143

---

## Cooling Load Integration for CHP Heuristic Sizing

**Enhancement:** Extended CHP sizing heuristic to account for cooling loads served by absorption chillers, enabling more accurate CHP sizing when thermal output serves both heating and cooling demands.

**Implementation:** 
- Added cooling parameters (`avg_cooling_load_kw`, `chiller_cop`, `include_cooling_in_size`) to `get_chp_defaults_prime_mover_size_class()` function
- Updated `CHP()` constructor to accept and pass cooling parameters from processed `CoolingLoad` object
- Modified `Scenario()` constructor to position CHP construction after `CoolingLoad` processing but before `AbsorptionChiller` construction, allowing cooling data to flow into CHP sizing
- Automatic detection: `include_cooling_in_size` defaults to `true` when `AbsorptionChiller` is present in scenario inputs
- Cooling thermal load is converted to equivalent heating load using absorption chiller COP and CHP thermal efficiency for sizing calculations

**Location:** 
- `src/core/chp.jl` lines 153-158 (constructor), 228-236 (call to sizing function), 420-426 (sizing function parameters), 515-527 (cooling load calculations)
- `src/core/scenario.jl` lines ~441-564 (CoolingLoad → CHP → AbsorptionChiller ordering)

**Testing:** Added Case 4 test in "CHP Sizing Heuristic" testset verifying that full Scenario construction with CHP + AbsorptionChiller produces same sizing as direct function call with equivalent parameters.

PR #583, opened 2026-03-11
Fix ExistingChiller sizing issue

atpham88:
### Overview
This PR addresses two issues related to ExistingChiller sizing (https://github.com/NatLabRockies/REopt.jl/issues/569 and https://github.com/NatLabRockies/REopt.jl/issues/568):
- ExistingChiller only applies an extra 1.25 factor on peak to chiller's size in the optimal case when no capital cost is provided. It still only applies the 1.25 factor one time (correctly) in BAU or when capital cost is provided
- Fix: Switch to use `max(dvCoolingProduction)` for chiller's size instead of `dvSize`. I'm unclear why `dvSize` gives a different value than `max(dvCoolingProduction)` in some cases at this point but will look into this more later (issued documented: https://github.com/NatLabRockies/REopt.jl/issues/587).

### Validation
- Test case for a commercial building in Atlanta with both heating and cooling loads. Confirmed that chiller size only applies 1.25 factor ONE TIME across settings (without cap cost, with cap cost, in BAU, in Optimal without GHP deployment).
- Also tested with Amanda's demo case study for a commercial building in Denver where capital cost of ExistingChiller is not provided. Confirmed chiller's size in BAU is the same as in optimal case without GHP deployment.
- Details of results for both of these tests are here https://github.com/atpham88/existing_chiller_debug_REopt/tree/main/results.

- [x] CHANGELOG.md is updated
- [ ] Docs have been added / updated if applicable
- [ ] Tests for the changes have been added. These tests should compare against expected values that are calculated independently of running REopt. Tests should also include garbage collection (see other tests for examples).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ Classify the change according to the following categories:
 ### Fixed
 - Fixed a bug in which the CHP system requires a **DomesticHotWater** load.
 - Fixed a bug in which the storage to steam turbine flow was included in the thermal heating load served.
+- Switched to use maximum value of **dvCoolingProduction** for **ExistingChiller**'s size instead of **dvSize**. This fixed the issue with **ExistingChiller**'s size being applied **max_thermal_factor_on_peak_load** twice in some cases.
 
 ### Changed
 - **HotThermalStorage** and **HighTempThermalStorage** output **storage_to_turbine_series_mmbtu_per_hour** to **storage_to_steamturbine_series_mmbtu_per_hour**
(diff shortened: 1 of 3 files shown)


Bill-Becker (approved):

Bill-Becker (approved):

adfarth (approved):
@atpham88 before merging, could you update the CHANGELOG and also create a new issue for this: "I'm unclear why dvSize gives a different value than max(dvCoolingProduction) in some cases at this point but will look into this more later."?

adfarth (approved):
@atpham88 sorry--I thought I had approved this before! Approving pending getting tests to pass (I see some are failing now)

PR #604, opened 2026-05-20
Add dicttype arg to JSON.parsefile calls throughout code

hdunham:
- [x] CHANGELOG.md is updated
- [x] Any new REopt outputs and required inputs have been added to the corresponding Django model in the REopt API


zolanaj (approved):
Thanks @hdunham!  This looks good as is, and when testing this branch on the API all tests are passing.  

This is good to merge but one question: do we want to update the test suite so that the JSON.parsefile(...) calls within are updated to call this new function?

PR #605, opened 2026-05-26
Water power original batch of updates

toddleif:

Original batch of code updates for modeling water power in REopt.


CHANGELOG.md
```diff
@@ -415,6 +415,12 @@ Update the following inputs from the previous --> new values:
 - Fixed a reshape call in function `parse_urdb_tou_demand` that incorrectly assumed row major instead of column major ordering
 - Fixed a loop range in function `parse_urdb_tou_demand` that incorrectly started at 0 instead of 1
 - Added the missing tier index when accessing `p.s.electric_tariff.tou_demand_rates` in function `add_elec_utility_expressions`
+## Develop 2024-04-19
+### Added
```
hdunham, line 424: Why are these being added the changelog? I don't see how elec stor o&m changes are part of this pr and actually don't see these changes in the pr either

CHANGELOG.md
```diff
@@ -415,6 +415,12 @@ Update the following inputs from the previous --> new values:
...
+- Added new inputs **om_cost_per_kw** and **om_cost_per_kwh** to `ElectricStorage` for modeling capacity-based O&M
```
hdunham, line 425: water power changes should be added to the changelog eventually but no huge rush on that

src/results/electric_storage.jl
```diff
@@ -37,6 +37,13 @@ function add_electric_storage_results(m::JuMP.AbstractModel, p::REoptInputs, d::
             r["size_kw"] * p.s.storage.attr[b].installed_cost_per_kw +
             p.s.storage.attr[b].installed_cost_constant
 
+        # TODO: correct the StoragePerUnitOMCosts
```
hdunham, line 40: what's the deal with this? how is it related to the water power work and why is this being left as a todo with StoragePerUnitOMCosts = -9999999 ?

src/results/electric_storage.jl
```diff
@@ -59,11 +66,12 @@ function add_electric_storage_results(m::JuMP.AbstractModel, p::REoptInputs, d::
                     )
                 end
             end
-            r["residual_value"] = value(m[:residual_value])
```
hdunham, line 62: why remove?

test/Manifest.toml
```diff
@@ -0,0 +1,391 @@
+# This file is machine-generated - editing it directly is not advised
+
+julia_version = "1.8.0"
```
hdunham, line 3: we don't need to include this file in the repo

test/scenarios/storage_om.json
```diff
@@ -0,0 +1,43 @@
+{
```
hdunham, line 1: seems to be a test for the storage o&m changes mentioned in the changelog but I'm not seeing in the PR? did we mean to include?

hdunham, line 1: seems to be a test for the storage o&m changes mentioned in the changelog but I'm not seeing in the PR? did we mean to include?

src/results/results.jl
```diff
@@ -44,6 +44,18 @@ function reopt_results(m::JuMP.AbstractModel, p::REoptInputs; _n="")
         add_wind_results(m, p, d; _n)
     end
     
+    if "WaterPower_Turbine1" in p.techs.elec
+        add_water_power_results(m, p, d; _n)
```
hdunham, line 48: will there always be a turbine if there's a pump, so only need to check of there's a turbine, and then can add all pump/turbine results?

src/results/water_storage.jl
```diff
@@ -0,0 +1,88 @@
...
+- `upper_reservior_initial_capital_costs`
+
+!!! note "'Series' and 'Annual' energy outputs are average annual"
+	REopt performs load balances using average annual production values for technologies that include degradation. 
+	Therefore, all timeseries (`_series`) and `annual_` results should be interpretted as energy outputs averaged over the analysis period. 
+
+"""
+
+function add_upstream_reservoir_water_storage_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _n="")
+	# Adds the `WaterPower` results to the dictionary passed back from `run_reopt` using the solved model `m` and the `REoptInputs` for node `_n`.
```
hdunham, line 18: should say adds the upstream reservoir results

hdunham, line 18: should say adds the upstream reservoir results

src/results/water_storage.jl
```diff
@@ -0,0 +1,88 @@
...
+- `combined_upstream_and_downstream_resevoir_costs`
+
+!!! note "'Series' and 'Annual' energy outputs are average annual"
+	REopt performs load balances using average annual production values for technologies that include degradation. 
+	Therefore, all timeseries (`_series`) and `annual_` results should be interpretted as energy outputs averaged over the analysis period. 
+
+"""
+
+function add_downstream_reservoir_water_storage_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _n="")
+	# Adds the `WaterPower` results to the dictionary passed back from `run_reopt` using the solved model `m` and the `REoptInputs` for node `_n`.
```
hdunham, line 61: should say adds the downstream reservoir results

src/results/water_storage.jl
```diff
@@ -0,0 +1,88 @@
+# REopt®, Copyright (c) Alliance for Sustainable Energy, LLC. See also https://github.com/NREL/REopt.jl/blob/master/LICENSE.
+"""
+`UpstreamReservoir` results keys:
+- `combined_upstream_and_downstream_resevoir_costs` the turbine input into the model capacity
```
hdunham, line 4: the presence of this combined result is one motivation to combine all the water power results into one results section

src/results/water_storage.jl
```diff
@@ -0,0 +1,88 @@
...
+- `upstream_reservoir_water_capacity_cubic_meters`
+- `upstream_reservoir_water_volume_cubic_meters`
+- `input_to_model_tributary_water_flow`
```
hdunham, line 7: would be good to have more aligned naming convention between input_to_model_tributary_water_flow and downstream_reservoir_water_outflow_cubic_meters_per_second

src/results/water_storage.jl
```diff
@@ -0,0 +1,88 @@
+# REopt®, Copyright (c) Alliance for Sustainable Energy, LLC. See also https://github.com/NREL/REopt.jl/blob/master/LICENSE.
+"""
+`UpstreamReservoir` results keys:
+- `combined_upstream_and_downstream_resevoir_costs` the turbine input into the model capacity
+- `upstream_reservoir_water_capacity_cubic_meters`
+- `upstream_reservoir_water_volume_cubic_meters`
```
hdunham, line 6: I think we generally try to name time series outputs with _series

src/results/water_storage.jl
```diff
@@ -0,0 +1,88 @@
...
+function add_upstream_reservoir_water_storage_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _n="")
+	# Adds the `WaterPower` results to the dictionary passed back from `run_reopt` using the solved model `m` and the `REoptInputs` for node `_n`.
+	# Note: the node number is an empty string if evaluating a single `Site`.
+
+    r = Dict{String, Any}()
+
+	r["combined_upstream_and_downstream_resevoir_costs"] = m[Symbol("WaterStorageCapCosts"*_n)]
+
+	# Upstream reservoir volume
+	r["upstream_reservoir_water_capacity_cubic_meters"] = round.(value.(m[Symbol("dvUpstreamReservoirCapacity"*_n)]), digits=3)
```
hdunham, line 26: why use the dot/array version of the functions? isn't dvUpstreamReservoirCapacity a scalar?

src/results/water_storage.jl
hdunham: I don't think we need to include "input_to_model" in name since we don't do that anywhere else

src/results/water_storage.jl
hdunham: also reservoir is spelled wrong here

src/results/water_storage.jl
```diff
@@ -0,0 +1,88 @@
...
+function add_downstream_reservoir_water_storage_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _n="")
+	# Adds the `WaterPower` results to the dictionary passed back from `run_reopt` using the solved model `m` and the `REoptInputs` for node `_n`.
+	# Note: the node number is an empty string if evaluating a single `Site`.
+
+    r = Dict{String, Any}()
+
+	# Downstream reservoir volume
+	downstream_reservoir_volume = @expression(m, [ts in p.time_steps], m[Symbol("dvDownstreamReservoirWaterVolume"*_n)][ts])
+	r["downstream_reservoir_water_volume_cubic_meters"] = round.(value.(downstream_reservoir_volume).data, digits=3) 
+	r["downstream_reservoir_water_capacity_cubic_meters"] = round.(value.(m[Symbol("dvDownstreamReservoirCapacity"*_n)]), digits=3)
```
hdunham, line 69: scalar thing again

src/results/water_storage.jl
```diff
@@ -0,0 +1,88 @@
...
+
+"""
+
+function add_upstream_reservoir_water_storage_results(m::JuMP.AbstractModel, p::REoptInputs, d::Dict; _n="")
+	# Adds the `WaterPower` results to the dictionary passed back from `run_reopt` using the solved model `m` and the `REoptInputs` for node `_n`.
+	# Note: the node number is an empty string if evaluating a single `Site`.
+
+    r = Dict{String, Any}()
+
+	r["combined_upstream_and_downstream_resevoir_costs"] = m[Symbol("WaterStorageCapCosts"*_n)]
```
hdunham, line 23: should evaluate with round and value not just return the JuMP expression. same with the dupe line in downstream function

src/core/water_storage.jl
```diff
@@ -0,0 +1,252 @@
...
+```julia
+    tributary_water_inflow_cubic_meter_per_second::Array=[]
+    minimum_volume_fraction_upstream_reservoir::Float64 = 0.1
+    maximum_volume_fraction_upstream_reservoir::Float64 = 1.0
+    initial_reservoir_volume_fraction_upstream_reservoir::Float64 = 0.0
+    minimum_capacity_cubic_meters_upstream_reservoir::Float64 = 0.0
+    maximum_capacity_cubic_meters_upstream_reservoir::Float64 = 10000000.0
+    cost_per_cubic_meter_upstream_reservoir::Float64 = 25.0
+    om_cost_per_cubic_meter::Float64 = 0.0 # Yearly fixed O&M cost dependent on storage energy size
+    macrs_option_years::Int = 5 #Note: default may change if Site.sector is not "commercial/industrial"
```
hdunham, line 18: are these incentive defaults appropriate for water power?

src/core/water_storage.jl
```diff
@@ -0,0 +1,252 @@
...
+    macrs_bonus_fraction::Float64 = 1.0 #Note: default may change if Site.sector is not "commercial/industrial"
+    macrs_itc_reduction::Float64 = 0.5
+    total_itc_fraction::Float64 = 0.3 #Note: default may change if Site.sector is not "commercial/industrial"
+    total_rebate_per_cubic_meter::Float64 = 0.0
+```
+"""
+Base.@kwdef struct UpstreamReservoirStorageDefaults <: AbstractWaterStorageDefaults
+    tributary_water_inflow_cubic_meter_per_second::Array=[]
+    minimum_volume_fraction_upstream_reservoir::Float64 = 0.1
```
hdunham, line 27: for conciseness, I would remove "_upstream_reservoir" from all these since it's already in the UpstreamReservoir struct (and same for downstream)

src/core/water_storage.jl
```diff
@@ -0,0 +1,252 @@
...
+    initial_reservoir_volume_fraction_downstream_reservoir::Float64 = 0.5
+    minimum_volume_fraction_downstream_reservoir::Float64 = 0.2
+    maximum_volume_fraction_downstream_reservoir::Float64 = 1.0
+    minimum_capacity_cubic_meters_downstream_reservoir::Float64 = 0.0
+    maximum_capacity_cubic_meters_downstream_reservoir::Float64 = 10000000.0
+    cost_per_cubic_meter_downstream_reservoir::Float64 = 25.0
+    minimum_outflow_from_downstream_reservoir_cubic_meter_per_second::Float64 = 0.0
+    maximum_outflow_from_downstream_reservoir_cubic_meter_per_second::Float64 = 10000.0
+    om_cost_per_cubic_meter::Float64 = 0.0
+    tributary_water_inflow_cubic_meter_per_second::Array=[]
```
hdunham, line 74: is the downstream reservoir supposed to have the tributary inflow input?

src/core/water_storage.jl
```diff
@@ -0,0 +1,252 @@
...
+"""
+Base.@kwdef struct DownstreamReservoirStorageDefaults <: AbstractWaterStorageDefaults   
+    initial_reservoir_volume_fraction_downstream_reservoir::Float64 = 0.5
+    minimum_volume_fraction_downstream_reservoir::Float64 = 0.2
+    maximum_volume_fraction_downstream_reservoir::Float64 = 1.0
+    minimum_capacity_cubic_meters_downstream_reservoir::Float64 = 0.0
+    maximum_capacity_cubic_meters_downstream_reservoir::Float64 = 10000000.0
+    cost_per_cubic_meter_downstream_reservoir::Float64 = 25.0
+    minimum_outflow_from_downstream_reservoir_cubic_meter_per_second::Float64 = 0.0
+    maximum_outflow_from_downstream_reservoir_cubic_meter_per_second::Float64 = 10000.0
```
hdunham, line 72: is the downstream outflow like the flow leaving the system continuing down the river? why is there a max limit on this?

src/constraints/water_power_constraints.jl
```diff
@@ -0,0 +1,272 @@
...
+	@constraint(m, [ts in p.time_steps],
+		m[Symbol("dvWaterVolume"*_n)][ts] <= p.s.upstream_reservoir.maximum_volume_fraction_upstream_reservoir * m[Symbol("dvUpstreamReservoirCapacity"*_n)]
+				)
+	@constraint(m, [ts in p.time_steps],
+		m[Symbol("dvWaterVolume"*_n)][ts] >= p.s.upstream_reservoir.minimum_volume_fraction_upstream_reservoir * m[Symbol("dvUpstreamReservoirCapacity"*_n)] 
+				)
+	
+	# Water flow rate from all turbines combined is above the required minimum water flow
+	@constraint(m, [ts in p.time_steps],
+		 sum(m[Symbol("dvWaterOutFlow"*_n)][t, ts] for t in p.techs.water_power_turbines) + m[Symbol("dvSpillwayWaterFlow"*_n)][ts] >= p.s.water_power.minimum_water_output_cubic_meter_per_second_total_of_all_turbines
```
hdunham, line 46: if spillway is included in this constraint, then the min input should probably be more accurately named to so it's clear it's the min of the total flow from upper to lower reservoir

src/core/water_storage.jl
```diff
@@ -0,0 +1,252 @@
...
+            discount_rate = f.owner_discount_rate_fraction,
+            tax_rate = f.owner_tax_rate_fraction,
+            itc = stor.total_itc_fraction,
+            macrs_schedule = macrs_schedule,
+            macrs_bonus_fraction = stor.macrs_bonus_fraction,
+            macrs_itc_reduction = stor.macrs_itc_reduction
+        ) - stor.total_rebate_per_cubic_meter
+        
+        # TODO: implement the conversion to the correct timesteps for the tributary inflow
+        #stor.tributary_water_inflow_cubic_meter_per_second = convert_tributary_flow_to_correct_time_steps_per_hour(stor.tributary_water_inflow_cubic_meter_per_second, time_steps_per_hour)
```
hdunham, line 129: why would it need to be converted? if it's cubic meters per second then it's a flow rate not a quantity of water, so the time step length would be applied when the flow being used to calculate reservoir volume change from one time step to the next

hdunham: never mind, I see that it's just repeating values not changing them

src/core/reopt.jl
```diff
@@ -217,6 +217,13 @@ function build_reopt!(m::JuMP.AbstractModel, p::REoptInputs)
 				@constraint(m, [t in union(p.techs.heating, p.techs.chp), q in p.heating_loads, ts in p.time_steps], m[:dvHeatToStorage][b,t,q,ts] == 0)
 			end
 		else
+			# Additional variables for exporting storage energy to the grid
```
hdunham, line 220: why is this in here

src/core/reopt_inputs.jl
```diff
@@ -725,6 +736,41 @@ function setup_wind_inputs(s::AbstractScenario, max_sizes, min_sizes, existing_s
     return nothing
 end
 
+function setup_water_power_inputs(s::AbstractScenario, water_power_inputs, techs_by_exportbin, production_factor, techs, cap_cost_slope, tech_renewable_energy_fraction, max_sizes, min_sizes, existing_sizes, om_cost_per_kw)
```
hdunham, line 740: it doesn't seem like the arg water_power_inputs is used/updated, so then it gets passed to REoptInputs as an empty dict, and I don't think it's used after that either

src/core/scenario.jl
```diff
@@ -57,6 +61,11 @@ A Scenario struct can contain the following keys:
 - [GHP](@ref) (optional, can be Array)
 - [SteamTurbine](@ref) (optional)
 - [ElectricHeater](@ref) (optional)
+- [WaterPower](@ref) (optional)
+- [water_storage](@ref) (optional)
+- [upstream_reservoir](@ref) (optional)
+- [downstream_reservoir](@ref) (optional)
+- absorption_chillers_using_heating_load
```
hdunham, line 68: why is this being added here and with different syntax?

src/core/scenario.jl
```diff
@@ -25,6 +25,10 @@ struct Scenario <: AbstractScenario
     cooling_thermal_load_reduction_with_ghp_kw::Union{Vector{Float64}, Nothing}
     steam_turbine::Union{SteamTurbine, Nothing}
     electric_heater::Union{ElectricHeater, Nothing}
+    water_power::WaterPower
+    water_storage::Array{}
```
hdunham, line 29: what's the point of this? why not use p.s.storage.types.water instead like we do for other storage types?

src/core/scenario.jl
```diff
@@ -57,6 +61,11 @@ A Scenario struct can contain the following keys:
 - [GHP](@ref) (optional, can be Array)
 - [SteamTurbine](@ref) (optional)
 - [ElectricHeater](@ref) (optional)
+- [WaterPower](@ref) (optional)
+- [water_storage](@ref) (optional)
+- [upstream_reservoir](@ref) (optional)
```
hdunham: for the ref to work in the documentation, the text inside [] needs to be the name of a documented struct/function. in this case UpstreamReservoirWaterStorage. you can see the fix I made in commit e1109dccbc84445b5f794963f5a7b5216e15368a.

src/core/scenario.jl
```diff
@@ -229,6 +238,50 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
...
+        push!(water_storage, "downstream_reservoir")
+        downstream_reservoir = DownstreamReservoirStorage(
+                                    dictkeys_tosymbols(d["DownstreamReservoirWaterStorage"]), 
+                                    financial, site, settings.time_steps_per_hour
+                                )
+    else
+        downstream_reservoir = "none"              
+    end
+
+    if !(settings.off_grid_flag) # ElectricTariff only required for on-grid
```
hdunham, line 269: seems unrelated, why is this ElectricTariff part in this PR?

src/core/scenario.jl
```diff
@@ -229,6 +238,50 @@ function Scenario(d::Dict; flex_hvac_from_json=false)
...
+
+    water_storage = []  
+    if haskey(d, "UpstreamReservoirWaterStorage")
+        push!(water_storage, "upstream_reservoir")
+        upstream_reservoir = UpstreamReservoirStorage(
+                                dictkeys_tosymbols(d["UpstreamReservoirWaterStorage"]), 
+                                financial, site, settings.time_steps_per_hour
+                            )
+    else
+        upstream_reservoir = "none"
```
hdunham, line 257: use nothing for these instead of a string

src/results/water_power.jl
```diff
@@ -0,0 +1,177 @@
...
+
+		for i in p.techs.water_power_turbines
+					
+			r["individual_turbine_results"][string(i)*"_results"] = Dict([])
+			
+			r["individual_turbine_results"][string(i)*"_results"]["turbine_power_rating"] = round(value.(m[Symbol("turbine_power_rating"*_n)][i]), digits=3)
+
+			water_outflow_individual = @expression(m, [ts in p.time_steps], m[Symbol("dvWaterOutFlow"*_n)][i, ts])
+			r["individual_turbine_results"][string(i)*"_results"]["water_outflow_series_cubic_meters_per_second"] = round.(value.(water_outflow_individual).data, digits=3)
+			individual_turbine_power_curtailment = @expression(m, [ts in p.time_steps], m[Symbol("dvCurtail")][i, ts])
```
hdunham, line 149: need _n added to dvCurtail

src/results/water_power.jl
```diff
@@ -0,0 +1,177 @@
...
+			for t in p.techs.water_power, ts in p.time_steps)
+	)
+	r["annual_energy_produced_kwh"] = round(value(AnnualWaterPowerProd), digits=0) # includes curtailment
+    	
+	# Compile results for the pumps
+	if ("downstream_reservoir" in p.s.water_storage) && (p.s.water_power.number_of_pumps > 0)
+		
+		PumpsPerUnitSizeOMCosts = @expression(m, p.third_party_factor * p.pwf_om * sum( p.om_cost_per_kw[t] * m[Symbol("dvSize"*_n)][t] for t in p.techs.water_power_pumps))
+		r["pumps_lifecycle_fixed_om_cost_after_tax"] = round(value(PumpsPerUnitSizeOMCosts) * (1 - p.s.financial.owner_tax_rate_fraction), digits=0)
+		r["pumps_initial_capital_costs"] = p.s.upstream_reservoir.cost_per_cubic_meter_upstream_reservoir * sum(m[Symbol("dvSize"*_n)][t] for t in p.techs.water_power_turbines)
```
hdunham, line 87: should use pump cost not reservoir cost? also should loop over pumps not turbines?

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+"""
+`WaterPower` is an optional REopt input with the following keys and default values:
+```julia
+    
+    # Define turbine information
+    number_of_turbines::Real=0, 
+    turbine_cost_per_kw::Real=5000.0,
+    max_kw_turbine::Real=10000000,
+    min_kw_turbine::Real=0,
+    computation_type::String="average_power_conversion",
```
hdunham, line 11: need to document the options for this. or, even better, if there are only 2 options, make it a bool.

hdunham: a more descriptive name would be good too. I wouldn't be able to tell what this input does without looking at the code

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+`WaterPower` is an optional REopt input with the following keys and default values:
+```julia
+    
+    # Define turbine information
+    number_of_turbines::Real=0, 
+    turbine_cost_per_kw::Real=5000.0,
+    max_kw_turbine::Real=10000000,
+    min_kw_turbine::Real=0,
+    computation_type::String="average_power_conversion", 
+    water_turbine_average_cubic_meters_per_second_conversion_efficiency::Real=0, # only applied when the computation_type = "average_power_conversion"
```
hdunham, line 12: I also don't really get what this input represents. conversion to what? the important units to include in the name are the efficiency units not the water flow rate units.

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+    max_kw_turbine::Real=10000000,
+    min_kw_turbine::Real=0,
+    computation_type::String="average_power_conversion", 
+    water_turbine_average_cubic_meters_per_second_conversion_efficiency::Real=0, # only applied when the computation_type = "average_power_conversion"   
+    minimum_water_output_cubic_meter_per_second_total_of_all_turbines::Real=0,
+    minimum_water_output_cubic_meter_per_second_per_turbine::Real=0.0,
+    maximum_water_output_cubic_meter_per_second_per_turbine::Real=0.0,
+    minimum_operating_time_steps_individual_turbine::Real=0.0, # the minimum time (in time steps) that an invidual turbine must run for (can avoid turning a turbine on for just 15 minute, for instance)
+    minimum_operating_time_steps_at_local_maximum_turbine_output::Real=0.0,
+    minimum_turbine_off_time_steps::Real=0.0,
```
hdunham, line 18: is this a constraint that a system would have?

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+    turbine_cost_per_kw::Real=5000.0,
+    max_kw_turbine::Real=10000000,
+    min_kw_turbine::Real=0,
+    computation_type::String="average_power_conversion", 
+    water_turbine_average_cubic_meters_per_second_conversion_efficiency::Real=0, # only applied when the computation_type = "average_power_conversion"   
+    minimum_water_output_cubic_meter_per_second_total_of_all_turbines::Real=0,
+    minimum_water_output_cubic_meter_per_second_per_turbine::Real=0.0,
+    maximum_water_output_cubic_meter_per_second_per_turbine::Real=0.0,
+    minimum_operating_time_steps_individual_turbine::Real=0.0, # the minimum time (in time steps) that an invidual turbine must run for (can avoid turning a turbine on for just 15 minute, for instance)
+    minimum_operating_time_steps_at_local_maximum_turbine_output::Real=0.0,
```
hdunham, line 17: also not sure what this means. what is local maximum referring to? just need documentation for all of these inputs, especially  since it's a more complicated model

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+    # Define the pump information
+    number_of_pumps::Real=0,
+    max_kw_pump::Real=10000000,
+    min_kw_pump::Real=0,
+    pump_cost_per_kw::Real=5000.0,
+    water_pump_average_cubic_meters_per_second_conversion_efficiency::Real=0,
+   
+    are_pumps_reversible::Bool=false,  # If set to true, then establishes a fixed ratio of maximum power to pumps and turbines
+    pump_kw_to_turbine_kw_ratio_for_reversible_pumps::Real=1.0, # Define the maximum power ratio of the pumps to the turbines, if reversible pumps are being modeled
+    minimum_water_flow_cubic_meter_per_second_per_pump::Real=0,
```
hdunham, line 29: make naming consistent between pumps and turbines (e.g. flow vs output)

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+ 
+    # Additional inputs
+    spillway_maximum_cubic_meter_per_second::Real=nothing, # maximum water flow that can flow out of the spillway (structure that enables water overflowing from the reservoir to pass over/through the dam)
+    hydro_production_factor_series::Union{Nothing, Array{<:Real,1}} = nothing, # Optional user-defined production factors. Must be normalized to units of kW-AC/kW-DC nameplate. The series must be one year (January through December) of hourly, 30-minute, or 15-minute generation data.
+    can_net_meter::Bool = off_grid_flag ? false : true,
+    can_wholesale::Bool = off_grid_flag ? false : true,
+    can_export_beyond_nem_limit::Bool = off_grid_flag ? false : true,
+    can_curtail::Bool = true,
+
+    om_cost_per_kw_turbine::Real=0,
```
hdunham, line 41: minor point but a bunch of these seem like pump/turbine inputs so why not move them into the pump/turbine sections above instead of additional inputs section?

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+            replacement_year = 100,
+            discount_rate = f.owner_discount_rate_fraction,
+            tax_rate = f.owner_tax_rate_fraction,
+            itc = stor.total_itc_fraction,
+            macrs_schedule = macrs_schedule,
+            macrs_bonus_fraction = stor.macrs_bonus_fraction,
+            macrs_itc_reduction = stor.macrs_itc_reduction
+        ) - stor.total_rebate_per_kw_turbine
+
+        # Check the inputs for errors:
```
hdunham, line 208: lets do input validation before any other input processing whenever possible

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+    function WaterPower(d::Dict, f::Financial, s::Site, time_steps_per_hour::Int)
+        # TODO: update the struct_name for the sector-specific defaults
+        set_sector_defaults!(d; struct_name="Storage", sector=s.sector, federal_procurement_type=s.federal_procurement_type)
+        stor = WaterPowerDefaults(; d...)
+        
+        macrs_schedule = [0.0]
+        if stor.macrs_option_years == 5 || stor.macrs_option_years == 7
+            macrs_schedule = stor.macrs_option_years == 7 ? f.macrs_seven_year : f.macrs_five_year
+        elseif !(stor.macrs_option_years == 0)
+            throw(@error("ColdThermalStorage macrs_option_years must be 0, 5, or 7."))
```
hdunham, line 172: wrong struct name in error msg

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+            tax_rate = f.owner_tax_rate_fraction,
+            itc = stor.total_itc_fraction,
+            macrs_schedule = macrs_schedule,
+            macrs_bonus_fraction = stor.macrs_bonus_fraction,
+            macrs_itc_reduction = stor.macrs_itc_reduction
+        ) - stor.total_rebate_per_kw_turbine
+
+        # Check the inputs for errors:
+       
+        if stor.minimum_operating_time_steps_individual_turbine < 1
```
hdunham, line 210: the default is 0 though. should the >1 and <1 checks on this everywhere in the code use 0 instead or should the default be 1?

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+    water_pump_average_cubic_meters_per_second_conversion_efficiency::Real=0,
+   
+    are_pumps_reversible::Bool=false,  # If set to true, then establishes a fixed ratio of maximum power to pumps and turbines
+    pump_kw_to_turbine_kw_ratio_for_reversible_pumps::Real=1.0, # Define the maximum power ratio of the pumps to the turbines, if reversible pumps are being modeled
+    minimum_water_flow_cubic_meter_per_second_per_pump::Real=0,
+    maximum_water_flow_cubic_meter_per_second_per_pump::Real=1000000,
+    minimum_operating_time_steps_individual_pump::Real=1,
+ 
+    # Additional inputs
+    spillway_maximum_cubic_meter_per_second::Real=nothing, # maximum water flow that can flow out of the spillway (structure that enables water overflowing from the reservoir to pass over/through the dam)
```
hdunham, line 34: type should be Union{Nothing,:Real} or default should be big number

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+    net_present_cost_per_kw_pump
+    head_pump_meters
+    head_turbine_meters
+    head_loss_pump_meters
+    head_loss_turbine_meters
+
+    function WaterPower(d::Dict, f::Financial, s::Site, time_steps_per_hour::Int)
+        # TODO: update the struct_name for the sector-specific defaults
+        set_sector_defaults!(d; struct_name="Storage", sector=s.sector, federal_procurement_type=s.federal_procurement_type)
+        stor = WaterPowerDefaults(; d...)
```
hdunham, line 166: stor var name seems like a copy paste mistake to fix for clarity

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+        # TODO: implement off_grid capability for water_power
+        #=
+        if off_grid_flag && (can_net_meter || can_wholesale || can_export_beyond_nem_limit)
+            @warn "Setting Existing WaterPower can_net_meter, can_wholesale, and can_export_beyond_nem_limit to False because `off_grid_flag` is true."
+            can_net_meter = false
+            can_wholesale = false
+            can_export_beyond_nem_limit = false
+        end
+        =#
+        net_present_cost_per_kw_pump = effective_cost(;
```
hdunham, line 184: We only do this in the storage structs. I think for techs this is done in cost_curve.jl. These vars also don't seem to be used elsewhere.

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+    net_present_cost_per_kw_turbine
+    net_present_cost_per_kw_pump
+    head_pump_meters
+    head_turbine_meters
+    head_loss_pump_meters
+    head_loss_turbine_meters
+
+    function WaterPower(d::Dict, f::Financial, s::Site, time_steps_per_hour::Int)
+        # TODO: update the struct_name for the sector-specific defaults
+        set_sector_defaults!(d; struct_name="Storage", sector=s.sector, federal_procurement_type=s.federal_procurement_type)
```
hdunham, line 165: Wrong struct name

hdunham (commented):
partial review, haven't looked yet at:
cost_curve_constraints.jl
electric_utility_constraints.jl
load_balance.jl
storage_constraints.jl
water_power_constraints.jl
reopt_multinode.jl

also, I noticed that tests still need to be added

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+   
+    are_pumps_reversible::Bool=false,  # If set to true, then establishes a fixed ratio of maximum power to pumps and turbines
+    pump_kw_to_turbine_kw_ratio_for_reversible_pumps::Real=1.0, # Define the maximum power ratio of the pumps to the turbines, if reversible pumps are being modeled
+    minimum_water_flow_cubic_meter_per_second_per_pump::Real=0,
+    maximum_water_flow_cubic_meter_per_second_per_pump::Real=1000000,
+    minimum_operating_time_steps_individual_pump::Real=1,
+ 
+    # Additional inputs
+    spillway_maximum_cubic_meter_per_second::Real=nothing, # maximum water flow that can flow out of the spillway (structure that enables water overflowing from the reservoir to pass over/through the dam)
+    hydro_production_factor_series::Union{Nothing, Array{<:Real,1}} = nothing, # Optional user-defined production factors. Must be normalized to units of kW-AC/kW-DC nameplate. The series must be one year (January through December) of hourly, 30-minute, or 15-minute generation data.
```
hdunham, line 35: not used elsewhere in code

src/core/water_power.jl
```diff
@@ -0,0 +1,267 @@
...
+    max_kw_pump::Real=10000000,
+    min_kw_pump::Real=0,
+    pump_cost_per_kw::Real=5000.0,
+    water_pump_average_cubic_meters_per_second_conversion_efficiency::Real=0,
+   
+    are_pumps_reversible::Bool=false,  # If set to true, then establishes a fixed ratio of maximum power to pumps and turbines
+    pump_kw_to_turbine_kw_ratio_for_reversible_pumps::Real=1.0, # Define the maximum power ratio of the pumps to the turbines, if reversible pumps are being modeled
+    minimum_water_flow_cubic_meter_per_second_per_pump::Real=0,
+    maximum_water_flow_cubic_meter_per_second_per_pump::Real=1000000,
+    minimum_operating_time_steps_individual_pump::Real=1,
```
hdunham, line 31: need to make more clear that this is for each time it turns on not total over year

src/core/water_power.jl
hdunham: after hearing your explanation, I'm wondering if a ramp rate constraint would get at the root issue and avoid indicator constraints. let's discuss in next larger meeting.

src/core/water_storage.jl
hdunham: never mind, need this so can set to zero and model closed loop, or if there's a limit to prevent river flooding or something

src/constraints/electric_utility_constraints.jl
```diff
@@ -1,5 +1,6 @@
 # REopt®, Copyright (c) Alliance for Energy Innovation, LLC. See also https://github.com/NatLabRockies/REopt.jl/blob/master/LICENSE.
 function add_export_constraints(m, p; _n="")
+    print("\n #1: p.techs.all is: $(p.techs.all)")
```
hdunham, line 3: I'm making a note to remove these debug prints before merging finally into develop

src/constraints/water_power_constraints.jl
```diff
@@ -0,0 +1,272 @@
+# REopt®, Copyright (c) Alliance for Sustainable Energy, LLC. See also https://github.com/NREL/REopt.jl/blob/master/LICENSE.
+
+const WATER_DENSITY = 998
```
hdunham, line 3: need units

src/constraints/water_power_constraints.jl
```diff
@@ -0,0 +1,272 @@
...
+		@warn "Setting minimum_operating_time_steps_at_local_maximum_turbine_output to greater than 1 requires an optimization solver that can handle indicator constraints."
+		
+		if p.s.settings.solver_name in solvers_incompatible_with_indicator_constraints
+			throw(@error("A solver that can handle indicator constraints must be used if minimum_turbine_off_time_steps is set to greater than 1"))
+		end
+
+		print("\n Adding a constraint for the minimum operating time at a local maximum water flow \n")
+		for dv in dvs
+			if dv == "binTurbineActive"*_n
+				variable = Symbol("dvWaterOutFlow")
```
hdunham, line 237: add _n?

src/constraints/water_power_constraints.jl
```diff
@@ -0,0 +1,272 @@
...
+	
+		print("\n Adding minimum off duration for the turbines \n")
+		for t in p.techs.water_power_turbines, ts in 1:Int(length(p.time_steps)- p.s.water_power.minimum_turbine_off_time_steps - 1 )
+			@constraint(m, m[Symbol("indicator_turbine_turn_off"*_n)][t, ts] =>  { sum(m[Symbol("binTurbineActive"*_n)][t,ts+i] for i in 1:p.s.water_power.minimum_turbine_off_time_steps) <= 0 } ) 
+			@constraint(m, !m[Symbol("indicator_turbine_turn_off"*_n)][t, ts] => { m[Symbol("binTurbineActive"*_n)][t,ts+1] - m[Symbol("binTurbineActive"*_n)][t,ts] >= 0  } )
+		end
+	end
+	
+	# TODO: remove this constraint, which prevents a spike in the spillway use during the first time step
+	@constraint(m, [ts in p.time_steps], m[Symbol("dvSpillwayWaterFlow"*_n)][1] == 0)
```
hdunham: remove loop bs ts not used

src/constraints/cost_curve_constraints.jl
```diff
@@ -259,6 +259,30 @@ function initial_capex_no_incentives(m::JuMP.AbstractModel, p::REoptInputs; _n="
         )
     end
 
+    # Water power technologies
+    if !isempty(p.techs.water_power_pumps)
+        add_to_expression!(m[:InitialCapexNoIncentives],
```
hdunham, line 264: Need _n on all of these since it looks like you did add water power vars to reopt_multinode.jl. It's fine to not implement multinode for PHS capability, but need to give the user an error if they try instead.

hdunham: sorry this comment should have been placed a few lines down on the water storage constraints

src/core/techs.jl
```diff
@@ -58,6 +61,27 @@ function Techs(p::REoptInputs, s::BAUScenario)
...
+                push!(elec, water_power_name_turbine)
+                push!(water_power, water_power_name_turbine)
+                push!(water_power_turbines, water_power_name_turbine)
+            end
+                
+            # Add a tech for each pump separately
+            for i in 1:s.water_power.number_of_pumps
+                water_power_name_pump = "WaterPower_Pump"*string(i)
+                push!(all_techs, water_power_name_pump)
+                push!(elec, water_power_name_pump)
```
hdunham, line 79: does it make sense for pumps to be added to elec? just since they dont have elec production

src/constraints/storage_constraints.jl
```diff
@@ -78,16 +72,20 @@ function add_elec_storage_dispatch_constraints(m, p, b; _n="")
...
         m[Symbol("dvStoredEnergy"*_n)][b, ts] == m[Symbol("dvStoredEnergy"*_n)][b, ts-1] + p.hours_per_time_step * (  
             sum(p.s.storage.attr[b].charge_efficiency * m[Symbol("dvProductionToStorage"*_n)][b, t, ts] for t in p.techs.elec) 
-            + p.s.storage.attr[b].grid_charge_efficiency * m[Symbol("dvGridToStorage"*_n)][b, ts] 
-            - m[Symbol("dvDischargeFromStorage"*_n)][b,ts] / p.s.storage.attr[b].discharge_efficiency
+            + p.s.storage.attr[b].grid_charge_efficiency * m[Symbol("dvGridToStorage"*_n)][b, ts]
+            - (m[Symbol("dvDischargeFromStorage"*_n)][b,ts]  / p.s.storage.attr[b].discharge_efficiency)
         )
 	)
+
+    # Note: Hydro to storage not added for off-grid scenarios
```
hdunham, line 80: add error to prevent user from running this

src/constraints/storage_constraints.jl
hdunham, line 2: This gets called on storage.types.all, which includes water storage, but water storage doesn't have the kw and kwh fields being used here. Seems like there would be an error when you try to run unless I'm missing something. This gets to my point of why the reservoirs themselves shouldn't be considered storage but the system as a whole can be if things are restructured.

hdunham: and then add_general_storage_dispatch_constraints depends on add_storage_size_constraints working

src/constraints/electric_utility_constraints.jl
```diff
@@ -389,8 +393,8 @@ end
 
 
 function add_elec_utility_expressions(m, p; _n="")
-
-    if !isempty(p.s.electric_tariff.export_bins) && !isempty(p.techs.all)
+    # TODO: if prefered, change the if statement below so that the water_power does not need to be called out separately (note: hydro is removed from p.techs.all in reopt.jl in this line of code: NonHydroTechs = filter!(x->x != water_power_tech, p.techs.all))
```
hdunham, line 396: seems this filter is no longer in place, which is good because I was going to suggest getting rid of it

hdunham: so don't need || !empty(waterpower) anymore then right, since it's in techs.all?

src/constraints/storage_constraints.jl
hdunham: add_general_storage_dispatch_constraints also don't really apply to reservoirs since the "SOC" equivalent and in/out flow constraints are already defined in a different way in the water_power_constraints file

PR #608, opened 2026-06-03
Add existing CHP capacity support

bpulluta:
- [x] CHANGELOG.md is updated
- [x] Tests for the changes have been added. These tests compare expected behavior and include CHP-specific regression coverage.

### Description of changes
This PR adds support for existing CHP capacity in REopt.jl by:
- adjusting net electric load for existing CHP using a flat-out dispatch assumption (`min(load, CHP.existing_kw)`)
- keeping CHP sizing semantics aligned so `min_kw` and `max_kw` apply to new capacity beyond existing capacity
- preserving BAU / proforma accounting so existing CHP is not billed as new capex

### Validation
- Added focused CHP existing-capacity regression coverage for:
  - net-load adjustment
  - sizing / capex
  - BAU CHP bounds
- Ran the focused CHP test file successfully.


src/results/chp.jl
```diff
@@ -163,8 +163,8 @@ function get_chp_results_for_tech(m::JuMP.AbstractModel, p::REoptInputs, chp_nam
 	r["year_one_standby_cost_after_tax"] = r["year_one_standby_cost_before_tax"] * (1 - p.s.financial.offtaker_tax_rate_fraction)
 	r["lifecycle_standby_cost_after_tax"] = round(r["year_one_standby_cost_before_tax"] * p.pwf_e * (1 - p.s.financial.offtaker_tax_rate_fraction), digits=0)
 	
-	# Calculate individual CHP capital costs
-	size_kw = value(m[Symbol("dvSize"*_n)][chp_name])
+	# Calculate individual CHP capital costs using purchased CHP capacity (excluding existing_kw)
+	size_kw = value(m[Symbol("dvPurchaseSize"*_n)][chp_name])
```
Bill-Becker, line 173: For PV, we use dvSize which is the combined existing+new PV kW. Thoughts on why we wouldn't do the same here?

src/core/bau_scenario.jl
```diff
@@ -130,6 +131,18 @@ function BAUScenario(s::Scenario)
     =#
     site = bau_site(s.site)
 
+    # Include only CHPs with existing capacity; no new capacity or supplementary firing in BAU
+    chps = CHP[]
+    for chp in s.chps
+        if chp.existing_kw > 0
+            bau_chp = deepcopy(chp)
+            bau_chp.min_kw = 0.0
```
Bill-Becker: Why not do what we do for PV here, setting min_kw and max_kw == existing_kw? I see that you later override this in bau_inputs.jl, but wondering why we can't take the same approach as PV.

src/core/bau_scenario.jl
```diff
@@ -130,6 +131,18 @@ function BAUScenario(s::Scenario)
     =#
     site = bau_site(s.site)
 
+    # Include only CHPs with existing capacity; no new capacity or supplementary firing in BAU
```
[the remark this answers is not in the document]
Bill-Becker: See other comment/question for setting max/min == 0 vs existing, but for supplementary, I think we just keep whatever setting the user specifies for new CHP the same as for existing CHP. If they want to model it as a different CHP, they'd have to use the "multiple CHP" capability but that will only be available in the backend tools. I think we can remove any special setting of supplementary firing parameters. The CHP input supplementary_firing_max_steam_ratio will determine if both the BAU and optimal scenarios can do supplementary firing.

src/results/proforma.jl
```diff
@@ -522,10 +522,21 @@ function irr(cashflows::AbstractArray{<:Real, 1})
         return 0.0
     end
     f(r) = npv(cashflows, r)
-    rate = 0.0
-    try
-        rate = fzero(f, [0.0, 0.99])
-    finally
-        return round(rate, digits=3)
+    a, b = 0.0, 0.99
```
Bill-Becker: Why are these updates needed when existing CHP was added?

test/test_chp_existing_kw.jl
```diff
@@ -0,0 +1,155 @@
+# using Revise
```
Bill-Becker, line 1: The purpose of some of these tests in unclear to me, and it looks like we're testing the same types of things multiple times. Can you add a "purpose" statement for each of the test sections, and try to remove redundant checks. Even if some redundant tests aren't adding that much time, I think it would be easier to understand the purpose of the complete test set by making it more concise and purposeful.

Bill-Becker (commented):
@bpulluta Looks good in general, but I have a few questions and requests. I had started with a copilot review that just came up with two questions/suggestions, but those comments are incorporated into my review.

src/results/chp.jl
Bill-Becker: I actually like that we separately report the size_kw_bau instead of relying on the input s.chp.existing_kw to distinguish between new and existing. I think it's Ok to be different from PV where size_kw is just the new kW, but we need to make sure the financial/proforma results processing is consistent to account for e.g. upfront CapEx.

src/core/bau_scenario.jl
bpulluta: Agreed, the BAU already reads CHP from `existing_kw`, so I removed the extra min/max assignment.

src/results/chp.jl
bpulluta: I kept `size_kw` as total CHP size, but charge capex on `dvPurchaseSize` so existing capacity is not billed twice.

src/results/proforma.jl
bpulluta: undid the change, i think for some reason tests were failing at one point and i did this to see if it helped. definitely not related to chp so no need to keep this change

bpulluta (commented):
@Bill-Becker, ive resolved the pending comments and addressed the things we talked about related to replicating pv logic for net load

Bill-Becker (approved):

PR #610, opened 2026-06-11
update developer env variable names

hdunham:
- [x] CHANGELOG.md is updated
- [x] Any new REopt outputs and required inputs have been added to the corresponding Django model in the REopt API


src/core/utils.jl
```diff
@@ -652,18 +651,32 @@ function convert_temp_degF_to_Kelvin(degF::Float64)
...
+            throw(@error("No NLR Developer API Key provided when trying to call PVWatts or Wind Toolkit.
+                    Within your Julia environment, specify ENV['NLR_DEVELOPER_API_KEY']='your API key'
                     See https://natlabrockies.github.io/REopt.jl/dev/ for more information."))
+        else
+            ENV["NLR_DEVELOPER_API_KEY"] = ENV["NREL_DEVELOPER_API_KEY"]
+            @warn "The environment variable NREL_DEVELOPER_API_KEY will be discontinued. Use NLR_DEVELOPER_API_KEY instead.
+                    Within your Julia environment, specify ENV['NLR_DEVELOPER_API_KEY']='your API key'
+                    See https://natlabrockies.github.io/REopt.jl/dev/ for more information."
+        end
     end
```
rathod-b, line 665: A minor note here and for NLR developer email warnings is if we should use "will be discontinued" or if we should say "has been discontinued." Thoughts?

hdunham: because of the backwards compatibility that I added, "will be" is correct

rathod-b (approved):

PR #616, opened 2026-07-31
do not double count pv capex

rathod-b:
do not double count pv capex, update multiple pv test remove any incentives for PV and test in runtests.jl to see if installed cost outputs in financial outputs match.

- [x] CHANGELOG.md is updated
- [ ] Docs have been added / updated if applicable


Bill-Becker (approved):
Nice fix!

PR #620, opened 2026-07-31
Chp follow thermal load

zolanaj:
- [x] CHANGELOG.md is updated

### Added
- Added attribute `CHP.follow_heating_load` which, when set to `true`, adds new constraints via function `add_chp_heating_load_following_constraints` to enforce a policy in which CHP runs at capacity if the CHP size is less than the heating load, and restricts the `ExistingBoiler` from serving the load when CHP size is greater than the heating load in each time period (this allows for other heating techs like CST to contribute potentially).


src/core/chp.jl
```diff
@@ -291,6 +293,23 @@ function CHP(d::Dict;
...
+    end 
+
+    # Validate CHP doesn't follow thermal load and only send to absorption chiller
+    if chp.follow_heating_load && chp.serve_absorption_chiller_only
+        throw(@error(
+                "CHP cannot both follow thermal heating load and serve only the " *
+                "absorption chiller. Update either CHP.serve_absoprtion_chiller_only " *
+                "or CHP.follow_heating_load to false."
+            ))
+    end
```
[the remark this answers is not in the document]
zolanaj: This will be kept consistent with the other error message passing in this repo; no changes planned in response to this comment.

src/constraints/chp_constraints.jl
```diff
@@ -219,6 +219,50 @@ function add_chp_electrical_load_following_constraints(m, p; _n="")
...
+    add_chp_heating_load_following_constraints(m, p; _n="")
+
+Used in function add_chp_constraints to add constraints that restrict output of CHP to serve full eligible heating load or 
+run at capacity otherwise in dispatch (without regard to electrical flows).
+"""
+function add_chp_heating_load_following_constraints(m, p; _n="")
+    heat_to_electric_ratio = p.s.chp.thermal_efficiency_full_load / p.s.chp.electric_efficiency_full_load
+    chp_eligible_heat_load = zeros(length(p.s.electric_load.loads_kw))
+    heating_loads_served_by_chp = []
+    if p.s.chp.can_serve_dhw
```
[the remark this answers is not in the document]
zolanaj: The suggested fix would break an if/then statement, so the recommended update was implemented in https://github.com/NatLabRockies/REopt.jl/commit/ec04ddc9fcced149cc61a511bbbd16b53bf6d952 instead.

test/runtests.jl
```diff
@@ -1471,6 +1471,26 @@ else  # run HiGHS tests
                 @test value(m[:dvRatedProduction]["CHP",4045]) < p.s.electric_load.loads_kw[4045]
                 @test value(m[:dvHeatingProduction]["CHP","DomesticHotWater",4045]) ≈ value(m[:dvHeatToAbsorptionChiller]["CHP","DomesticHotWater",4045]) rtol=1e-4
                 @test value(m[:dvProductionToWaste]["CHP","DomesticHotWater",4045]) ≈ 0.0 atol=1e-4
+            
+                #Part 5: CHP Thermal Load Following
+                d = JSON.parsefile("./scenarios/chp-abschl-flow.json")
+                delete!(d, "AbsorptionChiller")
+                d["CHP"]["follow_heating_load"] = true
+                d["CHP"]["can_curtail"] = true
```
Bill-Becker, line 1479: We talked about maybe making the default CHP.can_curtail = true when follow_heating_load = true. I see you're setting that to true here for this test. I could go either way on this because in reality the CHP system should either turn down if they're not allowed to export power and there is no storage to dump it to, so those should be separately specified, but I can see potential for infeasibilities if they do not. "Curtailing" for CHP would have to be dumping load into a resistor/load bank, so they'd need that equipment. But CHP is not allowed to turn down to avoid excess electricity when follow_heating_load = true - it would otherwise error for infeasibility, right?

Bill-Becker (commented):
Just one comment to address based on our discussion. I thought you had also updated the electrical load following in this branch, but did not see any updates here.

zolanaj:
> Just one comment to address based on our discussion. I thought you had also updated the electrical load following in this branch, but did not see any updates here.

The update to electric load following is actually implemented in https://github.com/NatLabRockies/REopt.jl/pull/612 but I don't think I updated the changelog sufficiently to describe the reformulation.  EDIT: the changelog for 0.59.3 looks ok and include the formulation update.

test/runtests.jl
zolanaj: You're correct that inability to turn down to avoid excess heating load would prelude this in a CHP only scenario, but the formulation would allow for other systems like TES to cover any shortfall and let the CHP system stay off.  As a result, we should provide a warning that this combination may lead to infeasibilities or large storage sizes as a result of the policy.  I added a new warning here: https://github.com/NatLabRockies/REopt.jl/commit/b8115342295ae7cfb152f45f4894c0b8f69ceee3

Question: should the condition `can_curtail == true` include either a warning or an `@info` note about the added cost?  Is there any special equipment for heat rejection or it that assumed as well?

Bill-Becker: I'm good with the warning you added. For heat rejection, I don't think there's typically extra equipment but the exhaust stack just has to handle higher temperatures because the exhaust doesn't get cooled down. I would assume they design the exhaust to handle those temperatures anyway even if not expected much. We've always been assuming heat can be "vented/wasted" without any additional equipment or warning, so I think we can leave it.

Bill-Becker (approved):
I'm going to merge a few different CHP things today, so I'll take care of the merges.

Issue #609, opened 2026-06-08
Bug with NEM and WHL combined inputs


adfarth:
In this post ([d.json](https://github.com/user-attachments/files/28724167/d.json) ): 
- the PV size is fixed to a size larger than the net metering limit
- There is a wholesale rate
- PV can both net meter or wholesale

We would expect REopt to choose to do wholesale, but instead it doesn't export at all.

In other posts too it seems like REopt is not doing the expected thing of choosing which is more beneficial: NEM or WHL.

PR #623, opened 2026-08-13
Add NEM and wholesale regression coverage

Gaurav890:
## Summary

- add regression coverage for a fixed PV system sized above the net-metering limit when wholesale exports are enabled
- verify the model selects wholesale rather than NEM and exports the independently calculated annual energy
- protect the existing wholesale max-benefit fix against regressions

The model fix is already present on `develop`; this PR adds focused coverage for the scenario reported in #609.

## Validation

- focused Julia 1.10.11 + HiGHS regression test: 4/4 assertions passed
- expected annual wholesale export: 192,809 kWh
- actual annual wholesale export: 192,809 kWh
- `git diff --check`

Refs #609

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1076,6 +1076,34 @@ else  # run HiGHS tests
             empty!(m)
             GC.gc()
 
+            # Case 6: fixed PV above the NEM limit must use the wholesale export bin (#609)
+            d = JSON.parsefile("./scenarios/net_metering.json")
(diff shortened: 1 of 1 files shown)


Bill-Becker:
@Gaurav890 I'm good with this PR, but we are phasing out the develop branch so can you change this PR to merge into master instead?

Bill-Becker:
> @Gaurav890 I'm good with this PR, but we are phasing out the develop branch so can you change this PR to merge into master instead?

Nevermind, I was able to change it to master.

PR #624, opened 2026-08-14
Fix CHP Customized Dispatch

Bill-Becker:
This pull request updates the CHP (Combined Heat and Power) electrical load-following constraints and improves the corresponding test coverage. The main change is to ensure that the CHP's electric production is correctly scaled by the `production_factor` in both the constraint definition and the test validation, making the model and tests more robust and accurate.

**Constraint correction:**

* The constraint in `add_chp_electrical_load_following_constraints` now multiplies the left-hand side by `production_factor`, ensuring the modeled CHP production correctly reflects the time-varying production capability.

**Test improvements for load-following:**

* The test for the load-following CHP scenario was rewritten to:
  - Use a time-varying `production_factor_series` (0.5 for half the year, 1.0 for the other half).
  - Check that the actual CHP output matches the expected output (max capacity times production factor) to within a small tolerance.
  - Confirm that average production is higher in the second half of the year, as expected from the higher production factor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ Classify the change according to the following categories:
     ### Deprecated
     ### Removed
 
+## v0.61.1
+### Fixed
+- Avoid double-multiplication of production_factor_series with electrical load-following
+
 ## v0.61.0
 ### Added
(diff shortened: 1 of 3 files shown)


zolanaj (approved):
Thanks for this @Bill-Becker!  That addition is more efficient as a removal of the production factor from the other side of the equation, and then looking at the load balancing, I noticed levelization factor should have been part of the load-following check as well.  Assuming you're good with the change, this can be merged.

PR #628, opened 2026-09-03
updated defaults to 2025ATB

rathod-b:
- [x] CHANGELOG.md is updated
- [ ] Tests for the changes have been added. These tests should compare against expected values that are calculated independently of running REopt. Tests should also include garbage collection (see other tests for examples).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ Classify the change according to the following categories:
 
 ## defaults-update-atb25
 ### Changed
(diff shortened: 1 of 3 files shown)


Bill-Becker (approved):
I reviewed the numbers/calcs on the ATB 2025 spreadsheet, and I think these all look correct and consistent with that. Let's include that we used ATB 2025 (as you already stated) and then escalated by 6% based on BLS inflation data to get 2025 dollars from the 2023 dollars in the ATB, in the CHANGELOG.

